PR #312, opened 2013-03-22
Python 3 compatibility

alexchandel:
I've added support for python 3. That includes changing module/class names (like cStringIO to io.ByteIO), changing print statements, fixing exception raising, and working out the bytes/str mess. It's backwards compatibile (except the new classnames obviously), and I've tested it against RMQ 3.0.4 on python 3.3 and python 2.7.2.

I normally use python 3, but I have to spawn a 2.7 subprocess to deal with pika. Please include 3.X support!

### DIFF
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -870,6 +870,8 @@ class Channel(object):
 
         """
         LOGGER.debug('Called with %r, %r, %r', method_frame, header_frame, body)
+        if type(body) is str:
+            body = body.encode()
         consumer_tag = method_frame.method.consumer_tag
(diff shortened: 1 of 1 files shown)


hayd:
This PR does some strange things with strange things with unicode (I don't think you can simply replace it with str and assume that this works in both python 2 and 3, surely it doesn't) :s 

@gmr to "sync up", is there a branch for py3? Is there a road-map (is making pika python 2 and 3 compatible the objective?). Is it worth running [`futurize -1`](https://github.com/PythonCharmers/python-future) on the code-base and trying to get the tests to pass?

PR #486, opened 2014-06-30
Don't reject message if consumer has 'no_ack' flag

anjensan:
Pika tries to reject messages even if 'no_ack' flag is set. 
It may cause "PRECONDITION_FAILED - unknown delivery tag" errors and unexpected channel closing.


andrewmichaelsmith:
I've raised #530 for this issue

gmr:
Closing since tests don't pass, will implement similar logic.

PR #492, opened 2014-07-27
Fix confirmation on reconnects

bra-fsn:


gmr:
I moved the reset to a more appropriate method, but appreciate the patch

PR #496, opened 2014-08-01
Fix number of arguments passed to _on_return callback

aeirola:
Callback function added with `add_on_return_callback` received all arguments within a tuple. This should fix it.

### DIFF
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -968,10 +968,10 @@ class Channel(object):
 
         """
         if not self.callbacks.process(self.channel_number, '_on_return', self,
-                                      (self,
-                                       method_frame.method,
-                                       header_frame.properties,
-                                       body)):
+                                      self,
+                                      method_frame.method,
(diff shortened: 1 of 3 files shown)


sjlongland:
I've been confused about this issue myself.  Either the documentation is wrong or the code is wrong.

I thought I was doing something wrong in my code, and it's only when I went digging I found this pull request.

It is not a difficult to fix code that uses the old (broken?) behaviour:

```
def on_return(on_return_tuple):
    (channel, basic_return, properties, body) = on_return_tuple
    # ...
```

can be changed to:
```
def on_return(*on_return_tuple):
    (channel, basic_return, properties, body) = on_return_tuple
    # ...
```

The question remains, how many people would be inconvenienced by this, versus how many are being inconvenienced by on_return callbacks that don't work for mysterious reasons?

gmr:
With the 0.10.0 change, I'm ok with the API change. Thanks for the patch!

PR #501, opened 2014-09-10
Fix SSLError event processing.

kherrala:
Do not try to use IOLoop to handle socket events.
Poll for events using select.
Catch socket timeout exception.
Remove unused class variable DO_HANDSHAKE.

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -282,9 +282,11 @@ class BaseConnection(connection.Connection):
         elif self.params.ssl and isinstance(error_value, ssl.SSLError):
 
             if error_value.args[0] == ssl.SSL_ERROR_WANT_READ:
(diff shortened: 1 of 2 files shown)


kherrala:
Any plans getting this to master? It has been working fine for me. Using standard select() during initial handshake doesn't seem like a big issue to me.

vitaly-krugl:
Using `select()` here will interfere with the native framework (e.g., Tornado). Also, `select()` has serious limitations with respect to max socket file descriptor number; so, this would cause failures (or require changes) in programs that create many thousands of sockets and other file descriptor-based objects. poll/epoll/etc. don't have that problem.

gmr:
Closing due to not using native adapters for behavior which seems wrong. Thanks for your patch. If you want to (and it's still needed), please resubmit without the select based implementation, based upon current master.

PR #505, opened 2014-09-21
Socket improvements

kherrala:
Most important fix here is to stop processing events if socket is closed during error on handle_read or handle_write. It also seems like previous check for closed socket should not be checking for method parameter fd.

Calling socket.shutdown() was inspired by this:
http://engineering.imvu.com/2014/03/06/charming-python-how-to-actually-close-a-socket-by-calling-shutdown-before-calling-close/

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -28,6 +28,7 @@ class BaseConnection(connection.Connection):
     ERROR = 0x0008
 
     ERRORS_TO_IGNORE = [errno.EWOULDBLOCK, errno.EAGAIN, errno.EINTR]
+    ERRORS_TO_ABORT = [errno.EBADF, errno.ECONNABORTED, errno.EPIPE]
     DO_HANDSHAKE = True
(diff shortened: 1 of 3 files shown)


gmr:
I like this, but need to cherry pick stuff since not all of it applies anymore.

I'll credit you but commit the changes directly in my own branch.

pika/adapters/base_connection.py
```diff
@@ -372,10 +371,22 @@ def _init_connection_state(self):
...
+
+    def _cleanup_socket(self):
+        """Close the socket.
+
+        :return:
+        """
+        if self.socket:
+            try:
+                self.socket.shutdown(socket.SHUT_RDWR)
+            except socket.error as err:
```
vitaly-krugl: No point for `as err`, since `err` is not used

vitaly-krugl:
@kherrala, that's a good point about dependency on garbage-collection. I wasn't aware of this socket subtlety in python. Thanks!

Issue #507, opened 2014-09-29
Publishing from within a callback can result in out-of-order frames


happybob007:
Using pika 0.9.14

I have a small program using one direct exchange (BlockingConnection) with two unnamed queues, and two callbacks.  The first callback receives messages and for each message it receives, publishes a different message to the same exchange but with a different routing key.  The second callback receives these new messages and just prints them out.

If the original message source publishes slowly (no more than one or two messages per second), then there are no problems.   However, if the original message source publishes messages rapidly, after a few messages, this error shows up:

```
WARNING:pika.adapters.blocking_connection:Disconnected from RabbitMQ at mq.gic.smartsheet.com:5672 (505): UNEXPECTED_FRAME - expected content body, got non content body frame instead
```

Any further publishing results in a CRITICAL error and program termination.
Looking at the logs on the rabbitmq server, I see:

```
=ERROR REPORT==== 29-Sep-2014::13:13:08 ===
AMQP connection <0.22161.15> (running), channel 1 - error:
{amqp_error,unexpected_frame,
            "expected content body, got non content body frame instead",
            'basic.publish'}
```

If I use pika 0.9.9, the program works.

I did some digging, and it looks like the problem stems from `connection._send_method()`.  In there, the message I am publishing gets turned into several frames (at least 3 (<METHOD()>, <Header()>, and <Body()>).  These frames are marshaled and pushed on to the `outbound_buffer` for the connection.  However, the frames are not pushed onto `outbound_buffer` as a single unit, and what appears to be happening for me is that the read event callback (from data available on the socket) is resulting in the first of my callbacks being called, which then calls `basic_publish` and injects a <METHOD())> frame where RabbitMQ is expecting a <Body()> frame.

The root of the problem is that `connection._send_method()` is not able to push a contiguous block of frames onto the `outbound_buffer`, and instead is pushing the frames on one at a time.

PR #508, opened 2014-10-11
Make _send_method() deliver frames contiguously, resolving issue 507

happybob007:
This addresses issue #507.
I think that the twisted_connection may still exhibit the same problem after this fix -- I simply patched twisted_connection._send_frame() to ignore the extra frame accumulator parameter.


gmr:
Thanks for this, I ended up using a different approach in de8b5450, but appreciate your take.

PR #512, opened 2014-11-03
docs: fix signature of basic_cancel

lclementi:
Channel.basic_cancel gets two named parameters in input.
Fix this example accordingly


gmr:
Why do you want them to be named, the behavior of the example works as is. Closing, feel free to reopen if I'm missing something.

Issue #12, opened 2010-07-22
Pika does not support longs in tables


ask:
Pika (along with almost every other dynamic language client for AMQP) does not support longs in tables.

I tried the obvious fix to simply add the following to encode_table:

```
    elif isinstance(value, long):
        pieces.append(struct.pack('>cl', 'l', value))
        tablesize += 5
```

But that didn't work, as RabbitMQ gives the following traceback:

```
=ERROR REPORT==== 22-Jul-2010::13:01:28 ===
** Generic server <0.11385.5> terminating 
** Last message in was {'EXIT',<0.11384.5>,
                       {function_clause,
                           [{rabbit_binary_parser,parse_field_value,
                                [<<113,0,0,0,0,0,0,39,16>>]},
                            {rabbit_binary_parser,parse_table,1},
                            {rabbit_framing,decode_method_fields,2},
                            {rabbit_framing_channel,mainloop,1}]}}
** When Server state == {ch,running,1,<0.11382.5>,<0.11383.5>,undefined,none,
                        {set,0,16,16,8,80,48,
                             {[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                              [],[]},
                             {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                               [],[]}}},
                        1,
                        {[],[]},
                        {[],[]},
                        <<"guest">>,<<"/">>,<<>>,
                        {dict,0,16,16,8,80,48,
                              {[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                               [],[]},
                              {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                                [],[]}}}}
```

I managed to parse the generated field manually with parse_field_value, so
maybe there's something I'm missing. You might have the context needed to solve this.

PR #516, opened 2014-11-21
make skip IncompatibleProtocolError as optional for client to adapt scenarios like ELB connection

jimhorng:
In my scenario with configurations
* tornado server publishes message to MQ through ELB
* ELB in front of rabbitMQ(s)
* rabbitMQ setup as a cluster

when some of rabbitMQ is down and the cluster is not in function, tornado's pika connection will receive error like below, and all the error handling and re-connection mechanism will NOT be invoked since the exception breaks the calls to _on_connection_closed which registered error handling callbacks.

"""
[2014-11-19 01:04:30,841][ERROR  ][tornado.application:681] : Exception in I/O handler for fd 21#012Traceback (most recent call last):#01
2  File "/tornado/ioloop.py", line 671, in start#012    self._handlers[fd](fd,
events)#012  File "/tornado/stack_context.py", line 331, in wrapped#012    raise_exc_info(exc)#012  File "/tornado/stack_context.py", line 302, in wrapped#012
    ret = fn(*args, **kwargs)#012  File "/pika/adapters/base_connection.py”, line 322, in _handle_events#012    self._handle_read()#012 
File "/pika/adapters/base_connection.py", line 348, in _handle_read#012   
return self._handle_disconnect()#012  File "/pika/adapters/base_connection.py", line 248, in _handle_disconnect#012    self._adapter_disconnect()#012  File "/pika/adapters/tornado_connection.py", line 75, in _adapter_disconnect#012   
super(TornadoConnection, self)._adapter_disconnect()#012  File "/pika/adapters/base_connection.py", line 137, in _adapter_disconnect#012
self._check_state_on_disconnect()#012  File "/pika/adapters/base_connection.py", line 149, in _check_state_on_disconnect#012   
raise exceptions.IncompatibleProtocolError#012IncompatibleProtocolError
"""

/adapters/base_connection.py
```
    def _handle_disconnect(self):
        """Called internally when the socket is disconnected already
        """
        self._adapter_disconnect()   # BREAKS HERE !
        self._on_connection_closed(None, True)
```

In this scenario "IncompatibleProtocolError" is not treated as severe error and the connection can be recovered by re-connecting. Therefore I proposed that client can have option to skip this error for the appropriate scenario.

Thanks.

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -146,7 +146,8 @@ class BaseConnection(connection.Connection):
         """
         if self.connection_state == self.CONNECTION_PROTOCOL:
             LOGGER.error('Incompatible Protocol Versions')
-            raise exceptions.IncompatibleProtocolError
+            if not self.params.skip_incompatible_protocol_error:
+                raise exceptions.IncompatibleProtocolError
         elif self.connection_state == self.CONNECTION_START:
             LOGGER.error("Socket closed while authenticating indicating a "
                          "probable authentication error")
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -62,6 +62,7 @@ class Parameters(object):
     DEFAULT_PASSWORD = 'guest'
     DEFAULT_PORT = 5672
     DEFAULT_RETRY_DELAY = 2.0
+    DEFAULT_SKIP_INCOMPATIBLE_PROTOCOL_ERROR = False
     DEFAULT_SOCKET_TIMEOUT = 0.25
     DEFAULT_SSL = False
     DEFAULT_SSL_OPTIONS = {}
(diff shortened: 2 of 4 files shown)


gmr:
I'm not convinced this is the right way to handle this. IncompatibleProtocolError is raised when you open a connection and the remote side closes the connection before the AMQP connection can be negotiated. One could make a case that it should do an on_connection_error and that should contain the callback -- or that the TornadoConnection adapter some of Tornado's built in constructs to capture errors at this stage.

jimhorng:
@gmr , thanks for comments, 
1) do you mean adding another error capturer around
pika/adapters/base_connection.py 
``` 
def _handle_error(self, error_value):
```
for this kind of error? and set a flag to skip raising IncompatibleProtocolError
pika/adapters/base_connection.py
```
_check_state_on_disconnect()
```
so that `self._on_connection_closed` can be called to trigger `connection.on_close()` callback registered by client?

2) or create a new callback listener like `on_connection_error` around 
pika/adapters/base_connection.py 
``` 
def _handle_error(self, error_value):
```
?

gmr:
I think your problem may live with your ELB configuration. I use pika to talk to RabbitMQ through ELB and do not run into any issues. I am also not convinced the approach is correct.

I appreciate the PR but will not be merging it at this time.  Thanks for your contribution and raising the issue.

jimhorng:
@gmr , got it, thanks for taking time to look at this PR anyway.
I'l try to draft more solid scenario for reproducing this problem in future since I can reproduce it easily :(

PR #523, opened 2015-01-21
pika: callback: Display exception message when callback fails.

sjlongland:
This makes things a *LOT* easier to trace when callbacks mysteriously
don't work.

Otherwise, if the callback generates an exception, you get nothing in the logs and your error disappears into a big black hole, at best you know it didn't work but you don't know why.


gmr:
Thanks

Issue #525, opened 2015-02-03
on_open_error_callback not called for ProbableAuthenticationError


awelzel:
Hi Gavin,

I was trying the following snippet and came across a potential issue with the TornadoConnection. If I provide the wrong credentials, or a non-existing virtual host in the ConnectionParameters, the on_open_error_callback is never called. Further, the ioloop is not terminated upon such an error either.

With the following snippet (very simplified compared to the actual code), I had expected the _on_open_error_cb to trigger in case the connection fails and it would be possible to cleanly handle the issue there. Maybe I'm missing something, but I currently see no way to handle the connection error cleanly.

I see the same behavior with the SelectConnection - the error callback is not called. However, in that case, the exception bubbles through start() and the program terminates. This is better, but I'd like to have used the tornado loop elsewhere.

Is this the expected behavior? Otherwise do you have pointers for a workaround or maybe a fix?

Thanks,
   Arne

```
import pika

parameters = pika.ConnectionParameters(
    virtual_host="notexistent",
    credentials=pika.PlainCredentials("user", "pass")
)

def _on_channel_open(channel):
    print "----------- GOT A CHANNEL"

def _on_open_cb(conn):
    print "----------- OPEN"
    conn.channel(on_open_callback=_on_channel_open)

    
def _on_open_error_cb(conn):
    print "----------- OPEN ERROR"
    conn.close()

def _on_close_cb(conn):
    print "----------- OPEN ERROR"
    conn.close()


conn = pika.TornadoConnection(
    parameters=parameters,
    on_open_callback=_on_open_cb,
    on_open_error_callback=_on_open_error_cb,
    on_close_callback=_on_close_cb,
    stop_ioloop_on_close=True
)

try:
    conn.ioloop.start()
except KeyboardInterrupt:
    pass
```

awelzel:
This is with pika 0.9.14.

awelzel:
PR #672 fixes this.

PR #526, opened 2015-02-03
on_open_error_callback not called for ProbableAuthenticationError

awelzel:
Hi Gavin,

I don't have much insight into all the interactions of the code, but the following seems to do the right thing for the TornadoConnection and should not change the existing behavior (except for calling the on_open_error_callbacks).

Thoughts?

Thanks,
   Arne

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -134,9 +134,23 @@ class BaseConnection(connection.Connection):
         if self.socket:
             self.socket.close()
         self.socket = None
-        self._check_state_on_disconnect()
-        self._handle_ioloop_stop()
-        self._init_connection_state()
+
+        try:
+            self._check_state_on_disconnect()
+        except exceptions.AMQPConnectionError as e:
+            # If the connection is neither open, not closed, it probably
+            # was never successfully established.
+            if not self.is_open and not self.is_closed:
+                self.callbacks.process(0, self.ON_CONNECTION_ERROR,
+                                       self, self, repr(e))
+
+            # reraise this exception to keep existing behavior.
+            raise
+
+        finally:
+            # Make sure we cleanup
+            self._handle_ioloop_stop()
+            self._init_connection_state()
 
     def _check_state_on_disconnect(self):
         """Checks to see if we were in opening a connection with RabbitMQ when
--- a/pika/adapters/tornado_connection.py
+++ b/pika/adapters/tornado_connection.py
@@ -3,6 +3,7 @@ from tornado import ioloop
 import logging
 import time
 
+from pika import exceptions
 from pika.adapters import base_connection
 
 LOGGER = logging.getLogger(__name__)
@@ -72,7 +73,10 @@ class TornadoConnection(base_connection.BaseConnection):
         """Disconnect from the RabbitMQ broker"""
         if self.socket:
             self.ioloop.remove_handler(self.socket.fileno())
-        super(TornadoConnection, self)._adapter_disconnect()
+        try:
+            super(TornadoConnection, self)._adapter_disconnect()
+        except exceptions.AMQPConnectionError as e:
+            LOGGER.warning("TornadoConnection._adapter_disconnect: %s", repr(e))
 
     def add_timeout(self, deadline, callback_method):
         """Add the callback_method to the IOLoop timer to fire after deadline


awelzel:
#525

gmr:
Have you examined the impact that your change to BaseConnection will have on all of the connection adapters?

pika/adapters/tornado_connection.py
```diff
@@ -72,7 +73,10 @@ def _adapter_disconnect(self):
         """Disconnect from the RabbitMQ broker"""
         if self.socket:
             self.ioloop.remove_handler(self.socket.fileno())
-        super(TornadoConnection, self)._adapter_disconnect()
+        try:
+            super(TornadoConnection, self)._adapter_disconnect()
+        except exceptions.AMQPConnectionError as e:
```
vitaly-krugl: Is the exception being intentionally suppressed here? Please add a comment explaining why the exception is being suppressed and how user code would become informed of the failure.

awelzel:
> Have you examined the impact that your change to BaseConnection will have on all of the connection adapters?

Not in too close detail as they all use ```_adapter_disconnect()``` from the base connection. I have tested the ```SelectConnection``` and confirmed it is still working and will indeed call ```on_open_error_callback``` when provided.

 A possibly user visible change is that for errors that happen after ```adapter_connect()```, but before ```_on_connection_open()```, a call to the provided ```on_open_error_callback``` will now be done for all adapters (see below for twisted). But, that seems to be the right thing to do and the reason for #525 to begin with.

Further, ```_handle_ioloop_stop``` and ```_init_connection_state``` will be properly called in above scenario.

The default implementation (```_on_connection_error```) will simply raise an ```exceptions.AMQPConnectionError``` as well, so this will keep the behavior.

The twisted connection uses its own ```_adapter_disconnect()``` so it's not affected. However, I think it has the same problem - ```on_open_error_callback``` won't be called for "probably auth" errors.

pika/adapters/base_connection.py
```diff
@@ -134,9 +134,23 @@ def _adapter_disconnect(self):
         if self.socket:
             self.socket.close()
         self.socket = None
-        self._check_state_on_disconnect()
-        self._handle_ioloop_stop()
-        self._init_connection_state()
+
+        try:
+            self._check_state_on_disconnect()
+        except exceptions.AMQPConnectionError as e:
```
vitaly-krugl: The title of the PR refers to ProbableAuthenticationError, but the exception being trapped here is more general.

pika/adapters/tornado_connection.py
awelzel: @vitaly-krugl , thanks for checking. The commit message had the reason:

>  tornado_connection: silence AMQPConnectionError in _adapter_disconnect
> It's not possible to handle this exception anywhere and it's super noisy.
18de5f8fcf1c25877522e3ed37404d13795fbc60

AFAICT, if a ```AMQPConnectionError``` happens at that point inside a tornado callback, you as a user can't catch it anywhere reasonably well and it.

However - I recently came across the following messages which seem to be caused by this change (in the scenario of #525 ):
```
CRITICAL:pika.adapters.base_connection:Tried to handle an error where no error existed
```
Seems some other code is relying for the exception to happen so it's not executed... :-/

pika/adapters/base_connection.py
awelzel: > The title of the PR refers to ProbableAuthenticationError, but the exception being trapped here is more general.

Yes, that is true. I observed the problem only with the ```ProbableAuthenticationError```. IMO, however, the ```on_open_error_callback``` should be called for any connection errors in cases where the connection has not been fully opened.

pika/adapters/tornado_connection.py
vitaly-krugl: I can see how an exception here is a problem for frameworks like Tornado and Twisted that rely on continuation-passing style. I am not aware of an ideal solution for this in the existing pika code. Does TornadoConnection's `stop_ioloop_on_close=True` cause the ioloop to stop in the #525 scenario? I see that there are other types of exceptions that would be problematic for these frameworks, such as exceptions.ProtocolVersionMismatch.

awelzel: For torando I think only after applying b999636b4f1371268896840090a635a32fa4f980 - the code of the ```finally``` clause was skipped previously.

vitaly-krugl: I wonder if the following might do the trick for you (in addition to suppressing AMQPConnectionError in tornado_connection.py):

1. Add try/finally in `BaseConnection._handle_disconnect()`:
 ```
    def _handle_disconnect(self):
        """Called internally when the socket is disconnected already
        """
        try:
            self._adapter_disconnect()
        finally:
            self._on_connection_closed(None, True)
```
2. Pass `on_close_callback` callback function to the TornadoConnection constructor.

I think this will assure that your `on_close_callback` will be called if AMQP handshake fails. The `on_open_error_callback` callback should be called if connection setup (TCP/IP connection or SSL handshake) fails prior to AMQP handshake.

awelzel: I used ```on_open_error_callback``` because at that point, the state of the connection is not in ```CONNECTION_OPEN``` yet. Also, I think you do not get a reason/message from the broker. The connection essentially times out. That's why ```ProbableAuthenticationError```.

I leave this decision to @gmr .

vitaly-krugl: @awelzel, the reason I proposed the try/finally `BaseConnection._handle_disconnect()` fix above is that it doesn't alter the documented semantics of `on_close_callback` and `on_open_error_callback`, shouldn't negatively impact any other connection classes, and is also deterministic versus "If the connection is neither open, not closed, it \*probably\* was never successfully established".

CC @gmr

vitaly-krugl:
@awelzel, it looks like the issue has been addressed by the merged PR #672. Could you please close this PR? Thx!

CC @gmr

awelzel:
@vitaly-krugl , I have tested that PR #672 fixes the issue. This PR is not longer relevant.

vitaly-krugl:
Thank you.

PR #529, opened 2015-03-10
Add correct stack trace on AttributeError

thomdixon:
When an `AttributeError` occurs in a callback, it is swallowed by the surrounding `try` block and a `ConnectionClosed` is raised. This hides the underlying exception, making debugging more difficult. In this case, we add the stack trace from the original exception to show the true source of the error.

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -14,6 +14,7 @@ import os
 import logging
 import select
 import socket
+import sys
 import time
 import warnings
 import errno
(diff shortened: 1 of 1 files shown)


gmr:
Seems reasonable, I'll take a look ASAP.

wjps:
Having been bitten by this, I wonder why is AttributeError being treated specially here?

I assume it's trying to catch some specific condition when something hasn't been assigned in BlockingConnection (I'm guessing _read_poller?) In which case isn't the correct fix to detect that specific case and avoid catching all AttributeErrors raised in user code?

gmr:
I couldn't raise a condition in testing where AttributeError was being caught and removed the except block for it in a previous commit. Thanks for the PR :)

PR #536, opened 2015-03-29
Removes unused self.fd attribute from BaseConnection

vitaly-krugl:


gmr:
Thanks for the PR, I'll take a look ASAP.

vitaly-krugl:
Will open another one, hopefully coveralls will be in a better mood

PR #537, opened 2015-03-30
Fix #535, allowing the expected `reply_code` and `reply_text` to be passed to `on_close_callback`

### LINKED ISSUE #535: Connection._on_connection_closed() loses reply_code and reply_text before calling _on_disconnect
`Connection._on_connection_closed` resets `reply_code` and `reply_text` and ends up calling _on_disconnect with `reply_code=0`, losing the code that was actually passed in.

This negatively impacts `SelectConnection`, among others, by causing wrong information to be passed to user's `on_close_callback` function that was registered via `SelectionConnection` constructor.
```
    def _on_connection_closed(self, method_frame, from_adapter=False):
        if method_frame and self._is_connection_close_frame(method_frame):
            self.closing = (method_frame.method.reply_code,

        # Stop the heartbeat checker if it exists
        if self.heartbeat:
            self.heartbeat.stop()

        # If this did not come from the connection adapter, close the socket
        if not from_adapter:
            self._adapter_disconnect()

        # Invoke a method frame neutral close
        self._on_disconnect(self.closing[0], self.closing[1])
```

`BasicConnection._adapter_disconnect()` calls `Connection._init_connection_state()`, which resets `self.closing = 0, 'Not specified`, so by the time that `self._on_disconnect(self.closing[0], self.closing[1])` is called, the values of `self.closing[0] and self.closing[1]` have already been lost, causing `self._on_disconnect` to be called with incorrect, unintended values. This results in the wrong information being passed to the user's `on_close_callback`.

vitaly-krugl:
Pass expected reply_code and reply_text from method frame to Connection._on_disconnect from Connection._on_connection_closed. Fixes #535

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -1188,6 +1188,9 @@ class Connection(object):
             self.closing = (method_frame.method.reply_code,
                             method_frame.method.reply_text)
 
+        # Save the codes because self.closing gets reset by _adapter_disconnect
+        reply_code, reply_text = self.closing
+
         # Stop the heartbeat checker if it exists
         if self.heartbeat:
             self.heartbeat.stop()
@@ -1197,7 +1200,7 @@ class Connection(object):
             self._adapter_disconnect()
 
         # Invoke a method frame neutral close
-        self._on_disconnect(self.closing[0], self.closing[1])
+        self._on_disconnect(reply_code, reply_text)
 
     def _on_connection_error(self, connection_unused, error_message=None):
         """Default behavior when the connecting connection can not connect.


vitaly-krugl:
This fixes #535. Need to re-run the build, as it failed while github was under severe DDoS attack.

gmr:
Thanks for the PR, I'll take a look ASAP.

vitaly-krugl:
@gmr, I removed `blocking_connection.py` changes from this PR that overlapped/collided with changes in PR #542.

robochat:
There's also another problem with this part of the code though. I always see a warning message from the `BaseConnection._adapter_disconnect()` method (actually  `_check_state_on_disconnect()`), that says
```
       WARNING -  ... -  Unknown state on disconnect: 6
```
connection_state = 6 meaning CONNECTION_CLOSING which makes sense since  `_adapter_disconnect()` is called before `Connection._on_disconnect()` (which sets the connection_state as closed). It almost seems like in `_on_connection_closed()` we should be disconnecting the adapter after we run `_on_disconnect()` ie.
```
        # Invoke a method frame neutral close
        self._on_disconnect(self.closing[0], self.closing[1])
        
        # If this did not come from the connection adapter, close the socket
        if not from_adapter:
            self._adapter_disconnect()
```
But there are probably issues with swapping the order of the calls that I'm not aware of.

vitaly-krugl:
Thanks @robochat, I will see about cleaning up the warning.

gmr:
Will merge with the intention of doing a full regression test before releasing.

vitaly-krugl:
@robochat, would you mind filing your feedback about the warning "Unknown state on disconnect: 6" as an Issue in pika? Thx!

Issue #538, opened 2015-03-30
SelectConnection fails with socket.timeout most of the time when publishing a large number of large messages


vitaly-krugl:
Both RabbitMQ broker and my test are running on my 2013 MacBookPro i7 with 16GB RAM.

Publishing a large number of large message via SelectConnection fails with socket.timeout most of the time. I implemented an app that reproduces this failure. It's on github at https://github.com/vitaly-krugl/pika-perf

Here is the command that reproduces the failure on my system:
```
python pika_perf.py publish --impl SelectConnection --exg test --size=50000 --msgs=100000
```

If it succeeds, repeat execution of the command in rapid succession. It usually fails for me the first time, though, and often with smaller messages sizes (e.g., 20000)
```
Traceback (most recent call last):
  File "pika_perf.py", line 187, in <module>
    main()
  File "pika_perf.py", line 92, in main
    messageSize=options.messageSize)
  File "pika_perf.py", line 183, in runSelectPublishTest
    connection.ioloop.start()
  File "/Users/vkruglikov/ossdev/pika/pika/adapters/select_connection.py", line 138, in start
    self.poller.start()
  File "/Users/vkruglikov/ossdev/pika/pika/adapters/select_connection.py", line 373, in start
    self.poll()
  File "/Users/vkruglikov/ossdev/pika/pika/adapters/select_connection.py", line 398, in poll
    self._handler(self.fileno, events, write_only=write_only)
  File "/Users/vkruglikov/ossdev/pika/pika/adapters/base_connection.py", line 322, in _handle_events
    self._handle_read()
  File "/Users/vkruglikov/ossdev/pika/pika/adapters/base_connection.py", line 351, in _handle_read
    self._on_data_available(data)
  File "/Users/vkruglikov/ossdev/pika/pika/connection.py", line 1288, in _on_data_available
    self._process_frame(frame_value)
  File "/Users/vkruglikov/ossdev/pika/pika/connection.py", line 1355, in _process_frame
    if self._process_callbacks(frame_value):
  File "/Users/vkruglikov/ossdev/pika/pika/connection.py", line 1325, in _process_callbacks
    frame_value)                 # Args
  File "/Users/vkruglikov/ossdev/pika/pika/callback.py", line 61, in wrapper
    return function(*tuple(args), **kwargs)
  File "/Users/vkruglikov/ossdev/pika/pika/callback.py", line 92, in wrapper
    return function(*args, **kwargs)
[...]

PR #545, opened 2015-04-12
write starvation fixes

wjps:
Hi, this contains some changes to the way writes are sent in pika that addresses some issues we were seeing when using the tornado adapter. There's more detail in the individual commit messages but the basic premise is that we should get the data out on the wire as quickly as possible. Unfortunately the use sendall() and timeouts isn't really suited to that way of working so the socket has been changed to non-blocking (after the connect sequence) and we simply handle the write errors when the socket buffer is full.

The changes seem to be generally applicable to other connection adapters and in fact there seems to be a bug with BlockingConnections on master that is fixed by applying these changes.

I've load tested the various adapters and In general I see much more consistent and reliable behaviour now (no random timeouts, pipeline stalls etc).

Cheers,

Will

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -7,7 +7,6 @@ import errno
 import logging
 import socket
 import ssl
-import collections
 
 from pika import connection
 from pika import exceptions
@@ -389,7 +388,6 @@ class BaseConnection(connection.Connection):
                 LOGGER.debug("Would block, requeuing frame")
                 self.outbound_buffer.appendleft(frame)
             else:
-                LOGGER.warning("Socket error: %s", errno.errorcode[error.errno])
                 return self._handle_error(error)
 
         #LOGGER.debug("wrote %s bytes", bytes_written)
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -7,6 +7,7 @@ import errno
 import logging
 import socket
 import ssl
+import collections
 
 from pika import connection
 from pika import exceptions
@@ -66,6 +67,7 @@ class BaseConnection(connection.Connection):
         self.socket = None
         self.stop_ioloop_on_close = stop_ioloop_on_close
         self.write_buffer = None
+        self.buffer_hist = None
         super(BaseConnection, self).__init__(parameters,
                                              on_open_callback,
                                              on_open_error_callback,
@@ -105,7 +107,7 @@ class BaseConnection(connection.Connection):
         self.ioloop.remove_timeout(timeout_id)
 
     def _adapter_connect(self):
-        """Connect to the RabbitMQ broker, returning True if connected
+        """Connect to the RabbitMQ broker, returning True if connected.
 
         :rtype: bool
 
@@ -123,6 +125,8 @@ class BaseConnection(connection.Connection):
         for sock_addr in addresses:
             error = self._create_and_connect_to_socket(sock_addr)
             if not error:
+                # Make the socket non-blocking after the connect
+                self.socket.setblocking(0)
                 return None
         # Failed to connect
         return error
@@ -239,7 +243,10 @@ class BaseConnection(connection.Connection):
         return None
 
     def _flush_outbound(self):
-        """Call the state manager who will figure out that we need to write."""
+        """write early, if the socket will take the data why not get it out
+        there asap.
+        """
+        self._handle_write()
         self._manage_event_state()
 
     def _handle_disconnect(self):
@@ -337,9 +344,15 @@ class BaseConnection(connection.Connection):
                 data = self.socket.read(self._buffer_size)
             else:
                 data = self.socket.recv(self._buffer_size)
+        
         except socket.timeout:
+            # We're using non-blocking sockets now so we should never
+            # get a timeout.
             raise
+        
         except socket.error as error:
+            if error.errno in (errno.EAGAIN, errno.EWOULDBLOCK):
+                return 0
             return self._handle_error(error)
 
         # Empty data, should disconnect
@@ -347,23 +360,43 @@ class BaseConnection(connection.Connection):
             LOGGER.error('Read empty data, calling disconnect')
             return self._handle_disconnect()
 
+        #LOGGER.debug("read %s bytes", len(data))
         # Pass the data into our top level frame dispatching method
         self._on_data_available(data)
         return len(data)
 
     def _handle_write(self):
-        """Handle any outbound buffer writes that need to take place."""
+        """Try and write as much as we can, if we get blocked requeue 
+        what's left"""
         bytes_written = 0
-        if self.outbound_buffer:
-            frame = self.outbound_buffer.popleft()
-            try:
-                self.socket.sendall(frame)
-                bytes_written = len(frame)
-            except socket.timeout:
-                raise
-            except socket.error as error:
+        try:
+            while self.outbound_buffer:
+                frame = self.outbound_buffer.popleft()
+                bw = self.socket.send(frame)
+                frame = frame[bw:]
+                bytes_written += bw
+
+                if frame:
+                    LOGGER.warning("Partial write, requeing remaining data")
+                    self.outbound_buffer.appendleft(frame)
+                    break
+
+        except socket.timeout:
+            # Will only come here if the socket is blocking
+            LOGGER.warning("socket timeout, requeuing frame")
+            self.outbound_buffer.appendleft(frame)
+            
+        except socket.error as error:
+            if error.errno in (errno.EAGAIN, errno.EWOULDBLOCK):
+                LOGGER.warning("Would block, requeuing frame")
+                self.outbound_buffer.appendleft(frame)
+            else:
+                LOGGER.warning("Socket error: %s", errno.errorcode[error.errno])
                 return self._handle_error(error)
-        return bytes_written
+
+        #LOGGER.debug("wrote %s bytes", bytes_written)
+        return bytes_written     
+
 
     def _init_connection_state(self):
         """Initialize or reset all of our internal state variables for a given
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -348,7 +348,8 @@ class BaseConnection(connection.Connection):
         except socket.timeout:
             # We're using non-blocking sockets now so we should never
             # get a timeout.
-            raise
+            LOGGER.warning("handle_read received a socket.timeout, ignoring")
+            pass
         
         except socket.error as error:
             if error.errno in (errno.EAGAIN, errno.EWOULDBLOCK):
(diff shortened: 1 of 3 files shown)


pika/adapters/base_connection.py
```diff
@@ -347,23 +356,43 @@ def _handle_read(self):
...
-            except socket.error as error:
+        try:
+            while self.outbound_buffer:
+                frame = self.outbound_buffer.popleft()
+                bw = self.socket.send(frame)
+                frame = frame[bw:]
+                bytes_written += bw
+
+                if frame:
+                    LOGGER.warning("Partial write, requeing remaining data")
```
vitaly-krugl: @wjps, a partial write in this scenario is to be expected and not abnormal at all, so let's not log at as a warning. debug level would be more appropriate, if it's necessary to log it at all.

pika/adapters/base_connection.py
```diff
@@ -123,6 +123,8 @@ def _adapter_connect(self):
         for sock_addr in addresses:
             error = self._create_and_connect_to_socket(sock_addr)
             if not error:
+                # Make the socket non-blocking after the connect
+                self.socket.setblocking(0)
```
vitaly-krugl: @wjps, what impact will this change have on the SSL connection?

pika/adapters/base_connection.py
```diff
@@ -347,23 +356,43 @@ def _handle_read(self):
             LOGGER.error('Read empty data, calling disconnect')
             return self._handle_disconnect()
 
+        #LOGGER.debug("read %s bytes", len(data))
```
vitaly-krugl: Let's delete this dead code

pika/adapters/base_connection.py
```diff
@@ -347,23 +356,43 @@ def _handle_read(self):
...
+                bytes_written += bw
+
+                if frame:
+                    LOGGER.warning("Partial write, requeing remaining data")
+                    self.outbound_buffer.appendleft(frame)
+                    break
+
+        except socket.timeout:
+            # Will only come here if the socket is blocking
+            LOGGER.warning("socket timeout, requeuing frame")
```
vitaly-krugl: For a blocking socket, a write timeout is to be expected and not a cause for warning. @wjps, the reason that I am pointing these out is that it's difficult to debug real problems when the log is filled with warning messages that result from normal conditions.

pika/adapters/tornado_connection.py
```diff
@@ -56,13 +59,14 @@ def __init__(self, parameters=None,
...
         """Connect to the remote socket, adding the socket to the IOLoop if
-        connected
+        connected. Set the socket non-blocking after super() call.
 
         :rtype: bool
 
         """
         error = super(TornadoConnection, self)._adapter_connect()
         if not error:
+            self.socket.setblocking(0)
```
vitaly-krugl: @wjps, what is the purpose of this `setblocking(0)` call here if your change in super `BaseConnection._adapter_connect()` already does the same?

pika/adapters/base_connection.py
```diff
@@ -347,23 +356,43 @@ def _handle_read(self):
...
-                self.socket.sendall(frame)
-                bytes_written = len(frame)
-            except socket.timeout:
-                raise
-            except socket.error as error:
+        try:
+            while self.outbound_buffer:
+                frame = self.outbound_buffer.popleft()
+                bw = self.socket.send(frame)
+                frame = frame[bw:]
```
vitaly-krugl: Consider skipping object allocation if the entire frame was written. The check should be cheaper than an allocation unless you partial frame writes will be the norm.

pika/adapters/base_connection.py
```diff
@@ -239,7 +241,10 @@ def _get_error_code(self, error_value):
         return None
 
     def _flush_outbound(self):
-        """Call the state manager who will figure out that we need to write."""
+        """write early, if the socket will take the data why not get it out
+        there asap.
+        """
+        self._handle_write()
```
vitaly-krugl: Adding a call to `_handle_write()` here shouldn't be necessary with the Twisted framework: if there is anything in `outbound_buffer`, then `_manage_event_state()` will call `IOLoopReactorAdapter.update_handler()` with the `WRITE` bit turned on, resulting in a call to `self.reactor.addWriter(self.connection)`, and a subsequent call in short order to `TwistedConnection.doWrite()` from the reactor if/when the socket is writable. I think that it would be cleaner to just let the Twisted reactor do its job naturally.

vitaly-krugl:
If I am not mistaken, the propsed change in `base_connection.py` might fix #538 as well.

pika/adapters/base_connection.py
```diff
@@ -347,23 +356,43 @@ def _handle_read(self):
...
+
+                if frame:
+                    LOGGER.warning("Partial write, requeing remaining data")
+                    self.outbound_buffer.appendleft(frame)
+                    break
+
+        except socket.timeout:
+            # Will only come here if the socket is blocking
+            LOGGER.warning("socket timeout, requeuing frame")
+            self.outbound_buffer.appendleft(frame)
```
vitaly-krugl: Previously, `socket.timeout` was re-raised in both `_handle_read()` and `_handle_write()`. This PR makes them inconsistent: re-raised in `_handle_read()`, but suppressed in `_handle_write()`.

vitaly-krugl: If `socket.timeout` exception is going to be suppressed in `_handle_write()`, it necessitates clean-up in `BlockingConnection._flush_outbound()`, which still expects to handle `socket.timeout` from `BaseConnection._handle_write()`, which becomes dead code with this change. Then there are also `BlockingConnection._handle_timeout()` and `BlockingConnection._socket_timeouts` logic bits that need to be considered.

vitaly-krugl: Personally, I like the proposed change in `_handle_write()`, but I also wonder what was pika author's intention behind `pika.connection.Properties.socket_timeout` and the corresponding handling and re-raising of the `socket.timeout` exceptions in both the original read and write paths. Was pika's author concerned about stalled connections, relying on `socket.timeout` exceptions to detect stalls and abort stalled connections? @gmr, would you mind chiming in on this? Thx!

pika/adapters/base_connection.py
wjps: Yup, fair enough.

pika/adapters/base_connection.py
wjps: The whole connect sequence including SSL handshake is still done with a blocking socket. I have a separate sequence of patches that makes the whole lot non-blocking (which I think is cleaner) but it requires quite a lot of fixing to Select and Asyncore Connections.

pika/adapters/base_connection.py
wjps: Will do

pika/adapters/base_connection.py
wjps: sure, np

pika/adapters/tornado_connection.py
wjps: Sorry, should have gone when I moved the changes to base_connection

pika/adapters/base_connection.py
wjps: I actually benchmarked that bit and couldn't see any appreciable difference, and it made the code a little neater like this.

pika/adapters/base_connection.py
wjps: Yeah, you could do, though I'd be surprised if that overhead would really be noticeable compared to everything else we're doing in python. Also if we haven't dropped back into the ioloop in a while the kernel may have cleared some space in the buffer in the meantime, and in cases where we're consuming a full queue of small messages we can go quite a long time before dropping into the ioloop (it's what lead me to take a look at this in the first place). So I'd prefer to keep it as is, might not be the absolutely most efficient but it will keep the output buffer full and avoid stalling the pipeline.

wjps:
Yup, it will fix #538 and various other intermittent failures that people who publish large volumes of messages will see. The existing code was basically giving up on a socket.timeout on write when this is actually very possible if you're generating messages faster than the network can send.

gmr:
I'll take a look at this ASAP.

pika/adapters/base_connection.py
vitaly-krugl: @wjps, I agree that 100% non-blocking would be best. From my experience in developing 100% non-blocking I/O integration with OpenSSL a few years ago (in C), I recall that there were a number of gotchas. At the time, I captured the gotchas that I discovered in code comments in https://github.com/openwebos/libpalmsocket/blob/master/src/psl_channel_fsm_crypto.c.

wjps: @vitaly-krugl, thanks. Though I think a lot of your gotchas will have to have been handled at the python wrapper layer, the interface exposed to python is reasonably straightforward. We'll see how we go.

vitaly-krugl: When tornado ioloop calls `BaseConnection._handle_events` and READ is indicated, it calls `BaseConnection._handle_read`, which will attempt to read by default at most `pika.spec.FRAME_MAX_SIZE = 131072` bytes from the socket and pass that to `Connection._on_data_available`. The pika overhead of processing this data should be pretty minimal, so processing time of incoming messages largely depends on processing by user code.

wjps: Yup sure, I'm not casting any aspersions in pika's `_handle_read` overhead.

Perhaps my original comment was too flippant. I am actually _extremely_ interested in the performance of pika, but more than that, I'm interested in getting behaviour that is reliable, consistent and what a user might reasonably expect. I think this PR gets us a step closer to that goal, AND it is actually significantly more efficient than the existing code.


Let's keep in mind that we're only talking about a situation here where the publisher is consistently generating messages faster than the network can send them, in this situation we would actually be pushing back using the backpressure detection, if it worked (I have a commit for that too).  In such a situation, taken to extreme, we're talking about swapping and potentially memory exhaustion, IMO a small additional overhead (1.4us* on my laptop) is probably not the first thing to worry about.

Furthermore, scenarios such as the following continue to break down...

Imagine you have an application that is reading a large amount of data from a non rabbit source (say a large XML document retrieved over http, data from some sort of DB or even generated internally), as it processes that data it publishes a large number of messages. In this situation the code might well not drop into the ioloop until it's completed processing of all the data. On pika/master it would never attempt to send a single message until processing of the full document was complete whereas this PR would keep putting messages in the socket buffer whenever space was available. With your suggestion we would fill the socket buffer once and then never send another message until we finished processing the document.

My aim with this PR is to get the data on the wire ASAP to minimise latency and maximise opportunities for parallel-isation.

*1.4us is the overhead on my laptop of attempting a 1byte write on a blocked socket in python, including catching and handling the generated exception.
0.5us is the overhead of an equivalent write on an unblocked socket.
4.3us is the overhead of dropping into the tornado ioloop and getting triggered when the socket becomes writeable as the existing code does, individually for EVERY frame sent.

vitaly-krugl: @wjps, thanks for the explanation/analysis

gmr:
There are quite a few changes in this PR. Please rebase down to a single commit, and reopen in a new PR.

Issue #403, opened 2013-10-07
ConnectionClosed raised if the callback function raises AttributeError


Dan-stahr:
The process_data_events in  blocking_connection.py is defined as follows : 

    def process_data_events(self):
        """Will make sure that data events are processed. Your app can
        block on this method.

        """
        try:
            if self._handle_read():
                self._socket_timeouts = 0
        except AttributeError:
            raise exceptions.ConnectionClosed()
        except socket.timeout:
            self._handle_timeout()

The ```_handle_read()``` method calls - somewhere deep - a callback specified by the pika user. However, if this piece of custom code throws AttributeError, pika still consider this error as ConnectionClosed. It's very annoying to debug AttributeErrors in such callbacks.

Therefore, it'd be nice to at least place the original exception as a cause of this ConnectionClosed error. The least hacky way would be (in my opinion) catching the error more specificall or wrapping the original error in another one, e.g. CallbackError.

thomdixon:
@gmr Ran into this today. I recommend passing `sys.exc_info()[2]` as the third argument to `raise` so the traceback more accurately reflects the source of the error until your more permanent solution is in place.

vitaly-krugl:
@gmr, this is fixed in master via e7b6d73aa2a38b7e2d4ac57e111bec9f9d518e08

PR #542, opened 2015-04-03
Fixed BlockingConnection handling of server-initiated Connection.Close as well as of sudden loss of connection with broker

vitaly-krugl:
Fixed #541 - Make sure connection state is properly reset when BlockingConnection._check_state_on_disconnect raises ConnectionClosed. This supplements the previously-merged PR #450 by getting the connection into consistent state.

Fixed #540 - fix handling of Connection.Close in BlockingConnection._on_connection_closed that was preventing ConnectionClosed from being raised.

Also, together with PR #533, this should address #412

PR #533, opened 2015-03-26
PIKA-532 Fix bug in BlockingConnection._handle_timeout that was preventing _on_connection_closed from being called when not closing

vitaly-krugl:

Issue #412, opened 2013-10-29
BlockingConnection uses wrong file descriptor and raises the wrong exception after rabbitmq broker stops


vitaly-krugl:
Here is an easy way to reproduce it:
1. Start rabbitmq broker (e.g., ```rabbitmq-server -detached```)
2. Create a BlockingConnection and channel()
3. Stop rabbitmq broker (e.g., ```rabbitmqctl stop```
4. Call ```queue_declare```: ```channel.queue_declare(queue="abc", passive=True, nowait=False)```

This results in ```error: (9, 'Bad file descriptor')``` exception. It should have been ConnectionClosed exception. Same problem occurs while using BlockingChannel.consume(), and I imagine in other cases as well.

The reason that this is really, really bad is that the app is expecting the ConnectionClosed exception in such cases and also this: consider a multithreaded app, where on thread is using pika and the second thread does something else completely unrelated to pika. Let's say, the second thread may be opening/closing files, sockets, or pipes, etc. So, while processing ```queue_declare```, pika detects that the server shut down, and BlockingConnection closes the socket in ```_on_connection_closed``` by calling ```_adapter_disconnect```. However, subsequently, ```_handle_read``` is called, and it calls ```self._read_poller.ready()```. ```self._read_poller``` still has the old file descriptor, so it polls on it. Considering that file descriptor numbers get recycled (at least in unix/linux), it's quite possible to have a race condition where the second thread opens some other file or socket with that same file descriptor just before ```self._read_poller.ready()```.

**Proposed solution**
It seems like ```_adapter_disconnect()``` should also close ```self._read_poller``` (ReadPoller would need a method for this) just to be safe, and ```BlockingConnection._handle_read``` should check for self.socket being None and raise ConnectionClosed instead of calling ```self._read_poller.ready()```. I also don't understand why ```BlockingConnection._on_connection_closed()``` doesn't just go ahead and raise ConnectionClosed in this scenario.

Also, ```BlockingConnection.process_data_events``` is doing something quite hacky: it's catching AttributeError and mapping it to ConnectionClosed. What if the AttributeError is due to a bug (e.g. typo in code) and has nothing to do with the connection being closed?

**Here is the full Traceback:**

```
    577     try:
    578       r = self._channelMgr.channel.queue_declare(
--> 579         queue=mqName, passive=True, nowait=False)
    580       return r.method.message_count == 0
    581     except pika_exceptions.ChannelClosed as e:

[...]

PR #546, opened 2015-04-13
Alternative fix for #403

wjps:
I think catching the AttributeError in process_data_events was to
try and catch situations where the read_poller hadn't yet been created.
So explicitly check for that situation in _handle_read and stop
improperly trapping AttributeErrors

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -239,8 +239,6 @@ class BlockingConnection(base_connection.BaseConnection):
         try:
             if self._handle_read():
                 self._socket_timeouts = 0
-        except AttributeError:
-            raise exceptions.ConnectionClosed()
         except socket.timeout:
             self._handle_timeout()
         self._flush_outbound()
@@ -344,6 +342,8 @@ class BlockingConnection(base_connection.BaseConnection):
         keeps track of how many frames have been written since the last read.
 
         """
+        if not hasattr(self, '_read_poller'):
+            raise exceptions.ConnectionClosed()
         if self._read_poller.ready():
             super(BlockingConnection, self)._handle_read()
             self._frames_written_without_read = 0


pika/adapters/blocking_connection.py
```diff
@@ -344,6 +342,8 @@ def _handle_read(self):
         keeps track of how many frames have been written since the last read.
 
         """
+        if not hasattr(self, '_read_poller'):
+            raise exceptions.ConnectionClosed()
```
vitaly-krugl: @wjps, this is not the right fix because the raised exception would not be consistent with the state of the connection - i.e., connection.is_closed would still evaluate to False after this. It's the same problem as with the `AttributeError` exception handler that this PR removed.

The issue is reported in #412 and I believe that PRs #533 and #542 together address this issue, while at the same time maintaining consistent connection state. I didn't remove the `AttributeError` block in those PRs, though.

wjps: @vitaly-krugl, not sure I agree on that one, it fixes the issue it sets out
to fix and I don't think it introduces any new ones. It wasn't aiming to
address any broader issues in BlockingConnection which are the subject of
your PRs.

vitaly-krugl: Hi @wjps, just to clarify. I think that my two PRs (#533 and #542) make this "AttributeError" issue go away altogether. I could no longer reproduce it with those PRs in place.

wjps: @vitaly-krugl, I think we're talking about different things. I _think_ I've tested your PRs and assuming I've pulled them correctly I don't think they address the issue.

Test case below....

```python
import pika
import logging

class PubSub:

  def __init__(self, url='amqp://guest:guest@localhost:5672/%2F'):
    self.conn = pika.BlockingConnection(pika.URLParameters(url))
    self.chan = self.conn.channel()
    # works more reliably if we make sure the q is empty
    self.chan.queue_delete(queue='testq') 
    self.chan.queue_declare(queue='testq')
    self.chan.queue_bind(queue='testq',exchange='amq.direct',routing_key='test')
    self.chan.basic_consume(self.callback,queue='testq')
    self.count = 0

  def run(self):
    while True:
        try:
            logging.info("Start consuming.....")
            self.chan.start_consuming()

        except pika.exceptions.ConnectionClosed:
            logging.error("WTF?!, my connection got closed")
            raise
       
        except Exception as e:
            logging.error("start_consuming raised: %s", e)
  
 
  def callback(self, cha, meth, header, body):
    if self.count == 0:
        self.count += 1
        logging.info("In callback, about to generate some generic exception")
        a = undefined_var
    else:
        logging.info("In callback, about to generate an AttributeError")
        a = body.some_random_attribute

  def publish(self, body):
    self.chan.basic_publish('amq.direct','test', body)

logging.basicConfig(level=logging.INFO)

p = PubSub()
p.publish('message1') 
p.publish('message2') 
p.run()
```

With the broken code, output is...

```
INFO:pika.adapters.base_connection:Connecting to ::1:5672
INFO:root:Start consuming.....
INFO:root:In callback, about to generate some generic exception
ERROR:root:start_consuming raised: global name 'undefined_var' is not defined
INFO:root:Start consuming.....
INFO:root:In callback, about to generate an AttributeError
ERROR:root:WTF?!, my connection got closed
  File "attribute_error_test.py", line 47, in <module>
    p.run()
  File "attribute_error_test.py", line 20, in run
    self.chan.start_consuming()
  File "/home/will/git/vitaly-krugl-pika/pika/adapters/blocking_connection.py", line 965, in start_consuming
    self.connection.process_data_events()
  File "/home/will/git/vitaly-krugl-pika/pika/adapters/blocking_connection.py", line 243, in process_data_events
    raise exceptions.ConnectionClosed()
pika.exceptions.ConnectionClosed
```
Should be...
```
INFO:pika.adapters.base_connection:Connecting to ::1:5672
INFO:root:Start consuming.....
INFO:root:In callback, about to generate some generic exception
ERROR:root:start_consuming raised: global name 'undefined_var' is not defined
INFO:root:Start consuming.....
INFO:root:In callback, about to generate an AttributeError
ERROR:root:start_consuming raised: 'str' object has no attribute 'some_random_attribute'
INFO:root:Start consuming.....
```

gmr:
This was addressed by removing the except AttributeError block in e7b6d73aa2a38

Issue #544, opened 2015-04-09
No Way To Check if BlockingConnection Generator Has Items


markcf:
It would be good if there was a function for the BlockingConnection.consume Generator to check if it has items before calling consume, this way a program could avoid being blocked in the function if there are no items in rabbit.

I'll gladly write a pull request for this, but I'd like to check if there's a reason to not have this behaviour first.

I'm imagining something like this:

```
def getWaitingMessages(self):
    return len(self._generator_messages)
```

vitaly-krugl:
Hi @markcf, are you thinking that you would call something like `getWaitingMessageCount()` and if it returned non-zero N you would then start a `consume()` loop and let it run only for N iterations and then break out to avoid blocking in the generator while waiting for the next messages to arrive? If so, you might end up with a poorly behaved polling-style application in terms of CPU consumption. You might be better off with something like the Twisted Connection or see if pika BlockingConnection is compatible with Gevent.

markcf:
Hello,
That's essentially what I'm thinking.

Normally I'd agree with you and use an event driven system. 

For this specific project, performance isn't a huge concern. Especially since the messages in Rabbit tend to be a bit larger, so the time spent polling an in memory list is minimal compared to the time spent processing them.

 The priorities of this system are high reliability and easy maintenance. Polling avoids a lot of potential pitfalls (timing bugs come to mind) and is much simpler to debug if something does come up. In addition, it makes log messages easier to understand and makes the system generally simpler. 

For this project both options were evaluated, and polling came out on top.

I understand this is not the general case, and if the general opinion is that such a feature shouldn't be included in order to discourage polling based systems that is reasonable.

gmr:
Merged your PR.

PR #547, opened 2015-04-14
#544 get_waiting_message_count()

markcf:
Added self.get_waiting_message_count() to the blocking_connection. This
allows for checking how many messages are waiting in the generator.

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -490,7 +490,7 @@ class BlockingChannel(channel.Channel):
                   self._on_cancelok, replies)
 
     def basic_get(self, queue=None, no_ack=False):
-        """Get a single message from the AMQP broker. Returns a set with the 
+        """Get a single message from the AMQP broker. Returns a set with the
         method frame, header frame and body.
 
         :param queue: The queue to get a message from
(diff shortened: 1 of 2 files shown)


markcf:
I'm a little bit confused about how my changes hurt coveralls testing of __init__.py, but I don't think it's a problem.

vitaly-krugl:
@markcf, I examined the pika API and found a comparable solution for your polling use case that should work well using the existing pika API:

1. Register your consumer and provide a callback via `basic_consume()`. Your callback will be called whenever pika BlockingConnection receives a message destined for this consumer.
2. If you prefer not to process incoming messages directly in the callback function, you can have the callback function simply append the message to your own queue (or collections.deque, etc.). Then, you can check the length of your own queue and process the messages as you had intended.

I am guessing that your polling application makes periodic calls to `BlockingConnection.process_data_events()` to keep the I/O and timers going.

markcf:
Yes. the code for option 2 is actually all in blocking_connection.py, since that's how the blocking_connection generator works, so I could just copy and paste out of there. 

This issue isn't particularly important for me, since I can do that or use one of a few other solutions. I just thought it would be nice for it to be included in Pika, so I tossed up a pull request.

It seems strange to me to have a generator with no way to check if there's anything in it, especially since calling it when there's nothing in it results in a tight polling loop. This function allows a user to define their own polling loop, which can be designed with their use case in mind, without them having to copy and paste significant amounts of code out of Pika or having to write their own.

vitaly-krugl:
This is what I had in mind, which adds a very tiny amount of user code:

```
messages = []

channel.basic_consume((lambda *args: messages.append(args)), queue="my_input_q")

# And when you want to process messages that arrived:
while messages:
  channel, method, properties, body = messages.pop(0)
  <process message>
```

My concern with the proposed API has to do with exposing internal implementation details of the generator and possibly might making it more difficult to maintain/evolve in the future.

markcf:
Fair. I was operating under the (perhaps false) assumption that the generator was unlikely to change in the future. If you think it has potential to change then it's probably not worth adding the function.

Thanks.

gmr:
Seems reasonable for 0.9, thanks.

Issue #548, opened 2015-04-14
Select Connection Close No such file or directory


rmb938:
So I have a thread that contains this
```
def run(self):
        self.connection = self.connect()
        self.connection.ioloop.start()
        if not self.connection.is_closed and not self.connection.is_closing:
            self.connection.close()
```

ioloop.start() blocks until it is stopped and once it is stopped then the connection is closed

However when connection.close() is called I receive this error
```
Traceback (most recent call last):
  File "/usr/lib64/python2.7/threading.py", line 811, in __bootstrap_inner
    self.run()
  File "/pub/cmt/venv/lib/python2.7/site-packages/cmtwarehouse/rabbitmq/routing/subscriber.py", line 28, in run
    self.connection.close()
  File "/pub/cmt/venv/lib/python2.7/site-packages/pika/adapters/base_connection.py", line 95, in close
    super(BaseConnection, self).close(reply_code, reply_text)
  File "/pub/cmt/venv/lib/python2.7/site-packages/pika/connection.py", line 690, in close
    self._close_channels(reply_code, reply_text)
  File "/pub/cmt/venv/lib/python2.7/site-packages/pika/connection.py", line 910, in _close_channels
    self._channels[channel_number].close(reply_code, reply_text)
  File "/pub/cmt/venv/lib/python2.7/site-packages/pika/channel.py", line 383, in close
    self.basic_cancel(consumer_tag=consumer_tag)
  File "/pub/cmt/venv/lib/python2.7/site-packages/pika/channel.py", line 180, in basic_cancel
    {'consumer_tag': consumer_tag})] if nowait is False else [])
  File "/pub/cmt/venv/lib/python2.7/site-packages/pika/channel.py", line 1046, in _rpc
    self._send_method(method_frame)
  File "/pub/cmt/venv/lib/python2.7/site-packages/pika/channel.py", line 1057, in _send_method
    self.connection._send_method(self.channel_number, method_frame, content)
  File "/pub/cmt/venv/lib/python2.7/site-packages/pika/connection.py", line 1503, in _send_method
    self._send_frame(frame.Method(channel_number, method_frame))
  File "/pub/cmt/venv/lib/python2.7/site-packages/pika/connection.py", line 1490, in _send_frame
    self._flush_outbound()
  File "/pub/cmt/venv/lib/python2.7/site-packages/pika/adapters/select_connection.py", line 73, in _flush_outbound
    self.ioloop.poller._manage_event_state()
[...]

PR #549, opened 2015-04-15
Make SelectConnection behave like a proper ioloop

wjps:
I pulled these changes out of some other work I was doing on `SelectConnection` because I saw  #548 come in and they address that. As such they're rather more wide ranging than they might need to be to simply address this issue. (Of course it's all good stuff IMO at least).  Details below...

Fixup all the select_connection pollers to allow filenos
to be added and removed like a general purpose ioloop.
This means we can create the poller at init time
and then add the socket once it gets created, which means
that timeouts can be added before socket creation. This
is needed for doing full non-blocking connect.

Also make the poll loop actually drop out at the time of
the next timer, not after a fixed interval so that we run
the timeouts at the time they were intended.

Add a mechanism to interupt the IOLoop by sending a byte down
a pipe that we're monitoring in our poller.

Can't test Kqueue but all the others seem to work.

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -2,9 +2,12 @@
 platform pika is running on.
 
 """
+import os
 import logging
 import select
+import errno
 import time
+from operator import itemgetter
 
 from pika.adapters.base_connection import BaseConnection
 
@@ -60,10 +63,16 @@ class SelectConnection(BaseConnection):
         """
         error = super(SelectConnection, self)._adapter_connect()
         if not error:
-            self.ioloop.start_poller(self._handle_events,
-                                     self.event_state,
-                                     self.socket.fileno())
+            self.ioloop.add_handler(self.socket.fileno(),
+                                    self._handle_events,
+                                     self.event_state)
         return error
+
+    def _adapter_disconnect(self):
+        """Disconnect from the RabbitMQ broker"""
+        if self.socket:
+            self.ioloop.remove_handler(self.socket.fileno())
+        super(SelectConnection, self)._adapter_disconnect()
 
     def _flush_outbound(self):
         """Call the state manager who will figure out that we need to write then
@@ -93,24 +102,14 @@ class IOLoop(object):
         :param method state_manager: The method to manage state
 
         """
-        self.poller = None
         self._manage_event_state = state_manager
+        self.setup_poller()
 
-    def add_timeout(self, deadline, callback_method):
-        """Add the callback_method to the IOLoop timer to fire after deadline
-        seconds. Returns a handle to the timeout. Do not confuse with
-        Tornado's timeout where you pass in the time you want to have your
-        callback called. Only pass in the seconds until it's to be called.
-
-        :param int deadline: The number of seconds to wait to call callback
-        :param method callback_method: The callback method
-        :rtype: str
-
+    def __getattr__(self, attr):
+        """Call through to the poller for timeout and ioloop.add/update/remove
+        methods.
         """
-        if not self.poller:
-            time.sleep(deadline)
-            return callback_method()
-        return self.poller.add_timeout(deadline, callback_method)
+        return getattr(self.poller, attr)
 
     @property
     def poller_type(self):
@@ -121,24 +120,11 @@ class IOLoop(object):
         """
         return self.poller.__class__.__name__
 
-    def remove_timeout(self, timeout_id):
-        """Remove a timeout if it's still in the timeout stack of the poller
-
-        :param str timeout_id: The timeout id to remove
-
-        """
-        self.poller.remove_timeout(timeout_id)
-
     def start(self):
         """Start the IOLoop, waiting for a Poller to take over."""
-        LOGGER.debug('Starting IOLoop')
-        self.poller.open = True
-        while not self.poller:
-            time.sleep(SelectPoller.TIMEOUT)
         self.poller.start()
-        self.poller.flush_pending_timeouts()
 
-    def start_poller(self, handler, events, fileno):
+    def setup_poller(self):
         """Start the Poller, once started will take over for IOLoop.start()
 
         :param method handler: The method to call to handle events
@@ -151,36 +137,23 @@ class IOLoop(object):
         if hasattr(select, 'epoll'):
             if not SELECT_TYPE or SELECT_TYPE == 'epoll':
                 LOGGER.debug('Using EPollPoller')
-                self.poller = EPollPoller(fileno, handler, events,
-                                          self._manage_event_state)
+                self.poller = EPollPoller(self._manage_event_state)
         if not self.poller and hasattr(select, 'kqueue'):
             if not SELECT_TYPE or SELECT_TYPE == 'kqueue':
                 LOGGER.debug('Using KQueuePoller')
-                self.poller = KQueuePoller(fileno, handler, events,
-                                           self._manage_event_state)
+                self.poller = KQueuePoller(self._manage_event_state)
         if not self.poller and hasattr(select, 'poll') and hasattr(select.poll(), 'modify'):
             if not SELECT_TYPE or SELECT_TYPE == 'poll':
                 LOGGER.debug('Using PollPoller')
-                self.poller = PollPoller(fileno, handler, events,
-                                         self._manage_event_state)
+                self.poller = PollPoller(self._manage_event_state)
         if not self.poller:
             LOGGER.debug('Using SelectPoller')
-            self.poller = SelectPoller(fileno, handler, events,
-                                       self._manage_event_state)
+            self.poller = SelectPoller(self._manage_event_state)
 
     def stop(self):
         """Stop the poller's event loop"""
         LOGGER.debug('Stopping the poller event loop')
-        self.poller.open = False
-
-    def update_handler(self, fileno, events):
-        """Pass in the events to process for the given file descriptor.
-
-        :param int fileno: The file descriptor to poll for
-        :param int events: The events to handle
-
-        """
-        self.poller.update_handler(fileno, events)
+        self.poller.stop()
 
 
 class SelectPoller(object):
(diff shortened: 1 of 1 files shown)


vitaly-krugl:
@wjps, this seems like a good strategy. On a related note, have you had experience with `selectors` in python 3.4 or its 2.x back-port https://github.com/berkerpeksag/selectors34? Could that be a good substitute/foundation for SelectConnection's own management of the ioloop pollers?

wjps:
Thanks @vitaly-krugl. No, unfortunately I've not used python 3 much at all and know nothing about `selectors`. I'll take a look if I get a moment but personally I'd be reluctant to add a dependency on an external package for 2.x. However the SelectConnection code is actually pretty simple and with a bit more cleaning up will be in reasonable shape. Do you know what the general thoughts are about all the various adapters? There seem to be rather a lot of them and they seem to have quite different behaviours, 'fixes' have obviously been applied to one but not identical code in others. I was going down the path of making them all present a standard ioloop like interface but can't help thinking I might be wasting my time with some of them.

vitaly-krugl:
@wjps, I concur that additional thought needs to be given to the large number of adapters and their structure. Fortunately, some of them have a small code footprint and get most of their functionality from BaseConnection and Connection.

I recently prototyped a new blocking-style adapter implemented solely on top of SelectConnection versus BlockingConnection, which has a fair amount of protocol logic and its own ioloop.

My prototype, called SynchronousConnection, is located in this branch https://github.com/vitaly-krugl/pika/tree/pika-synchronous. Since it's built on top of SelectConnection, it gets SelectConnection's performance benefits and a reduced maintenance/testing footprint. Mush of the API is the same as in BlockingConnection; however I took the liberty to experiment with the consumer interface there.

A recent test that I ran on my Mac (https://github.com/vitaly-krugl/pika-perf) that simply published 10000 1K messages, showed roughly the following Real Time (per `time` utility) performance characteristics:

SelectConnection: approx. 1 sec
SynchronousConnection: approx. 2 sec
BlockingConnection: approx. 4 min (most of it waiting, but kernel and user CPU time was still several times more than SelectConnection and SynchronousConnection)

wjps:
@vitaly-krugl, that's great news. That's exactly how I think the BlockingConnection should be written and one of the things I was hoping to get to once I had the async stuff working fully non-blocking with a standardised ioloop interface. 

My thought was that the blocking connection should be possible to implement on top of any of the async ioloops/connections. It hadn't progressed past an idle thought though, and I'm not even sure whether it's something you'd want to be able do. You're right, the current performance of BlockingConnection is shocking for some things, there's no reason why it shouldn't be comparable to the async adapaters as you've shown.

I presume you're looking to completely replace BlockingConnection? In which case you'd want to keep the interface backwards compatible. No harm in extending it though.

vitaly-krugl:
@wjps, yes I prototyped it with the thought of replacing BlockingConnection. At the moment, it's a functional proof of concept. I had a couple of reasons for contemplating a different consumer API versus those in BlockingConnection:
1. It seemed confusing to have multiple ways of accomplishing essentially the same thing
2. In the generator-based consumer, if you cancel a consumer, there was not an apparent mechanism for extracting the the messages that arrived for this consumer between the time that you cancel the consumer and the arrival of Basic-CancelOk.

That said, if we're to replace BlockingConnection, we would need to provide a backward-compatible API.

Before investing any more effort in it (tests and the remainder of the BlockingConnection APIs), I need to run it by @gmr to see if he would be open to this degree of change.

pika/adapters/select_connection.py
```diff
@@ -189,24 +162,27 @@ class SelectPoller(object):
...
-        self._handler = handler
-        self._timeouts = []
+        self.filenos = dict()
+        self.r_fds = set()
+        self.w_fds = set()
+        self.e_fds = set()
+        self.stopping = False
+        self._timeouts = {}
+        self._next_timeout = None
+        r_interupt, self.w_interupt = os.pipe()
```
vitaly-krugl: According to python doc, `select.select()` on Windows only works with sockets; see: https://docs.python.org/2/library/select.html

> On Windows, the underlying select() function is provided by the WinSock library, and does not handle file descriptors that don’t originate from WinSock

So, you will need to create a socket pair instead of a pipe for this in order to be compatible with all supported OS's. But again, there is a caveat: you can't use `socket.socketpair()` for this on Windows. `rabbitpy` solved it like this using its own `_socketpair()` method: https://github.com/gmr/rabbitpy/commit/314924c9485195be20129ecfe2710980e7a01a10#diff-278cd36141e045c14fba20692cbdd05fR460

vitaly-krugl: misspelling: `*_interupt` should be `*_interrupt`

pika/adapters/select_connection.py
```diff
@@ -189,24 +162,27 @@ class SelectPoller(object):
...
-        :param int events: The events to look for
         :param method state_manager: The method to manage state
 
         """
-        self.fileno = fileno
-        self.events = events
-        self.open = True
-        self._handler = handler
-        self._timeouts = []
+        self.filenos = dict()
```
vitaly-krugl: @wjps, I found the name `self.filenos` to detract from code readability. Perhaps something like `self.handlers` or `self.fileno_to_handler_map` would make it easier to follow the logic in the code.

pika/adapters/select_connection.py
```diff
@@ -290,19 +273,42 @@ def remove_timeout(self, timeout_id):
...
             self.process_timeouts()
-            self._manage_event_state()
 
+    def stop(self):
+        LOGGER.debug('Stopping IOLoop')
+        self.stopping = True
+        # Sent byte to interupt the poll loop
+        os.write(self.w_interupt, 'X')
+
+    def read_interupt(self, c, e, write_only):
```
vitaly-krugl: @wjps, would you mind documenting the params here?

vitaly-krugl: Also, the arg name `c` is non-standard. Would you mind changing it to `fileno` to follow the convention in other methods here? I was scratching my head when I saw it and had to trace the call back to its origin to figure out what `c` really was.

pika/adapters/select_connection.py
```diff
@@ -373,6 +389,19 @@ def start(self):
             self.poll()
```
vitaly-krugl: @wjps, `KQueuePoller.start()` still references `self.open` in `while self.open:`, which no one sets or even initializes any longer. Should this be `while not self.stopping:` instead, like in the other pollers?

pika/adapters/select_connection.py
```diff
@@ -290,19 +273,42 @@ def remove_timeout(self, timeout_id):
...
 
+    def stop(self):
+        LOGGER.debug('Stopping IOLoop')
+        self.stopping = True
+        # Sent byte to interupt the poll loop
+        os.write(self.w_interupt, 'X')
+
+    def read_interupt(self, c, e, write_only):
+        # read the interrupt byte
+        os.read(c, 512)
```
vitaly-krugl: Is there a possibility that this might get an OSError exception with `EWOULDBLOCK`? Should there be an exception handler that suppresses it just in case?

pika/adapters/select_connection.py
```diff
@@ -238,51 +234,38 @@ def poll(self, write_only=False):
...
+            events = 0
+            if fileno in read:
+                events |= READ
+            if fileno in write:
+                events |= WRITE
+            if fileno in error:
+                events |= ERROR
+           
+            if events:
+                handler = self.filenos[fileno]
```
vitaly-krugl: @wjps, it's conceivable that an earlier handler callback in this loop's cycle might call something that removes one of the the other handlers, so this code will break. `for fileno in self.filenos` might break, too, if `self.filenos` were modified during iteration. Same applies to the other pollers in this module. I think that you need to deal with it as follows, just to be on the safe side:

1. before the loop, save a copy: `filenos = self.filenos[:]`, then loop over the copy `for fileno in filenos:` instead of over `self.filenos`
2. Skip if `fileno not in self.filenos`

The problem might be more tricky if filenos are not sockets, but actual integer file descriptors, in which case closing one and opening another might re-use the same file descriptor number, so the check `fileno not in self.filenos` might not be enough.

Another complication might occur if an earlier handler callback (in the same loop cycle) modifies the events of interest for another fileno that was already registered in `self.filenos`, and the event that triggered is no longer of interest.

pika/adapters/select_connection.py
```diff
@@ -189,24 +162,27 @@ class SelectPoller(object):
...
         :param method state_manager: The method to manage state
 
         """
-        self.fileno = fileno
-        self.events = events
-        self.open = True
-        self._handler = handler
-        self._timeouts = []
+        self.filenos = dict()
+        self.r_fds = set()
```
vitaly-krugl: @wjps, shouldn't all the member variables here be designated as "private" or "protected"? I.e., `self._w_fds`, `self._e_fds`, `self._filenos`, etc.

pika/adapters/select_connection.py
vitaly-krugl: @wjps, there needs to be some way for `r_interupt` and `self.w_interupt` to be closed. Otherwise, and app that creates/destroys many SelectConnection instances over its lifetime will end up leaking lots of kernel resources. The pollers should probably have a `close()` method that should be called reliably; perhaps from `SelectConnection._adapter_disconnect()`:

```
 def _adapter_disconnect(self):
        """Disconnect from the RabbitMQ broker"""
        try:
            if self.socket:
                self.ioloop.remove_handler(self.socket.fileno())
            super(SelectConnection, self)._adapter_disconnect()
        finally:
            if self.ioloop:
                self.ioloop.close() # and it's implementation calls the underlying poller's `close()` method
```

pika/adapters/select_connection.py
```diff
@@ -60,11 +63,17 @@ def _adapter_connect(self):
...
-                                     self.socket.fileno())
+            self.ioloop.add_handler(self.socket.fileno(),
+                                    self._handle_events,
+                                     self.event_state)
         return error
 
+    def _adapter_disconnect(self):
+        """Disconnect from the RabbitMQ broker"""
+        if self.socket:
+            self.ioloop.remove_handler(self.socket.fileno())
```
vitaly-krugl: @wjps, there is no `remove_handler` in `self.ioloop`

vitaly-krugl:
@wjps, I reviewed the PR and found a number of issues here. Please take a look at my feedback. Thx

pika/adapters/select_connection.py
```diff
@@ -310,12 +316,28 @@ def update_handler(self, fileno, events):
...
+            else:
+                fd_set.discard(fileno)
+
+    def remove_handler(self, fileno):
+        """Remove a fileno to the set
+        
+        :param int fileno: The file descriptor
+
+        """
+        del self.filenos[fileno]
```
vitaly-krugl: This is not enough. `fileno` might still be in `self.r_fds`, etc. Need something like this:
```
self.update_handler(fileno, 0)
del self.filenos[fileno]
```

pika/adapters/select_connection.py
```diff
@@ -238,51 +234,38 @@ def poll(self, write_only=False):
...
-        if events:
-            self._handler(self.fileno, events, write_only=write_only)
+        for fileno in self.filenos:
+            events = 0
+            if fileno in read:
+                events |= READ
+            if fileno in write:
+                events |= WRITE
+            if fileno in error:
+                events |= ERROR
```
vitaly-krugl: Not sure if it's an issue overall, but thought I'd mention it: in Mac OS `select`, the 3rd fd set arg is for errors, but on linux and unix (and windows?) it's for "exceptfds", which are for out-of-band data in case of TCP/IP (but not errors)

vitaly-krugl:
@wjps, I modified one of my feedback comments and added some more.

wjps:
@vitaly-krugl, thanks for the review, been away for the last week but I'll take a look over the next couple of days.

gmr:
Glad to see this is progressing, please rebase off of master and format with [yapf](http://pypi.python.org/pypi/yapf).

wjps:
replaced with #555

PR #550, opened 2015-04-20
Update base_connection.py

bstemshorn:
I got this error on solaris 
"[Errno 9] service name not available for the specified socket type"

here is their fix
https://bugs.launchpad.net/zodb/+bug/1004513


gmr:
Thanks

Issue #543, opened 2015-04-08
mandatory=True queue is not exist ,but basic_publish return is always True.why


weinv:
Running on CENTOS 
run this code:
I set basic_publish mandatory=True,I sure queue is not exist,but basic_publish return is always true.why??

import pika

# Open a connection to RabbitMQ on localhost using all default parameters
connection = pika.BlockingConnection()

# Open the channel
channel = connection.channel()

# Declare the queue
#channel.queue_declare(queue="test", durable=True, exclusive=False, auto_delete=False)

# Send a message
if channel.basic_publish(exchange='test',
                         routing_key='test',
                         body='Hello World!',
                         properties=pika.BasicProperties(content_type='text/plain',
                                                         delivery_mode=1),
                         mandatory=True):
    print 'Message was published'
else:
    print 'Message was returned'

vitaly-krugl:
@weinv, this is to be expected. If delivery-confirmation is not enabled, the message return back to client will happen asynchronously at a slightly later time. Enable delivery confirmations via `channel.confirm_delivery()` once before you publish anything and you should get the desired outcome.

gmr:
@vitaly-krugl I sent you an email to what I believe is your correct address. If you didn't get it, can you drop me an email to gavinmroy@gmail.com?

PR #553, opened 2015-04-28
Add confirm_delivery to example

andrewmichaelsmith:
As per #543, I believe this is necessary for this to work


vitaly-krugl:
Yes, that is the right thing to do. Thanks!

PR #554, opened 2015-04-29
Remove unused self.fd attribute from BaseConnection

vitaly-krugl:


vitaly-krugl:
@gmr, I resubmitted this PR in hopes that coveralls would be better behaved this time. This PR simply deletes two lines of code, so I don't understand why coveralls is complaining.

PR #555, opened 2015-05-01
Make SelectConnection behave like an ioloop

wjps:
This is a replacement for PR#549

The code has been rebased to master, run through yapf and 
refined to take into account the review comments from 
@vitaly-krugl, I think they're all covered. To cover them fully it
required quite a lot more surgery so seemed better to create
a new PR.

Rework all the SelectConnection pollers to behave like a
standard ioloop and deal with multiple filedescriptors.

Also fix the timeout handling code so that timeouts fire when
they are scheduled to do so rather than on a periodic timer.

Add an interrupt socketpair so that a second thread can interrupt
the ioloop if required (to make it exit). That's not to say it's
the use of threads is heavily tested or recommended.

This work is required to make full non-blocking connect possible
in pika. It will also allow connections to multiple rabbitmq servers
to be handled in a single ioloop. The SelectConnection ioloop can
now also be used by other code that wants to deal with non pika
sockets using the a single generic ioloop.

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -2,9 +2,14 @@
 platform pika is running on.
 
 """
+import os
 import logging
+import socket
 import select
+import errno
 import time
+from operator import itemgetter
+from collections import defaultdict
 
 from pika.adapters.base_connection import BaseConnection
 
@@ -30,7 +35,8 @@ class SelectConnection(BaseConnection):
                  on_open_callback=None,
                  on_open_error_callback=None,
                  on_close_callback=None,
-                 stop_ioloop_on_close=True):
+                 stop_ioloop_on_close=True,
+                 custom_ioloop=None):
         """Create a new instance of the Connection object.
 
         :param pika.connection.Parameters parameters: Connection parameters
@@ -40,10 +46,11 @@ class SelectConnection(BaseConnection):
         :type on_open_error_callback: method
         :param method on_close_callback: Method to call on connection close
         :param bool stop_ioloop_on_close: Call ioloop.stop() if disconnected
+        :param custom_ioloop: Override using the global IOLoop in Tornado
         :raises: RuntimeError
 
         """
-        ioloop = IOLoop(self._manage_event_state)
+        ioloop = custom_ioloop or IOLoop()
         super(SelectConnection, self).__init__(parameters, on_open_callback,
                                                on_open_error_callback,
                                                on_close_callback, ioloop,
@@ -58,20 +65,25 @@ class SelectConnection(BaseConnection):
         """
         error = super(SelectConnection, self)._adapter_connect()
         if not error:
-            self.ioloop.start_poller(self._handle_events, self.event_state,
-                                     self.socket.fileno())
+            self.ioloop.add_handler(self.socket.fileno(), self._handle_events,
+                                    self.event_state)
         return error
 
-    def _flush_outbound(self):
-        """Call the state manager who will figure out that we need to write then
-        call the poller's poll function to force it to process events.
+    def _adapter_disconnect(self):
+        """Disconnect from the RabbitMQ broker"""
+        if self.socket:
+            self.ioloop.remove_handler(self.socket.fileno())
+        super(SelectConnection, self)._adapter_disconnect()
 
+    def _flush_outbound(self):
+        """Call the state manager who will figure out that we need to write
+            then call the poller's poll function to force it to process events.
         """
-        self.ioloop.poller._manage_event_state()
+        self._manage_event_state()
         # Force our poller to come up for air, but in write only mode
         # write only mode prevents messages from coming in and kicking off
         # events through the consumer
-        self.ioloop.poller.poll(write_only=True)
+        self.ioloop.poll(write_only=True)
 
 
 class IOLoop(object):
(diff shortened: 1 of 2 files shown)


rmb938:
+1 been waiting for this for a while

gmr:
Thanks for your work on this

pika/adapters/select_connection.py
```diff
@@ -188,23 +135,54 @@ class SelectPoller(object):
...
+    def get_interrupt_pair(self):
+        """ Use a socketpair to be able to interrupt the ioloop if called
+            from another thread. Socketpair() is not supported on some OS (Win)
+            so use a pair of simple UDP sockets instead. The sockets will be
+            closed and garbage collected by python when the ioloop itself is.
+        """
+        try:
+            return socket.socketpair()
+
+        except:
```
vitaly-krugl: @wjps, this is a bug; the code needs to use `except NameError:` explicitly here. `except:` by itself is almost always problematic for the following reasons:
1. It prevents system-level exceptions, such as `KeyboardInterrupt` (based on BaseException, not on Exception), from working properly. Note that Ctl-C as well as SIGINT both trigger the `KeyboardInterrupt` exception by default.
2. It's too broad and will catch unintended exceptions.

See except NameError

pika/adapters/select_connection.py
```diff
@@ -188,23 +135,54 @@ class SelectPoller(object):
...
+            LOGGER.debug("Using custom socketpair for interrupt")
+            read_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
```
vitaly-krugl: @wjps, this may be a bug. My understanding is that UDP does not guarantee packet delivery, possibly not even on the loopback interface.

pika/adapters/select_connection.py
```diff
@@ -188,23 +135,54 @@ class SelectPoller(object):
...
+    def read_interrupt(self, interrupt_sock, events, write_only):
+        """ Read the interrupt byte(s). We ignore the event mask and write_only
+        flag as we can ony get here if there's data to be read on our fd.
+
+        :param int interrupt_sock: The file descriptor to read from
+        :param int events: (unused) The events generated for this fd
+        :param bool write_only: (unused) True if poll was called to trigger a
+            write 
+        """
+        os.read(interrupt_sock, 512)
```
vitaly-krugl: @wjps, this may be a bug. Since `get_interrupt_pair()` configured the socket pair as non-blocking, OSError exception with errno EAGAIN or EWOULDBLOCK may be a possibility

pika/adapters/select_connection.py
```diff
@@ -217,120 +195,202 @@ def add_timeout(self, deadline, callback_method):
...
+        self._stopping = True
+        
         try:
-            read, write, error = select.select(input_fileno, output_fileno,
-                                               error_fileno,
-                                               SelectPoller.TIMEOUT)
-        except select.error as error:
-            return self._handler(self.fileno, ERROR, error)
+            # Send byte to interrupt the poll loop
+            self._w_interrupt.send('X')
```
vitaly-krugl: @wjps, is socket.error with errno EAGAIN or EWOULDBLOCK a possibility here?

I am also curious why you chose to use `os.read()` on `self._r_interrupt`, but `self._w_interrupt .send()` socket method here? Should they both use send/recv or os.read/os.write for consistency?

pika/adapters/select_connection.py
```diff
@@ -188,23 +135,54 @@ class SelectPoller(object):
...
-        self._handler = handler
-        self._timeouts = []
-        self._manage_event_state = state_manager
+        self._fd_handlers = dict()
+        self._fd_events = {READ: set(), WRITE: set(), ERROR: set()}
+        self._stopping = False
+        self._timeouts = {}
+        self._next_timeout = None
+        self._processing_fd_event_map = {}
+        self._r_interrupt, self._w_interrupt = self.get_interrupt_pair()
```
vitaly-krugl: @wjps, this change may be leaking UDP sockets, as there is no code that explicitly closes these two sockets `self._r_interrupt, self._w_interrupt`

pika/adapters/select_connection.py
wjps: Sure, np. Is probably an edge case but easily fixed.

pika/adapters/select_connection.py
wjps: No, it doesn't. But then the socketpair is basically an optimisation to
drop out of the ioloop immediately when triggered from elsewhere. It will
work for 99.99% of cases and, as long as the default timeout is kept at
around 5 secs, it's unlikely to be an issue there. I didn't want to use a
tcp socket because then we have to handle the connect/accept cycle and the
potential to block or fail there. I'd rather we just lose the optimisation.

pika/adapters/select_connection.py
wjps: Well, if the socket has been triggered as readable, there _should_ be at
least one byte to read. Theoretically I guess it could happen but then it
would either be a bug in our code or the OS.

pika/adapters/select_connection.py
wjps: No, the socket should be garbage collected by python.

Also remember we're talking about one pair ioloop, not per connection and
usually you'd only ever want one ioloop for the lifetime of the process.

pika/adapters/select_connection.py
wjps: Yes, again a fairly theoretical one, but we're still catching it. If we do get EWOULDBLOCK then it means we've sent an awful lot of unread interrupt bytes so there's not much we can do except log it.

WRT send/write. The ioloop works in terms of filedescriptors so the code is a lot cleaner simply doing read/write on those. This could be changed to os.write(self._w_interrupt.fileno(), 'X') but I don't think it adds anything. Happy to change if you feel strongly otherwise.

pika/adapters/select_connection.py
```diff
@@ -217,120 +195,202 @@ def add_timeout(self, deadline, callback_method):
...
-                del self._timeouts[i]
-                break
 
-    def start(self):
-        """Start the main poller loop. It will loop here until self.closed"""
-        while self.open:
-            self.poll()
-            self.process_timeouts()
-            self._manage_event_state()
+        self._processing_fd_event_map = fd_event_map
```
vitaly-krugl: @wjps, would the following recursive flow be possible/problematic for this logic?
1. This logic is executing and a handler is invoked with the READ event and data representing an incoming message or some other event.
2. The handler calls a user's callback (such as a `basic_consume` callback)
3. A user's callback calls some pika API, which in turn ends up in `poll()` and subsequently in `_process_fd_events`

I am trying to remember what in pika would prevent re-entry here?

pika/adapters/select_connection.py
vitaly-krugl: @wjps, in a default `SelectConnection`, it's ioloop per connection. Some apps create many short-lived connections over their execution lifetime.

pika/adapters/select_connection.py
vitaly-krugl: Thanks @wjps, I can't say I feel very strongly about it, but I find that, in general, code is easier to understand and maintain when things are consistent. :)

pika/adapters/select_connection.py
```diff
@@ -217,120 +195,202 @@ def add_timeout(self, deadline, callback_method):
...
+        
         try:
-            read, write, error = select.select(input_fileno, output_fileno,
-                                               error_fileno,
-                                               SelectPoller.TIMEOUT)
-        except select.error as error:
-            return self._handler(self.fileno, ERROR, error)
+            # Send byte to interrupt the poll loop
+            self._w_interrupt.send('X')
+        except Exception as err:
```
vitaly-krugl: @wjps, Would you mind catching a specific exception (`socket.error` in this case) and re-raising the exception if it's `not err.errno == errno.EWOULDBLOCK`? I am worried that this handler will make it more difficult to detect other unexpected errors. Thanks!

vitaly-krugl:
Sincere thanks @wjps for your work on this! I am done with my feedback.

pika/adapters/select_connection.py
wjps: So, as it stands, a read can trigger a write which could recursively drop into poll with the write_only flag set. When/if #556 is merged this nasty recursive behaviour goes away. Until then, whilst being ugly, does it cause a problem?

We'd end up overwriting self._processing_fd_event_map which would mean we'd drop out of the outer loop immediately after the inner. We'd then pop straight out of poll again with any events that hadn't been triggered. I guess if you had a fd that was always readable you could starve events on other descriptors?

Hopefully #556 will get merged shortly, if not I'll think about it some more.

wjps:
@gmr, @vitaly-krugl, my pleasure, thanks for your reviews.

pika/adapters/select_connection.py
```diff
@@ -217,120 +195,202 @@ def add_timeout(self, deadline, callback_method):
...
+            if events:
+                handler = self._fd_handlers[fileno]
+                handler(fileno, events, write_only=write_only)
+
+    def flush_outbound(self):
+        """Call the state manager who will figure out that we need to write
+            then call the poller's poll function to force it to process events.
         """
-        self.events = events
+        self._()
```
vitaly-krugl: @wjps, this call will crash. Also, this `flush_outbound()` method didn't exist prior to this PR and appears to not be used by anything internally, so should be safe to just delete this method.

pika/adapters/select_connection.py
vitaly-krugl: My feedback here is based entirely on code review, which anticipates that there might be a problem. By the way, I don't see how we would drop out of the outer/inner loops: `fd_event_map.keys()` returns a python `list` instance that will not be affected by overwriting of `self._processing_fd_event_map`, so the loops will just continue executing.
```
        self._processing_fd_event_map = fd_event_map

        for fileno in fd_event_map.keys():
```

vitaly-krugl: Also, regarding

> When/if #556 is merged this nasty recursive behaviour goes away

Sorry, if you wouldn't mind explaining it to me, I don't see how #556 makes the recursive behavior go away.

pika/adapters/select_connection.py
wjps: Whoops, that was search and replace gone awry.

pika/adapters/select_connection.py
wjps: Do you think it would be preferable to have a default ioloop that was re-used in that case? It would also mean that two SelectConnection instances would automatically use the same ioloop without having to extract it from one and pass into the other.

Wasn't from me, but I don't think you need a shutdown() here. The only references to these sockets are held by the ioloop and we don't care about potentially losing data in the socket buffer. In my tests the sockets get correctly garbage collected when the object is de-allocated.

If we add a close function then we'll need to handle situations where somebody attempts to re-use a closed ioloop.

pika/adapters/select_connection.py
wjps: WRT the fd_event_map, you're right. I was thinking we removed fds from the map as we went through, but it's the lists we modify so should be fine.

#556 removes SelectConnection._flush_outbound() which causes the recursive call.

pika/adapters/select_connection.py
vitaly-krugl: Thanks @wjps for sharing the info about garbage-collection of the sockets in this case. The problem with relying on garbage collection exclusively is that if something else happens to retain a living reference to the un-closed socket, than that kernel  resource would leak. Given your verification, let's leave it be for now.

Regarding a default loop, I think that with your latest changes, an app can do that if it wanted to. So, let's leave it as is, too.

pika/adapters/select_connection.py
```diff
@@ -81,105 +93,40 @@ class IOLoop(object):
...
-                                         self._manage_event_state)
-        if not self.poller:
+                poller = PollPoller()
+        
+        if not poller:
             LOGGER.debug('Using SelectPoller')
-            self.poller = SelectPoller(fileno, handler, events,
-                                       self._manage_event_state)
-
-    def stop(self):
```
vitaly-krugl: @wjps, this triggers a bug in `BaseConnection._handle_ioloop_stop()` when it attempts to all `self.ioloop.stop()` here https://github.com/pika/pika/commit/e03c7adbb609f91e7523057a33c05817f096acad#diff-1a5308d8b684fdafac1f89e376e8933dR209

If confirmed, then in addition to the fix, it would be great to have a test for this code path.

PR #556, opened 2015-05-01
A fix the write starvation problem that we see with tornado and pika

wjps:
This is #545 rebased to master.

As noted in the original PR this is primarily to fix a problem we see
with the tornado adapter but seems to be applicable to all adapters
so the patch addresses the issue in base_connection.

By default (in base_connection) Pika buffers all writes and only
attempts to send the next time it drops into the ioloop and detects
the socket as writable. This causes some odd behaviour in a number
of cases.
1. A process generating large numbers of messages will not actually
send them until it finishes the processing and drops into the ioloop.
2. A process that is consuming a large queue will only send messages
when it's read buffer is empty. If the messages are small this means
it may end up consuming 000s of messages for every one it manages to
publish. This behaviour stalls pipelines of processes.

SelectConnection had previously overridden _flush_outbound to force
a drop into the ioloop with a write_only flag set. This fixed
the behaviour for this connection type but left all the others broken.

This patch tries to send the data on the socket as soon as it's
generated and handles failed sends due to the socket buffer being full
by requeing the data.

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -104,7 +104,7 @@ class BaseConnection(connection.Connection):
         self.ioloop.remove_timeout(timeout_id)
 
     def _adapter_connect(self):
-        """Connect to the RabbitMQ broker, returning True if connected
+        """Connect to the RabbitMQ broker, returning True if connected.
 
         :returns: error string or exception instance on error; None on success
 
@@ -124,6 +124,8 @@ class BaseConnection(connection.Connection):
         for sock_addr in addresses:
             error = self._create_and_connect_to_socket(sock_addr)
             if not error:
+                # Make the socket non-blocking after the connect
+                self.socket.setblocking(0)
                 return None
             self._cleanup_socket()
 
@@ -259,7 +261,10 @@ class BaseConnection(connection.Connection):
         return None
 
     def _flush_outbound(self):
-        """Call the state manager who will figure out that we need to write."""
+        """write early, if the socket will take the data why not get it out
+        there asap.
+        """
+        self._handle_write()
         self._manage_event_state()
 
     def _handle_disconnect(self):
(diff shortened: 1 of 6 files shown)


wjps:
Hmm, not sure what's going on there. The tests run fine here, will take a look.

wjps:
@gmr, don't know if you have any ideas? It seems to have failed on py2.6 but I've run the tests on python2.6 over and over and cannot reproduce, I've also now set-up travis on my own repo and that's passed the same code!?

I suspect it's some timing related issue with the test and may have been triggered by the removal of the _flush_outbound in SelectConnection but, not being able to reproduce, it's hard to say either way. I can kill this PR and re-push but I'd rather not spam you with PRs either.

vitaly-krugl:
@wjps, it does sound like a race condition. Can you try to reproduce it by setting up a long-running test loop and letting it run for a day in the failing environment?

wjps:
ok there's definitely some sort of race in there, I can reproduce the failure on both master and my PR for both TestZ_PublishAndConsume and TestZ_PublishAndConsumeBig (more easily on the latter).

Sometimes it seems to take several thousand iterations to do so though, loading the machine up does appear to make it more easy to trigger..

wjps:
Ok the problem was with the tests, they were using a value in seconds rather than ms as the x-expires argument to the queue.declare. Sometimes the queue was getting deleted before the rest of the test could run. Fixed in #558

vitaly-krugl:
Looks good to me

wjps:
@gmr, any thoughts on merging this?

gmr:
There are a lot of commits in this and I'd rather not review the whole chain, any chance of a rebase flattening the PR

wjps:
sure np.

wjps:
sent a new PR #578

PR #559, opened 2015-05-05
Make SelectConnection behave like an ioloop - bug fix

wjps:
A bug fix for #555 and clean-ups following review comments.

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -163,7 +163,7 @@ class SelectPoller(object):
         try:
             return socket.socketpair()
 
-        except:
+        except NameError:
             LOGGER.debug("Using custom socketpair for interrupt")
             read_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
             read_sock.setblocking(0)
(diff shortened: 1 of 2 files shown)


pika/adapters/select_connection.py
```diff
@@ -367,11 +371,11 @@ def _process_fd_events(self, fd_event_map, write_only):
                 handler = self._fd_handlers[fileno]
                 handler(fileno, events, write_only=write_only)
 
-    def flush_outbound(self):
+    def _flush_outbound(self):
```
vitaly-krugl: @wjps, I think it would be better to simply delete this unnecessary method from SelectPoller. I believe that nothing is using it and I think that it was added anew in your recent PR.

vitaly-krugl:
Thanks for the followup.

pika/adapters/select_connection.py
wjps: No, that was existing code, the bug was that I'd accidentally renamed it in my original patch. It is removed by #556, but the other changes in there are required in order to do so and not create performance problems.

vitaly-krugl: Got it! Thank you.

pika/adapters/select_connection.py
```diff
@@ -367,11 +373,11 @@ def _process_fd_events(self, fd_event_map, write_only):
...
         """Call the state manager who will figure out that we need to write
             then call the poller's poll function to force it to process events.
         """
-        self._()
+        self._manage_event_state()
```
vitaly-krugl: @wjps, this will crash, because SelectPoller doesn't have `_manage_event_state` as a member.

If confirmed, this code path could also benefit from a test.

wjps: Err, _manage_event_state() is defined is BaseConnection()

vitaly-krugl: @wjps, but this `_flush_outbound` is inside `SelectPoller`, which is not subclassed from `BaseConnection`. I think that you may have accidentally pasted `_flush_outbound()` into `SelectPoller` class. The legitimate `_flush_outbound` is in `SelectConnection`.

wjps: Ah right, apologies, yes that is garbage. Never gets called and should just
be deleted.

PR #560, opened 2015-05-05
Python 2/3 polyglotting

ztane:
A conservative modification of the pika source code to support 2/3 from the same source. 
All tests pass without exceptions on Python 2.6, 2.7, PyPy;

<del>On Python 3.4 pyev seems to be crashing, other tests pass on Python 3.4; Python 3.3 has one extra test failure that needs yet to be sorted out.</del>

Correction: 463 tests passing on Python 2.6, 2.7, 3.3, 3.4 and Pypy 2.5.0.


everilae:
Had pyev tests pass on 3.4. Interesting that they should fail on similar setup elsewhere.

ztane:
It seems that the tox.ini is not used by the travis; need to install ordereddict for the Python 2.6 test

gmr:
This is quite a few change sets to sort through, any chance of a rebase squashing them down?

ztane:
https://github.com/pika/pika/pull/569 here as a completely flat commit

analytik:
Hi! Thanks to both of you for working on this.

@ztane - in blocking_connection.py@554, ```if isinstance(body, unicode):``` also needs to be replaced with ```if isinstance(body, unicode_type):```, plus import added. I'm just starting experimenting with RabbitMQ, Python 3.4 and pika, so I don't have any large project to test this on, but after this one fix, it works correctly as far as I can see.

everilae:
I'm not terribly proud of the "quick and dirty" str(...).encode() mantras in tests just for byte  stringing the results in PY3. Should've wrapped in a function. TODO? :)

ztane:
@analytik good point, seems that there is no test to even cover that `if`.

ztane:
I made a flat commit of this into https://github.com/pika/pika/pull/569 - and fixed the issue by @analytik there.

vitaly-krugl:
@ztane, if #569 supersedes this PR, would you mind closing this PR?

ztane:
@vitaly-krugl ah closed

PR #561, opened 2015-05-06
Deactivate heartbeats when heartbeat_interval is 0

paulinus:
This pull requests changes the behaviour of the heartbeat interval negotiation so that it is possible to deactivate heartbeats from the client.

The current behaviour is as follows
 * User sets heartbeat_interval in ConnectionParameters
 * When Connection.Tune is received, the min between the user parameter and the one sent by the server is used, but only if user parameter is different from 0

This makes it impossible to deactivate heartbeats since the parameter set by the user is ignored if it is 0.

The behaviour proposed on this PR is as follows
 * User sets heartbeat_interval in ConnectionParameters. It can set it to None if it wants to accept whatever the server proposes.
 * When Connection.Tune is received, the min between the user parameter and the one sent by the the server is used. If user parameter is None, the one from the server is used.

This enables the user to disable heartbeats by setting the heartbeat interval to 0.

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -1302,8 +1302,9 @@ class Connection(object):
                                               method_frame.method.frame_max)
         if self.params.heartbeat is None:
             self.params.heartbeat = method_frame.method.heartbeat
-        else:
-            self.params.heartbeat = min(self.params.heartbeat, method_frame.method.heartbeat)
+        elif self.params.heartbeat != 0:
+            self.params.heartbeat = self._combine(self.params.heartbeat,
+                                                  method_frame.method.heartbeat)
 
         # Calculate the maximum pieces for body frames
         self._body_max_length = self._get_body_frame_max_length()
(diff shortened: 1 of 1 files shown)


gmr:
Looks good, though the initial look has me wondering if you're ever setting self.params.heartbeat to 0. Did I miss it? How are you letting the server know what you're selecting?

paulinus:
As i understand it (i'm new to this),  server sends its preferences on with the Tune message, pika's Connection class "combine's" it with its own params, and then these are sent back to the server when sending the TuneOK message.
I've tested that this works on my setup using rabbitMQ, and it is in accordance to rabbit's doc. But i could be missing something.

gmr:
Your understanding is correct, I just didn't look at the patch in the full context. ``self.params.heartbeat`` is what is sent back and the tuning is replacing the value.  Thanks for the PR

paulinus:
Thanks for all the work! Using pika a lot here, and it is working great.

PR #563, opened 2015-05-06
Rationalise adapter acceptance tests

wjps:
Share the same test code across all adapters and make sure
we run exactly the same suite of tests using each.

Also test all of the SelectConnection pollers that are supported.


vitaly-krugl:
Excellent! Some food for thought, though:

`all_adapter_tests.py` might not be a good name, because it doesn't include all adapters; specifically, it doesn't test the `BlockingConnection` adapter. Perhaps `async_adapter_tests.py`?

PR #567, opened 2015-05-07
Fix issue #412: reset BlockingConnection._read_poller in BlockingConnection._adapter_disconnect() to guard against accidental access to old file descriptor

vitaly-krugl:


vitaly-krugl:
This PR should complete the fix for issue #412

PR #572, opened 2015-05-14
Added copyright and fixed NameError bug in forward_server.py

vitaly-krugl:


vitaly-krugl:
@gmr, I added the copyright header per https://github.com/vitaly-krugl/inetpy/blob/master/LICENSE.md, which is where the code for `forward_server.py` came from. Please let me know if this should be of a different form.

Issue #570, opened 2015-05-13
Closing SelectConnection after delivery confirmation stalls following merge of PR 555


vitaly-krugl:
On Mac OS X Yosemite:
Also reproducible on Fedora Linux running in VirtualBox on the above Mac OS X:

Known good commit from Dec 1 10:27:29 2014 -0500: Running at pika commit 409670bdbea13d11a43d3d02162d49a200e450a5, the following code terminates normally almost instantly.

Also, a good commit from Wed Apr 29 01:12:30 2015 -0400: dc9db2b089962451b7d6ab9826e917cdeb95a212

The first bad commit as verified on Fedora Linux is 7f91a684f4b8548ab18d9f2a3cd0d5f49fb60e3d from PR #555 "Make SelectConnection behave like an ioloop" dated Fri May 1 14:26:15 2015 +0100

cc @gmr, @wjps

NOTE: I don't encounter the problem if I call `runSelectPublishTest(...)` with `deliveryConfirmation=False`

Here is the code to reproduce and the debug log is at the end. The symptom is that the `runSelectPublishTest()` call doesn't return.
```
# Code from https://github.com/vitaly-krugl/amqp-perf
import logging
import pika

logging.getLogger('pika').setLevel(logging.DEBUG)

logging.basicConfig(
  level=logging.INFO,
  format='%(asctime)-15s %(name)s(%(process)s) - %(levelname)s - %(message)s',
  disable_existing_loggers=False)
g_log = logging

ROUTING_KEY = "test"

def runSelectPublishTest(implClassName,
                         exchange,
                         numMessages,
                         messageSize,
                         deliveryConfirmation):
  g_log.info("runSelectPublishTest: impl=%s; exchange=%s; numMessages=%d; "
             "messageSize=%s; deliveryConfirmation=%s", implClassName, exchange,
             numMessages, messageSize, deliveryConfirmation)

  message = "a" * messageSize

  def onDeliveryConfirmation(*args):
    # Got Basic.Ack or Basic.Nack
    pass

  def onChannelOpen(channel):
    if deliveryConfirmation:
      channel.confirm_delivery(callback=onDeliveryConfirmation)
      g_log.info("%s: enabled message delivery confirmation", implClassName)

    g_log.info("Select publishing...")
    for i in xrange(numMessages):
      channel.basic_publish(exchange=exchange, routing_key=ROUTING_KEY,
                            body=message)
    else:
      g_log.info("Published %d messages of size=%d via=%s",
[...]

PR #577, opened 2015-05-18
SelectConnection - make interrupt socketpair non-blocking

wjps:
This is a fix for issue #570.

The socketpair used for interrupting the iolop should be
non-blocking, if not it is possible for the ioloop to
block permanently if ioloop.start() is called recursively.

The issue occurs when an interrupt byte has been sent to interrupt the ioloop but, before the handler is called to perform the read, another callback causes the polling loop to be entered recursively (by calling ioloop.start()). The interrupt byte will then be read in the inner loop causing the socket to block later when a read is performed again in the outer loop.

Quite a specific sequence of calls is needed to make this happen but the intention was always for these sockets to be non-blocking, the previous code was an over-site.

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -163,17 +163,18 @@ class SelectPoller(object):
             closed and garbage collected by python when the ioloop itself is.
         """
         try:
-            return socket.socketpair()
-
-        except NameError:
+            read_sock, write_sock = socket.socketpair()
+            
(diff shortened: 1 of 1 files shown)


vitaly-krugl:
@wjps, this is ironic: the code snippet to reproduce #570 came from my performance test pika_perf.py that worked "properly" because of this bug. Once you fixed this bug, my test stopped working "correctly" :-)

PR #578, opened 2015-05-18
A fix the write starvation problem that we see with tornado and pika

wjps:
Rebase of #556 (and #545)

By default (in base_connection) Pika buffers all writes and only
attempts to send the next time it drops into the ioloop and detects
the socket as writable. This causes some odd behaviour in a number
of cases.
1. A process generating large numbers of messages will not actually
send them until it finishes the processing and drops into the ioloop.
2. A process that is consuming a large queue will only send messages
when it's read buffer is empty. If the messages are small this means
it may end up consuming 000s of messages for every one it manages to
publish. This behaviour stalls pipelines of processes.

SelectConnection had previously overridden _flush_outbound to force
a drop into the ioloop with a write_only flag set. This fixed
the behaviour for this connection type but left all the others broken.

This patch tries to send the data on the socket as soon as it's
generated and handles failed sends due to the socket buffer being full
by requeing the data.

Conflicts:
	pika/adapters/select_connection.py

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -75,16 +75,6 @@ class SelectConnection(BaseConnection):
             self.ioloop.remove_handler(self.socket.fileno())
         super(SelectConnection, self)._adapter_disconnect()
 
-    def _flush_outbound(self):
-        """Call the state manager who will figure out that we need to write
-            then call the poller's poll function to force it to process events.
-        """
-        self._manage_event_state()
-        # Force our poller to come up for air, but in write only mode
-        # write only mode prevents messages from coming in and kicking off
-        # events through the consumer
-        self.ioloop.poll(write_only=True)
-
 
 class IOLoop(object):
     """Singlton wrapper that decides which type of poller to use, creates an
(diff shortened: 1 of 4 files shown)


gmr:
Thanks :)

pika/adapters/select_connection.py
```diff
@@ -75,16 +75,6 @@ def _adapter_disconnect(self):
...
-            then call the poller's poll function to force it to process events.
-        """
-        self._manage_event_state()
-        # Force our poller to come up for air, but in write only mode
-        # write only mode prevents messages from coming in and kicking off
-        # events through the consumer
-        self.ioloop.poll(write_only=True)
-
 
 class IOLoop(object):
```
vitaly-krugl: @wjps, interestingly, after all these changes, it seems like IOLoop class is no longer needed, and could be replaced by a function that simply creates the appropriate poller.

vitaly-krugl:
@wjps and @gmr: I thought you might be glad to know that SelectConnection picked up a lot of basic_publish performance after the latest changes. Both RabbitMQ 3.5.1 and tests were running on my dev MacBookPro with SSD.

Before io loop and starvation fixes:

* Without pubacks

  ```
$ time python pika_perf.py publish --impl=SelectConnection --exg="" --msgs=100000
real	0m11.629s
user	0m7.640s
sys	0m2.741s
```

* With pubacks

  ```
$ time python pika_perf.py publish --impl=SelectConnection --exg="" --msgs=100000 --pubacks
real	0m16.863s
user	0m12.521s
sys	0m3.133s
```

After the fixes:

* Without pubacks

  ```
$ time python pika_perf.py publish --impl=SelectConnection --exg="" --msgs=100000
real	0m6.392s
user	0m4.620s
sys	0m1.723s
```

* With pubacks

  ```
$ time python pika_perf.py publish --impl=SelectConnection --exg="" --msgs=100000 --pubacks
real	0m8.519s
user	0m6.926s
sys	0m1.572s
```

pika/adapters/select_connection.py
wjps: Yeah, I did toy with that idea but decided to keep it as is in the end. Still working on trying to get all the ioloops behaving exactly the same and I think it makes sense to maintain an IOLoop class in each at that point. Not had so much time lately though.

wjps:
@vitaly-krugl, thanks for that, always nice to see some numbers.

PR #585, opened 2015-05-27
Update blocking_connection.retry_on_eintr workink on Python3

cgabard:
Update `retry_on_eintr` in `adapters.blocking_connection` to be compatible with Python >= 3.3 : 

 - [select.error](https://docs.python.org/2/library/select.html#select.error) in python 2.7
 - [select.error](https://docs.python.org/3/library/select.html#select.error) in python 3

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -44,7 +44,9 @@ def retry_on_eintr(f):
             try:
                 return f(*args, **kwargs)
             except select.error as e:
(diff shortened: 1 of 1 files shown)


vitaly-krugl:
@cgabard: blocking_connection has been completely re-implemented in master as a mostly thin wrapper around SelectConnection. This change is no longer necessary and is not compatible with the new implementation.

gmr:
Thank you for the contribution though :)

PR #587, opened 2015-06-05
Implement BlockingConnection adapter on top of SelectConnection

vitaly-krugl:
Copied synchronous_connection.py from pika-synchronous branch
Fixed pylint findings
Integrated SynchronousConnection with the new ioloop in SelectConnection
Defined dedicated message classes PolledMessage and ConsumerMessage and moved from BlockingChannel to module-global scope.
Got rid of nowait args from BlockingChannel public API methods
Signal unroutable messages via UnroutableError exception. Signal Nack'ed messages via NackError exception. These expose more information about the failure than legacy basic_publich API.
Removed set_timeout and backpressure callback methods

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -1,127 +1,164 @@
-"""The blocking connection adapter module implements blocking semantics on top
-of Pika's core AMQP driver. While most of the asynchronous expectations are
-removed when using the blocking connection adapter, it attempts to remain true
-to the asynchronous RPC nature of the AMQP protocol, supporting server sent
-RPC commands.
-
-The user facing classes in the module consist of the
-:py:class:`~pika.adapters.blocking_connection.BlockingConnection`
-and the :class:`~pika.adapters.blocking_connection.BlockingChannel`
-classes.
-
+"""BlockingConnection adapter implemented as a wrapper around SelectConnection.
+All public API calls complete synchronously without callbacks.
 """
-import os
+
+from collections import namedtuple, deque
 import logging
-import select
-import socket
 import time
-import warnings
-import errno
-from functools import wraps
 
-from pika import frame
-from pika import callback as callback_manager
-from pika import channel
+import pika.channel
 from pika import exceptions
-from pika import spec
-from pika import utils
-from pika.adapters import base_connection
-from pika.compat import dictkeys, unicode_type
-
-if os.name == 'java':
-    from select import cpython_compatible_select as select_function
-else:
-    from select import select as select_function
+import pika.spec
+# NOTE: import SelectConnection after others to avoid circular depenency
+from pika.adapters.select_connection import SelectConnection
 
 LOGGER = logging.getLogger(__name__)
 
 
-def retry_on_eintr(f):
-
-    @wraps(f)
-    def inner(*args, **kwargs):
-        while True:
-            try:
-                return f(*args, **kwargs)
-            except select.error as e:
-                if e[0] != errno.EINTR:
-                    raise
-
-    return inner
-
-
-class ReadPoller(object):
-    """A poller that will check to see if data is ready on the socket using
-    very short timeouts to avoid having to read on the socket, speeding up the
-    BlockingConnection._handle_read() method.
-
+class _CallbackResult(object):
+    """ CallbackResult is a non-thread-safe implementation for receiving
+    callback results; INTERNAL USE ONLY!
     """
-    POLL_TIMEOUT = 10
-
-    @retry_on_eintr
-    def __init__(self, fd, poll_timeout=POLL_TIMEOUT):
-        """Create a new instance of the ReadPoller which wraps poll and select
-        to determine if the socket has data to read on it.
-
-        :param int fd: The file descriptor for the socket
-        :param float poll_timeout: How long to wait for events (milliseconds)
+    def __init__(self, value_class=None):
+        """
+        :param callable value_class: only needed if the CallbackResult
+                                     instance will be used with
+                                     `set_value_once` and `append_element`.
+                                     *args and **kwargs of the value setter
+                                     methods will be passed to this class.
 
         """
-        self.fd = fd
-        self.poll_timeout = poll_timeout
-        if hasattr(select, 'poll') and os.name != 'java':
-            self.poller = select.poll()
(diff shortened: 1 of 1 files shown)


vitaly-krugl:
@gmr: this is the new BlockingConnection implemented as a client of SelectConnection that we discussed over email. I wanted to run it by you. It doesn't have the test work completed yet (coming soon), so it's not ready for merging. 

My publisher performance test shows startling performance differences between the new implementation and legacy:

These performance tests were publishing 10,000 messages 1024 bytes in size with and without publisher acknowledgments. They were executed on my MacBookPro with SSD against RabbitMQ 3.5.1 running locally.
* With pub-acks, it's 3 seconds (new) vs. 2 Minutes (legacy)
* Without pub-acks, it's 1 second (new) vs. 37 seconds (legacy)

```
New with publisher acknowledgments:
time python pika_perf.py publish --impl BlockingConnection --exg="" --msgs 10000 --pubacks
real	0m2.981s
user	0m1.632s
sys	0m0.330s

time python pika_perf.py publish --impl BlockingConnection --exg="" --msgs 10000
real	0m0.904s
user	0m0.629s
sys	0m0.230s


Old with publisher acknowledgments:
time python pika_perf.py publish --impl BlockingConnection --exg="" --msgs 10000 --pubacks
real	2m11.383s
user	0m7.180s
sys	0m1.683s

time python pika_perf.py publish --impl BlockingConnection --exg="" --msgs 10000
real	0m37.236s
user	0m2.177s
sys	0m0.763s
```

vitaly-krugl:
@gmr: with respect to API changes, you mentioned that you weren't happy with the legacy BlockingConnection API and implementation and that you were open to change. This implementation preserves many of the legacy APIs with the following exceptions:

1. Methods that take callbacks have been removed.
 * Justification: this is a blocking/synchronous interface vs. other adapters that are continuation-callback-based interfaces.
2. The `nowait` args have been removed (defaulted to False in legacy API) and are internally forced to False
 * Justification: In the blocking interface, we want to correlate a failure with the API that caused it, not get that error when calling some other API later
3. Replaced the consumer API mechanisms (there were 2 or 3 different ways to consume before) with a single, more generic event mechanism accessed via `BasicChannel.get_event()` method that also supports an optional inactivity timeout. This mechanism presently returns two types of events to the user: consumed messages and broker's consumer cancellation notices. And there is a generator iterator method that wraps around it with the semantic: `for evt in channel.read_events():`
 * Justification: in the legacy implementation, there were multiple API mechanisms for accessing consumers, including one with callbacks, and there was no support for inactivity timeouts. The new event-based semantic replaces those with a single mechanism that can be easily and naturally extended to support additional event types, such as timers, AMQP flow-control, etc.
4. While preserving the legacy `basic_publish`, also added a new `publish` method that communicates unroutable (Basic.Return) and Nack'ed messages via exceptions
 * Justification: these exceptions contain the returned messages along with the Basic.Return method that contains broker's error code and error messages. This useful diagnostic information isn't exposed by the legacy `basic_publish` API.

vitaly-krugl:
@gmr, did you intend to merge this PR in its current state?

andreas-kopecky:
Hello,

Even tho this is already merged i wanted to post some feedback here too: 

I am well aware that master is unstable - i use it still because in 0.9.14 the maximum publish rate seems to be about 300 messages/s due to some issues that have been fixed in master.

Now with this current master there is at least a few issues that come to mind:

- how does blocking connection work now? i get from the pull request description, that one is supposed to use "read_events" now since there is no start_consuming or the likes anymore... is that the only canoncial way now?
- how does one declare queues? or to be more precise: how does on REFERENCE declared queues since "declare_queue" no longer returns a queue object that allows to access the automatically generated name - it returns None to be exact.

vitaly-krugl:
@andreas-kopecky, fixes are coming rapidly, including the queue_declare: see PR #590. This PR got merged before I had the time to write all the tests, so expect some instability until the tests are all in.

andreas-kopecky:
@vitaly-krugl Thanks a lot for the quick reply. I have personally not contributed to this project so far but if i can be of any assistance please say so.

vitaly-krugl:
Thanks @andreas-kopecky for the help offer! I have some ideas about how to re-introduce some of the deprecated API's or something similar to them to ease the transition for apps migrating to 0.10.x version of pika. 

Presently, I could really use help with Acceptance tests in `blocking_adapter_test.py`. I fixed up the existing tests there in PR #590 (https://github.com/vitaly-krugl/pika/blob/new-blocking-connection/tests/acceptance/blocking_adapter_test.py). It would be great to add acceptance tests for public BlockingChannel methods (pick your favorite ones :) ). I am going to be off the grid for almost two weeks, so I am afraid I won't be able to help during that time.

vitaly-krugl:
@andreas-kopecky, please let me know which BlockChannel methods you will be intstrumenting in `blocking_adapter_test.py`, so that we don't step on each other. Thanks!

vitaly-krugl:
@andreas-kopecky: I went ahead and implemented many acceptance tests for BlockingConnection and BlockingChannel (and fixed some bugs) via PR #591. We still need acceptance tests for the following:
1. `BlockingChannel.read_events`
2. `BlockingChannel.basic_nack`
3. `BlockingChannel.cancel_consumer`
4. `BlockingChannel.basic_qos` (something more sophisticated than the one in TestPublishAndConsumeAndQos in the above-mentioned PR #591)
5. `BlockingChannel.basic_recover`
6. `BlockingChannel.basic_reject`
7. `BlockingChannel.exchange_bind/exchange_unbind`
8. `BlockingChannel.tx_select/tx_commit/tx_rollback`
9. Publishing and consuming a large message: e.g., 0.5 MBytes
10. Publishing/consuming many messages: e.g., 1000

andreas-kopecky:
@vitaly-krugl 
Hi - i was away over the weekend too. I am not sure if i should provide tests since in don't know the codebase all that well yet (just used it up until now) but i will read myself into it and keep this thread updated should i implement a few of those tests.

vitaly-krugl:
@andreas-kopecky, for acceptance tests, you just need to know the APIs at user level; implementation details are not important. Just need to call the API, and verify that the return value (or callback or other side-effect) matches the API's documentation. Here is an example of the one that I worked on over the weekend: https://github.com/pika/pika/blob/master/tests/acceptance/blocking_adapter_test.py

PR #591, opened 2015-06-06
BlockingConnection: Bug fixes; reintroduced process_data_events; implemented acceptance tests

vitaly-krugl:
Implemented Acceptance tests for most BlockingConnection and BlockingChannel publish API methods.

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -172,10 +172,7 @@ class BaseConnection(connection.Connection):
         """Close the socket cleanly"""
         if self.socket:
             try:
-                if hasattr(self.socket, 'socket'):
-                    self.socket.socket.shutdown(socket.SHUT_RDWR)
-                else:
-                    self.socket.shutdown(socket.SHUT_RDWR)
+                self.socket.shutdown(socket.SHUT_RDWR)
             except socket.error:
                 pass
             self.socket.close()
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -1,7 +1,15 @@
-"""BlockingConnection adapter implemented as a wrapper around SelectConnection.
-All public API calls complete synchronously without callbacks.
-"""
+"""The blocking connection adapter module implements blocking semantics on top
+of Pika's core AMQP driver. While most of the asynchronous expectations are
+removed when using the blocking connection adapter, it attempts to remain true
+to the asynchronous RPC nature of the AMQP protocol, supporting server sent
+RPC commands.
 
+The user facing classes in the module consist of the
+:py:class:`~pika.adapters.blocking_connection.BlockingConnection`
+and the :class:`~pika.adapters.blocking_connection.BlockingChannel`
+classes.
+
+"""
 from collections import namedtuple, deque
 import logging
 import time
@@ -136,6 +144,41 @@ class _CallbackResult(object):
             % (self._values,))
 
         return self._values
+
+
+class _TimerContext(object):
+    """Context manager for registering and safely unregistering a
+    SelectConnection ioloop-based timer
+    """
+
+    def __init__(self, duration, connection):
+        """
+        :param float duration: non-negative timer duration in seconds
+        :param SelectConnection connection:
+        """
+        assert hasattr(connection, 'add_timeout'), connection
+        self._duration = duration
+        self._connection = connection
+        self._callback_result = _CallbackResult()
+        self._timer_id = None
+
+    def __enter__(self):
+        """Register a timer"""
+        self._timer_id = self._connection.add_timeout(
+            self._duration,
+            self._callback_result.signal_once)
+        return self
+
+    def __exit__(self, *_args, **_kwargs):
+        """Unregister timer if it hasn't fired yet"""
+        if not self._callback_result:
+            self._connection.remove_timeout(self._timer_id)
+
+    def is_ready(self):
+        """
+        :returns: True if timer has fired, False otherwise
+        """
+        return self._callback_result.is_ready()
 
 
 class BlockingConnection(object):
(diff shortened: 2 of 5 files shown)


vitaly-krugl:
@ztane: would you mind taking a look at the test failure? They are passing for python 2.6 and 2.7, but failing for both 3.x. Many thanks!

vitaly-krugl:
@ztane @everilae @analytik @bkjones: Could you guys please help debug this Python 3.x test failure? Many thanks in advance. All the tests pass on Python 2.6 and Python 2.7, but several tests fail in Travis CI build on 3.3 and 3.4.

I instrumented one of the failing tests, `blocking_adapter_test.TestBasicGet`, with some extra logging. An excerpt from the test output is below. The last successful operation in this test is "ENABLED PUB-ACKS". Then the operation times out 15 seconds later as indicated by the label "TIMED OUT". The remaining steps should have completed in a fraction of a second. There are a couple of suspicious log messages from the pika.callback module, such as "Values in <class 'pika.spec.Queue.DeclareOk'> do not match for queue"
```
blocking_adapter_test: INFO: 2015-06-06 18:44:14.680085 ENABLED PUB-ACKS (start_test (blocking_adapter_test.TestBasicGet))
pika.channel: DEBUG: Adding in on_synchronous_complete callback
pika.callback: DEBUG: Added: {'only': None, 'arguments': {'queue': 'TestBasicGet_q078c11140c7c11e5b541c714b39b9a0d'}, 'one_shot': True, 'calls': 1, 'callback': <bound method Channel._on_synchronous_complete of <pika.channel.Channel object at 0x7f063cb9d910>>}
pika.channel: DEBUG: Adding passed in callback
pika.callback: DEBUG: Added: {'only': None, 'arguments': {'queue': 'TestBasicGet_q078c11140c7c11e5b541c714b39b9a0d'}, 'one_shot': True, 'calls': 1, 'callback': <bound method _CallbackResult.set_value_once of <pika.adapters.blocking_connection._CallbackResult object at 0x7f063cb9d710>>}
pika.callback: DEBUG: Processing 1:Queue.DeclareOk
pika.callback: DEBUG: Values in <class 'pika.spec.Queue.DeclareOk'> do not match for queue
pika.callback: DEBUG: Arguments do not match for {'only': None, 'arguments': {'queue': 'TestBasicGet_q078c11140c7c11e5b541c714b39b9a0d'}, 'one_shot': True, 'calls': 1, 'callback': <bound method Channel._on_synchronous_complete of <pika.channel.Channel object at 0x7f063cb9d910>>}, [<METHOD(['channel_number=1', 'frame_type=1', 'method=<Queue.DeclareOk([\'consumer_count=0\', \'message_count=0\', "queue=b\'TestBasicGet_q078c11140c7c11e5b541c714b39b9a0d\'"])>'])>]
pika.callback: DEBUG: Values in <class 'pika.spec.Queue.DeclareOk'> do not match for queue
pika.callback: DEBUG: Arguments do not match for {'only': None, 'arguments': {'queue': 'TestBasicGet_q078c11140c7c11e5b541c714b39b9a0d'}, 'one_shot': True, 'calls': 1, 'callback': <bound method _CallbackResult.set_value_once of <pika.adapters.blocking_connection._CallbackResult object at 0x7f063cb9d710>>}, [<METHOD(['channel_number=1', 'frame_type=1', 'method=<Queue.DeclareOk([\'consumer_count=0\', \'message_count=0\', "queue=b\'TestBasicGet_q078c11140c7c11e5b541c714b39b9a0d\'"])>'])>]
blocking_adapter_test: INFO: 2015-06-06 18:44:29.677356 TIMED OUT (start_test (blocking_adapter_test.TestBasicGet))

```

everilae:
The problems boil down to the good old strings vs. bytes issues. As was aptly put in the original pull request discussion, the PY3 port is very brute force and it does not in any way help.

So the issue is: when you publish strings in PY3, you get bytes out on the consumer side. When you declare queues with routing_keys, the routing_keys get converted to bytes. The tests were for example timing out because of this.

This is rather inconvenient, but we've learned to live with it for the time being. It's that or no pika at all for us, since we're committed to using PY3.

I'll create a patch on top of your work.

vitaly-krugl:
Thank you @everilae

vitaly-krugl:
@everilae, now we know that your change fixed Py3, but it broke Py2.x. I will take the next stab at it.

vitaly-krugl:
@everilae, it doesn't seem to make sense that user code (the acceptance test in this case) should be responsible for encoding

vitaly-krugl:
@everilae, both 2.x and 3.x pass now

vitaly-krugl:
@gmr: this is ready to go. All tests pass

```
Implemented Acceptance tests for most BlockingConnection and BlockingChannel publish API methods.

Fixed get_vent and cancel_consumer bugs.

Add logging in TestBasicGet to help debug test failure under Python 3.x

bytes fixes by @everilae

Removed invalid reference self.socket.socket.shutdown in base_connection.py. forward_server cleanup.

Fixed blocking_adapter_test.py to work on python 2.6/2.7 (after changes to make it work on 3.x broke 2.x)

Added a test for publishing/consume a huge message and another test for publishing/consuming many messages.

Test consumer cancellation and no_ack true/false side-effects

Fixed blocking_adapter_test.py failure on python 3.x
```

everilae:
Nah, it's not broken. It is just that ``"\"string\""`` literal in python2 is actually bytes, when in python3 it is text/string/what have you. So for the time being the way to canonicalise tests is to use ``b""``literals. The true fix would be to canonicalise those bytes and strings in the library so that if you pass text/strings in python3, it does-the-right-thing.

Keep in mind that if you use pika with python2 only as a User you will not see the difference that your tests brought up. Everything just works, because everything has and will be byte "strings". It is with python3 that you do have to actually know your data type (which is a good thing, in the long run).

vitaly-krugl:
Thanks @everilae !

vitaly-krugl:
I agree that the library needs to do the right thing. The library needs to be careful, though, not to incur unnecessary overhead of re-encoding data that is already in a compatible format.

everilae:
I think @ztane has been doing some thinking on the whole bytes/string canonicalisation thingie, but hasn't yet been able to get to it.

And yeah, no re-encoding bytes, if given. It's up to the User to know what they're up to at that point, because the library cannot (is this latin-1? utf-8? wtf-...?).

everilae:
The consumer side receiving issue is a bit trickier though, if you want to be able to send and receive text in python3. As it is now, pika does encoding when sending, but no decoding when consuming. We've sort of "worked around" this by accepting that we must talk to pika in bytes in python3 too. Still better than not having amqp at all :).

vitaly-krugl:
Perhaps these things should be symmetric and pika should just accept and return everything in bytes only. This would avoid unnecessary overhead of encoding in pika and do away with the current asymmetric situation. Since AMQP assumes the data is "bytes", only the application can possibly know what the target format should be.

vitaly-krugl:
Gavin, as I thought more about it, I concluded that changing the entire message consumer API would be too much of a burden for the many pika users who use the BlockingConnection adapter. So, I am going to bring back the legacy basic_consume/basic_cancel/consume/cancel API (but do it in a way that checks recursion) and get rid of the new `get_event` mechanism.

CC @andreas-kopecky

gmr:
:+1:

andreas-kopecky:
Actually I thought it an enticing idea to have only one canonical way of
consuming... Maybe at another time :)
Am 08.06.2015 20:20 schrieb "Gavin M. Roy" <notifications@github.com>:

vitaly-krugl:
@andreas-kopecky: I liked it, too, but I think it would be too disruptive for many other users doing the switch to pika 0.10.0. The thinking is along the lines of
> Don't let perfect be the enemy of good

vitaly-krugl:
Gavin, I think I have the legacy consumer mechanism under control now. Will work on updating tests next as I find bits of time to work on it (on vacation now)

PR #598, opened 2015-06-09
Allow passing exchange, queue and routing_key as text

everilae:
It turned out that ``queue`` argument was the reason why methods like ``queue_declare`` etc. seemed to require ``bytes`` values. This was due to the callback handling that compares frames/methods and stored arguments.

With these changes channel methods accept text (``str``) arguments in python3.

+  No new tests failed, the 5 failures (that Travis doesn't seem to produce) were present in the pika/master.

ping @ztane

### DIFF
--- a/pika/exceptions.py
+++ b/pika/exceptions.py
@@ -224,3 +224,10 @@ class RecursionError(Exception):
     Used by BlockingConnection/BlockingChannel
 
     """
+
+
+class ShortStringTooLong(AMQPError):
+
+    def __repr__(self):
+        return ('AMQP Short String can contain up to 255 bytes: '
+                '%.300s' % self.args[0])
--- a/pika/spec.py
+++ b/pika/spec.py
@@ -126,14 +126,7 @@ class Connection(amqp_object.Class):
 
         def decode(self, encoded, offset=0):
             (self.client_properties, offset) = data.decode_table(encoded, offset)
-            length = struct.unpack_from('B', encoded, offset)[0]
-            offset += 1
-            self.mechanism = encoded[offset:offset + length]
-            try:
-                self.mechanism = str(self.mechanism)
-            except UnicodeEncodeError:
-                pass
-            offset += length
+            self.mechanism, offset = data.decode_short_string(encoded, offset)
             length = struct.unpack_from('>I', encoded, offset)[0]
             offset += 4
             self.response = encoded[offset:offset + length]
@@ -142,14 +135,7 @@ class Connection(amqp_object.Class):
             except UnicodeEncodeError:
                 pass
             offset += length
-            length = struct.unpack_from('B', encoded, offset)[0]
-            offset += 1
-            self.locale = encoded[offset:offset + length]
-            try:
-                self.locale = str(self.locale)
-            except UnicodeEncodeError:
-                pass
-            offset += length
+            self.locale, offset = data.decode_short_string(encoded, offset)
             return self
 
         def encode(self):
@@ -157,9 +143,7 @@ class Connection(amqp_object.Class):
             data.encode_table(pieces, self.client_properties)
             assert isinstance(self.mechanism, str_or_bytes),\
                    'A non-string value was supplied for self.mechanism'
-            value = self.mechanism.encode('utf-8') if isinstance(self.mechanism, unicode_type) else self.mechanism
-            pieces.append(struct.pack('B', len(value)))
-            pieces.append(value)
+            data.encode_short_string(pieces, self.mechanism)
             assert isinstance(self.response, str_or_bytes),\
                    'A non-string value was supplied for self.response'
             value = self.response.encode('utf-8') if isinstance(self.response, unicode_type) else self.response
@@ -167,9 +151,7 @@ class Connection(amqp_object.Class):
             pieces.append(value)
             assert isinstance(self.locale, str_or_bytes),\
                    'A non-string value was supplied for self.locale'
-            value = self.locale.encode('utf-8') if isinstance(self.locale, unicode_type) else self.locale
-            pieces.append(struct.pack('B', len(value)))
-            pieces.append(value)
+            data.encode_short_string(pieces, self.locale)
             return pieces
 
     class Secure(amqp_object.Method):
@@ -311,22 +293,8 @@ class Connection(amqp_object.Class):
             return True
 
         def decode(self, encoded, offset=0):
-            length = struct.unpack_from('B', encoded, offset)[0]
-            offset += 1
-            self.virtual_host = encoded[offset:offset + length]
-            try:
-                self.virtual_host = str(self.virtual_host)
-            except UnicodeEncodeError:
-                pass
-            offset += length
-            length = struct.unpack_from('B', encoded, offset)[0]
-            offset += 1
-            self.capabilities = encoded[offset:offset + length]
-            try:
-                self.capabilities = str(self.capabilities)
-            except UnicodeEncodeError:
-                pass
-            offset += length
+            self.virtual_host, offset = data.decode_short_string(encoded, offset)
+            self.capabilities, offset = data.decode_short_string(encoded, offset)
             bit_buffer = struct.unpack_from('B', encoded, offset)[0]
             offset += 1
             self.insist = (bit_buffer & (1 << 0)) != 0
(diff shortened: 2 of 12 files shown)


tests/acceptance/async_adapter_tests.py
```diff
@@ -2,9 +2,18 @@
 import uuid
 
 from pika import spec
+from pika import compat
 
 from async_test_base import (AsyncTestCase, BoundQueueTestCase, AsyncAdapters)
 
+
+def _p3_as_bytes(value):
```
vitaly-krugl: @everilae: Instead of copying this function from blocking_adapter_tests.py, let's have a single shared version instead. Perhaps it could be a function in the pika.compat module?

everilae: True, though actually the method in itself is a bit pointless, since ``as_bytes`` does what it does and as such the whole _p3_as_bytes -> as_bytes. I'll replace.

vitaly-krugl: Good point, I took another look at `as_bytes`, and it makes sense to get rid of `_p3_as_bytes` and use `as_bytes` instead because of the bytes/str equivalence in python 2.x.

tests/acceptance/blocking_adapter_test.py
```diff
@@ -476,13 +469,13 @@ def test(self):
         self.assertIsInstance(method, pika.spec.Basic.GetOk)
         self.assertEqual(method.delivery_tag, 1)
         self.assertFalse(method.redelivered)
-        self.assertEqual(method.exchange, _p3_as_bytes(''))
-        self.assertEqual(method.routing_key, q_name)
+        self.assertEqual(method.exchange, as_bytes(''))
+        self.assertEqual(method.routing_key, as_bytes(q_name))
```
vitaly-krugl: @everilae: the asymmetry still doesn't make sense to me. It's unintuitive that I may pass a string for q_name, but it would come back as bytes in the result and that the two would not compare equal in py3.x. I understand that bytes and str are not equivalent in 3.x, but this asymmetry in pika seems rather error-prone to me. For the life of me, I can't seem to find the definitions of basic AMQP types such as `shortstr`. Would you mind pointing me to where they are defined? Is `shortstr` defined as UTF8 or just binary bytes in AMQP?

everilae: Here there is no visible asymmetry: method is a protocol class instance and never available for users (as much as anything can be made unavailable in python). This test asserts that the implementation detail of method containing attributes with certain values holds.

I confess not being well-versed in AMQP either, but: Pika did unicode to utf-8 bytes conversions in pika/spec.py before we started with the py3 changes. This is why ``as_bytes`` does make sense here.

Please have a look at how utils/codegen.py generates "string handling" for protocol methods. It was and still is unicode aware.

vitaly-krugl: @everilae: according to the RabbitMQ folks on this page https://www.rabbitmq.com/blog/2014/04/02/breaking-things-with-rabbitmq-3-3/:
 > The AMQP 'shortstr' data type (which is used for things like exchange and queue names, routing keys and so on) is defined by the spec as being in UTF-8 format.

If they are correct, it means that all of these values should be going in as UTF-8 strings (str would be okay in python2.x, too) and combing back as UTF-8 strings, but not as bytes.

CC @gmr

vitaly-krugl: @everilae: regarding
 > method is a protocol class instance and never available for users

This statement is incorrect. A number of pika functions and callbacks return the method to the user. For example, `basic_consume`'s consumer callback is explicitly called with the args `channel, method, properties, body`. Same is true for `basic_get`, etc.

everilae: From AMQP 0-9-1:

4.2.5.3 Strings
AMQP strings are variable length and represented by an integer length followed by zero or more octets of data. AMQP defines two native string types:

* Short strings, stored as an 8-bit unsigned integer length followed by zero or more octets of data. Short strings can carry up to 255 octets of UTF-8 data, but may not contain binary zero octets.
* Long strings, stored as a 32-bit unsigned integer length followed by zero or more octets of data. Long strings can contain any data.

everilae: Yeah, forgot those callbacks receiving ``method``. That's a bit evil, I agree.

Strings are certainly not "UTF-8" in python3. They are unicode codepoints. The internal encoding varies on string to string basis and cannot be relied upon / used in any meaningful manner.

"Strings" in python2 are byte data with unknown encoding. Usually it's ASCII.

vitaly-krugl: Thanks @everilae: I was looking for `shortstr` in that doc, and came up empty-handed. this is perfect! Since shortstr are UTF-8 and objects such as queue names, exchange names, routing keys are defined as shortstr, pika should accept strings for any shortstr fields from the user and return unicode strings for shortsr fields back to the user. There is not need for the bytes there at all, right?

everilae: That should be possible.

vitaly-krugl: Thanks for entertaining this thought. It should be much more robust to deal with all `shortstr` types as strings (or unicode strings). This would make equality tests work intuitively for those fields. (http://stackoverflow.com/questions/1864701/convert-utf-8-octets-to-unicode-code-points)

Only `longstr` should be bytes per AMQP definition that you sited. I am guessing that `body` is `longstr`, right?

everilae: From what I've been able to learn up till now, I think it is and as such your statement should be correct. There is one interesting thing though: pika seems to be able to pass properties in the message header for the body, such as "content-type". Now if one were to choose a standard encoding for unicode and send this information as the content-type property...

vitaly-krugl: "content-type" is for application use. Pika probably shouldn't get involved with that level of encoding/decoding ?

everilae: True, though would it be possible to have it in a double role of sorts, something like

- bytes -> do what it does at the moment
- something with application provided properties -> do what it does at the moment
- text (unicode codepoints) given, no application provided properties -> encode and provide proper content-type allowing magic decoding at consumer end

Just a thought, might as well keep it as is.

everilae: Naturally all such ideas should be backwards compatible, which is easy, since no python3 support and text vs. binary data issues have existed (sort of), since in python2 people usually pass around mystery binary blobs. In such environment py2 would continue to handle short strings and long strings as binary data and py3 would handle text to data conversions.

everilae:
@vitaly-krugl: added short-strings-as-text-in-python3, python2-does-what-it-did-before. How does it look? Moved some data handling from codegen.py/spec.py to pika/data.py. Seemed logical.

pika/exceptions.py
```diff
@@ -169,3 +169,9 @@ class InvalidMaximumFrameSize(ProtocolSyntaxError):
 
     def __repr__(self):
         return 'AMQP Maximum Frame Size is 131072 Bytes'
+
+
+class ShortStringTooLong(AMQPError):
+
+    def __repr__(self):
+        return 'AMQP Short String can contain up to 255 bytes'
```
vitaly-krugl: To enhance debugging, I think it would be good to print the first few hundred bytes of the value at `args[0]`, e.g.,

`return 'AMQP Short String can contain up to 255 bytes: %.300s' % self.args[0]` or
`return 'AMQP Short String can contain up to 255 bytes: %.300s' % self.args`

(working on my first code review from 35000 feet, compliments of JetBlue Beta WiFly :) )

pika/spec.py
```diff
@@ -126,14 +126,7 @@ def synchronous(self):
...
             (self.client_properties, offset) = data.decode_table(encoded, offset)
-            length = struct.unpack_from('B', encoded, offset)[0]
-            offset += 1
-            self.mechanism = encoded[offset:offset + length]
-            try:
-                self.mechanism = str(self.mechanism)
-            except UnicodeEncodeError:
-                pass
-            offset += length
+            self.mechanism, offset = data.decode_short_string(encoded, offset)
```
vitaly-krugl: Was the old code machine-generated? If it was, then future re-generation will result in loss of mods.

vitaly-krugl:
+1

Very clean!

vitaly-krugl:
@everilae: thanks for all the hard work.

everilae:
Will have a go at the exception str handling + data table keys decoding with short string decoder -> text keys (this PR is getting a bit out of hand...)

everilae:
Rewriting pika/data.py at fb9616ef5f766a2f2250a388735358a2e8121e34, but pushed that stuff to a separate branch in order to avoid bloating this PR even more. Can push here too, if you accept the ideas in it.

vitaly-krugl:
@everilae, after PR #602 got merged, there are now some merge conflicts in your PR that need to be resolved. Would you mind resolving those merge conflicts, rebasing against upstream/master and squashing your changes into a single commit? Also, the merged PR #602 includes a number of new blocking adapter acceptance tests that had to use the `as_bytes` and `_p3_as_bytes` that need to be removed when merging with your changes. Thanks!

vitaly-krugl:
Gavin, this PR gets pika back to being able to deal with exchange/queue/routing_key names intuitively as strings/unicode in both Python 2.x and 3.x. Please consider merging **after** the above-described merge conflict and test changes are resolved. Thx.

everilae:
History rewritten, lots of painful mistakes made. Apparently should not have merged before rebase etc...

vitaly-krugl:
Thanks, @everilae! What would be the correct workflow to handle the merge/rebase in this case? I don't have a good handle on this workflow, yet.

vitaly-krugl:
+1

everilae:
Followed basic squashing/rebasing instructions, except I had merged latest changes from master before that. Turned out to be a disaster, so had to undo that merge. From there on it turned out ok. I guess it's ok to do this kind of history rewriting since nobody else has based any work on top of this.

Had to rebase own work though...

vitaly-krugl:
thx for the merge/rebase info, @everilae

PR #604, opened 2015-06-16
Added EINTR loops in select_connection pollers and around socket.send/recv calls.

### LINKED ISSUE #596: UnboundLocalError: local variable 'read' referenced before assignment (Python 3) in SelectPoller.poll
There is a critical flaw in the [`poll`](https://github.com/pika/pika/blob/master/pika/adapters/select_connection.py#L320) function of `SelectPoller`.

If an `EINTR` exit occurs (and it **will** occur sooner or later), the `read, write, error` tuple is not assigned at all; neither would there be any exception being thrown, and the code crashes below with `UnboundLocalError: local variable 'read' referenced before assignment`.

Possibly at least each of these names should be assigned to an empty iterable by default, thus

    read = write = error = ()
    try:
         ....

It is to be noted that all Linux / Glibc-using Python versions prior to Python 3.5 are vulnerable to EINTR; in Python 3.5> Python ought to handle the EINTR instead.

vitaly-krugl:
Fix #596 UnboundLocalError: local variable 'read' referenced before assignment (Python 3) in SelectPoller.poll:

Added EINTR loops in select_connection pollers and around socket.send/recv calls.
Addressed some pylint findings, including an error or two.
Wrap socket.send and socket.recv calls in EINTR loops.

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -8,6 +8,7 @@ import logging
 import socket
 import ssl
 
+import pika.compat
 from pika import connection
 from pika import exceptions
 
@@ -15,6 +16,14 @@ try:
     SOL_TCP = socket.SOL_TCP
 except AttributeError:
     SOL_TCP = 6
+
+
+if pika.compat.PY2:
+    _SOCKET_ERROR = socket.error
+else:
+    # socket.error was deprecated and replaced by OSError in python 3.3
+    _SOCKET_ERROR = OSError
+
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


vitaly-krugl:
@ztane, this should address #596. Please review.

vitaly-krugl:
CC @wjps

ztane:
I guess one should be really using `EnvironmentError` which is the superclass of all `errno` errors. **Except** of course in Python 2.6, 2.7 the `select.error` is not a subclass of `EnvironmentError` nor `IOError`, and might not even have `.errno`?!

ztane:
Welcome to hell. Python 2:

```
>>> import select
>>> try:
...     select.select([666], [], [], 1)
... except select.error as e:
...     pass
... 
([666], [], [])
>>> 
>>> try:
...     select.select([42], [], [], 1)
... except select.error as e:
...     pass
... 
>>> e
error(9, 'Bad file descriptor')
>>> e.errno
AttributeError: 'error' object has no attribute 'errno'
```

The `select.error` will be created in Python 2.6, 2.7 with arguments `errno, message`, but there is no `errno` property; instead the `errno` value will be in `args[0]`, and this should be compared against the `EINTR`.

vitaly-krugl:
Thank you for the feedback, @ztane. I fixed things up according to python documentation, whereby the socket and select base exception classes socket.error and select.error were replaced by OSError in python 3.3.

vitaly-krugl:
@gmr, I think we're good to go with this PR. Thx.

Issue #609, opened 2015-06-25
Race condition in blocking_adapter_test.py:TestBasicCancelWithNonAckableConsumer


everilae:
While running tests had this come up ~~once, but could not reproduce~~. Perhaps a very elusive beast.

```
======================================================================
FAIL: BlockingChannel user cancels non-ackable consumer via basic_cancel
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/saarni/Work/liilak/pika/tests/acceptance/blocking_adapter_test.py", line 1489, in test
    self.assertEqual(len(messages), 2)
AssertionError: 1 != 2
-------------------- >> begin captured logging << --------------------
pika.adapters.select_connection: DEBUG: Using SelectPoller
pika.callback: DEBUG: Added: {'callback': <bound method _CallbackResult.set_value_once of <pika.adapters.blocking_connection._CallbackResult object at 0x7fb3723359e0>>, 'only': None, 'one_shot': False, 'arguments': None}
pika.callback: DEBUG: Added: {'callback': <bound method _CallbackResult.set_value_once of <pika.adapters.blocking_connection._CallbackResult object at 0x7fb372329fc8>>, 'only': None, 'one_shot': False, 'arguments': None}
pika.callback: DEBUG: Added: {'callback': <bound method _CallbackResult.set_value_once of <pika.adapters.blocking_connection._CallbackResult object at 0x7fb372335878>>, 'only': None, 'one_shot': False, 'arguments': None}
pika.callback: DEBUG: Added: {'callback': <bound method SelectConnection._on_connection_start of <pika.adapters.select_connection.SelectConnection object at 0x7fb372f86e10>>, 'only': None, 'one_shot': True, 'arguments': None, 'calls': 1}
pika.adapters.base_connection: INFO: Connecting to 127.0.0.1:5672
pika.callback: DEBUG: Processing 0:Connection.Start
pika.callback: DEBUG: Processing use of oneshot callback
pika.callback: DEBUG: 0 registered uses left
pika.callback: DEBUG: Removing callback #0: {'callback': <bound method SelectConnection._on_connection_start of <pika.adapters.select_connection.SelectConnection object at 0x7fb372f86e10>>, 'only': None, 'one_shot': True, 'arguments': None, 'calls': 0}
pika.callback: DEBUG: Calling <bound method SelectConnection._on_connection_start of <pika.adapters.select_connection.SelectConnection object at 0x7fb372f86e10>> for "0:Connection.Start"
[...]

PR #610, opened 2015-06-26
PIKA-609 fix race condition in blocking adapter tests

vitaly-krugl:
PIKA-609 Wait for broker to dispatch all messages to client before canceling consumer in acceptance tests TestBasicCancelWithNonAckableConsumer and TestBasicCancelWithAckableConsumer


vitaly-krugl:
@everilae, this should address the race condition that you discovered in blocking adapter acceptance tests per #609.  I was not able to reproduce the error on my MacBookPro, but was able to confirm presence of race condition upon examination of the test's code.

PR #614, opened 2015-06-27
TwistedProtocolConnection fix example for @gaochunzy's PR 613

vitaly-krugl:
@gaochunzy, I reviewed your PR #613. It still had one conflicting method `_send_frame` that needed to be removed and it turns out that `_flush_outbound` can even be simpler than I thought initially. I submitted this PR as an example of what I had in mind. Please start with this and also add a basic acceptance test in `tests/acceptance/twisted_protocol_connection_test.py` (new test script) for TwistedProtocolConnection that at least connects, creates a channel, creates a queue, publishes a message to that queue, consumes the message from the queue and validates its contents, deletes the queue.

### DIFF
--- a/examples/twisted_service.py
+++ b/examples/twisted_service.py
@@ -1,3 +1,4 @@
+"""
 # -*- coding:utf-8 -*-
 # based on:
 #  - txamqp-helpers by Dan Siemon <dan@coverfire.com> (March 2010)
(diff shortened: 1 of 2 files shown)


vitaly-krugl:
@gmr: This example PR is the correct fix #614. #614 is incomplete.

Issue #617, opened 2015-06-27
Fix errors found by pylint in twisted_service.py example


vitaly-krugl:
Running `pylint examples/twisted_service.py` reveals a number of errors (in addition to a bunch of warnings, etc., that would also be good to fix). Here is a sampling of errors from running pylint on https://github.com/pika/pika/blob/0d2287de27bed2dde02958a32a481b07b3d47e34/examples/twisted_service.py:

```
$ pylint twisted_service.py -E
No config file found, using default configuration
************* Module twisted_service
E: 41,16: Undefined variable 'ssl' (undefined-variable)
E: 42,19: Module 'twisted.application.internet' has no 'SSLClient' member (no-member)
E: 44,19: Module 'twisted.application.internet' has no 'TCPClient' member (no-member)
E: 59, 4: An attribute affected in twisted_service line 62 hide this method (method-hidden)
E: 59, 4: method already defined line 55 (function-redefined)
E:177,20: Too many positional arguments for function call (too-many-function-args)
```

Issue #616, opened 2015-06-27
Implement acceptance tests for TwistedConnection adapter


vitaly-krugl:
There are presently no acceptance tests for TwistedConnection adapter.

Here is an example of a basic test script from [python unittest framework](https://docs.python.org/2/library/unittest.html); the test method names need to begin with "test_" to be recognized by the framework; you can have helper/private methods with names that don't begin with "test_"; it's desirable to use test method names that are descriptive of what's being tested (long names are okay in these tests).

```
import unittest

class TestStringMethods(unittest.TestCase):

  def test_upper(self):
      self.assertEqual('foo'.upper(), 'FOO')

  def test_isupper(self):
      self.assertTrue('FOO'.isupper())
      self.assertFalse('Foo'.isupper())

  def test_split(self):
      s = 'hello world'
      self.assertEqual(s.split(), ['hello', 'world'])
      # check that s.split fails when the separator is not a string
      with self.assertRaises(TypeError):
          s.split(2)

if __name__ == '__main__':
    unittest.main()
```

Definitely use unittest's specialized `self.assert*` methods rather than python's builtin `assert`. Unittest's specialized assertion methods output useful info when assertions fail.

The framework instantiates a fresh instance of the test class for each test method. Don't define a constructor in your test class; if you need to initialize something common for each test method, then override the [def setUp(self)](https://docs.python.org/2/library/unittest.html#unittest.TestCase.setUp) framework "fixture" method and do it there.

There is a lot of flexibility for structuring these tests: you can have a dedicated test class for each test method (as in https://github.com/pika/pika/blob/master/tests/acceptance/blocking_adapter_test.py) or multiple test methods per test class (as in the above example). It's up to you.

The Twisted adapter to be tested is in https://github.com/pika/pika/blob/master/pika/adapters/twisted_connection.py. There are actually two adapters there: `TwistedConnection` and `TwistedProtocolConnection`. Ideally, both need to be tested, but practically just start with the one that you're most comfortable with. You would create tests for `TwistedConnection` in `pika/tests/acceptance/twisted_connection_test.py` and similarly for `TwistedProtocolConnection` in  `pika/tests/acceptance/twisted_protocol_connection_test.py`.

[...]

PR #621, opened 2015-07-09
Fixed SSL import and removed no_ack=True in favor of explicit AMQP me…

skftn:
Addresses [this issue](https://github.com/pika/pika/issues/617) and removes `no_ack=True` to provide an example that is more Twisted-like.

It provides a better base for creating an asynchronous task based consumer with a pre-defined queue limit.

### DIFF
--- a/examples/twisted_service.py
+++ b/examples/twisted_service.py
@@ -14,6 +14,10 @@ Fire up this test application via `twistd -ny twisted_service.py`
 The application will answer to requests to exchange "foobar" and any of the
 routing_key values: "request1", "request2", or "request3"
 with messages to the same exchange, but with routing_key "response"
+
+When a routing_key of "task" is used on the exchange "foobar",
+the application can asynchronously run a maximum of 2 tasks at once
+as defined by PREFETCH_COUNT
 """
 
 import pika
@@ -25,7 +29,9 @@ from twisted.internet import protocol
 from twisted.application import internet
 from twisted.application import service
 from twisted.internet.defer import inlineCallbacks
+from twisted.internet import ssl, defer, task
 from twisted.python import log
+from twisted.internet import reactor
 
 PREFETCH_COUNT = 2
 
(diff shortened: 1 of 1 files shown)


vitaly-krugl:
@skftn: thanks for your contribution. Would you mind contributing some basic acceptance tests for Twisted connection? See https://github.com/pika/pika/blob/master/tests/acceptance/blocking_adapter_test.py as an example. Even just a handful of tests (e.g., connect, create channel, publish, get, consume) would be great. There aren't any tests at all for Twisted right now. Many thanks in advance!

skftn:
@vitaly-krugl: Which files would need to be tested? I don't have any experience in writing tests for Twisted (or unittests in general). I can give it a try however.

vitaly-krugl:
@skftn: I was thinking about acceptance tests for twisted adapters, rather than unit tests, just to get started. The project uses the python [unittest](https://docs.python.org/2/library/unittest.html) framework for all its tests - both unit and acceptance. Each test is a targeted basic usage workflow: e.g.,

1. method `test_twisted_connect_and_close`: initiate the connection sequence

Here is an example of a basic test script from [python unittest framework](https://docs.python.org/2/library/unittest.html); the test method names need to begin with "test_" to be recognized by the framework; you can have helper/private methods with names that don't begin with "test_"; it's desirable to use test method names that are descriptive of what's being tested (long names are okay in these tests).
```
import unittest

class TestStringMethods(unittest.TestCase):

  def test_upper(self):
      self.assertEqual('foo'.upper(), 'FOO')

  def test_isupper(self):
      self.assertTrue('FOO'.isupper())
      self.assertFalse('Foo'.isupper())

  def test_split(self):
      s = 'hello world'
      self.assertEqual(s.split(), ['hello', 'world'])
      # check that s.split fails when the separator is not a string
      with self.assertRaises(TypeError):
          s.split(2)

if __name__ == '__main__':
    unittest.main()
```

Definitely use unittest's specialized `self.assert*` methods rather than python's builtin `assert`. Unittest's specialized assertion methods output a bunch of useful info when assertions fail.

There is a lot of flexibility for structuring these tests: you can have a dedicated test class for each test method (as in https://github.com/pika/pika/blob/master/tests/acceptance/blocking_adapter_test.py) or multiple test methods per test class (as in the above example). It's up to you.

https://github.com/pika/pika/blob/master/tests/acceptance/blocking_adapter_test.py is an example of an acceptance test suite for the pika blocking adapter. Note the use of `self.addCleanup` immediately after creation of resources to schedule their destruction and the use of uuid to avoid collisions, especially when tests are executed in parallel.

vitaly-krugl:
@skftn: issue #616 is tracking acceptance tests for the pika Twisted adapter

PR #627, opened 2015-07-20
PEP-343 for blocking connection

reddec:
Adds support for 'with' context for BlockingConnection
[PEP-343](https://www.python.org/dev/peps/pep-0343/)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,10 @@
-0.10.0dev0 2015-07-10
----------------------
+0.10.0b2 2015-07-15
+-------------------
+
+ - f72b58f - Fixed failure to purge _ConsumerCancellationEvt from BlockingChannel._pending_events during basic_cancel. (Vitaly Kruglikov)
+
+0.10.0b1 2015-07-10
+-------------------
 
 High-level summary of notable changes:
 
--- a/docs/version_history.rst
+++ b/docs/version_history.rst
@@ -1,7 +1,12 @@
 Version History
 ===============
 
-0.10.0dev0 2015-07-10
+0.10.0b2 2015-07-15
+-------------------
+
+ - f72b58f - Fixed failure to purge _ConsumerCancellationEvt from BlockingChannel._pending_events during basic_cancel. (Vitaly Kruglikov)
+
+0.10.0b1 2015-07-10
 ---------------------
 
 High-level summary of notable changes:
--- a/pika/__init__.py
+++ b/pika/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '0.10.0b1'
+__version__ = '0.10.0b2'
 
 import logging
 try:
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ long_description = ('Pika is a pure-Python implementation of the AMQP 0-9-1 '
                     'with other AMQP 0-9-1 brokers.')
 
 setup(name='pika',
-      version='0.10.0b1',
+      version='0.10.0b2',
       description='Pika Python AMQP Client Library',
       long_description=open('README.rst').read(),
       maintainer='Gavin M. Roy',


pika/adapters/blocking_connection.py
```diff
@@ -700,6 +700,15 @@ def channel(self, channel_number=None):
 
         return channel
 
+    def __enter__(self):
+        # Prepare `with` context
+        return self
+
+    def __exit__(self, *args):
+        # Close connection after `with` context
+        self.close()
```
vitaly-krugl: If exiting due to SystemExit, KeyboardInterrupt, or other non-Exception-derived exceptions, we most likely don't want to perform anything graceful that slows down the exit of the app. Anyone want to chime in?

pika/adapters/blocking_connection.py
```diff
@@ -700,6 +700,15 @@ def channel(self, channel_number=None):
...
+        return self
```
vitaly-krugl: BUG: This is the wrong return value for `__exit__`. Returning anything that evaluates to True (like `self` in this case) will suppress the exception. We don't want to suppress the exception. To fix, remove the `return` statement from `__exit__`. See http://effbot.org/zone/python-with-statement.htm

vitaly-krugl:
@reddec, please **fix the issues** noted in other comments and **add tests** for your change, before this PR may be merged. We need tests along with PR's to maintain quality of the adapter. The tests need to confirm that:
# the connection was indeed closed for non-system exception and that the exception is not suppressed
# the connection is not closed for system exception (one that is not derived from `Exception`; e.g., SystemExit) and  that the exception is not suppressed

reddec:
@vitaly-krugl thanks for review. I will fix it. Are you sure, that connection have not to be closed for system exception? At least like this:
```python
self._cancel_all_consumers()
self._cleanup()
```
without any notification message (that may be slow)

pika/adapters/blocking_connection.py
```diff
@@ -700,6 +700,18 @@ def channel(self, channel_number=None):
 
         return channel
 
+    def __enter__(self):
+        # Prepare `with` context
+        return self
+
+    def __exit__(self, tp, value, traceback):
+        # Close connection after `with` context
+        if isinstance(value, (KeyboardInterrupt, SystemExit, GeneratorExit)):
```
vitaly-krugl: @reddec: on second thought, let's keep them all the same; let's just call `self.cose()` in all cases. Sorry for the initial feedback - I was thinking out loud.

vitaly-krugl:
Also, please add a test for this in `tests/acceptance/blocking_adapter_test.py` near the other connection tests at the top. This could be something as simple as:

```
class TestConnectionContextManagerClosesConnection(BlockingTestCaseBase):
    def test(self):
        """BlockingConnection: connection context manager closes connection"""
        with self._connect() as connection:
            self.assertIsInstance(connection, pika.BlockingConnection)
            self.assertTrue(connection.is_open)

        self.assertTrue(connection.is_closed)

class TestConnectionContextManagerClosesConnectionAndPassesOriginalException(BlockingTestCaseBase):
    def test(self):
        """BlockingConnection: connection context manager closes connection and passes original exception"""
        class MyException(Exception):
            pass

        with self.assertRaises(MyException):
            with self._connect() as connection:
                self.assertTrue(connection.is_open)

                raise MyException()

        self.assertTrue(connection.is_closed)
```

reddec:
Thank you for help!
Does it look good now? (tests

pika/adapters/blocking_connection.py
```diff
@@ -700,6 +700,15 @@ def channel(self, channel_number=None):
...
         return channel
 
+    def __enter__(self):
+        # Prepare `with` context
+        return self
+
+    def __exit__(self, tp, value, traceback):
+        # Close connection after `with` context
+        self.close()
+        return isinstance(value, (KeyboardInterrupt, SystemExit, GeneratorExit))
```
vitaly-krugl: @reddec: we don't want the `return isinstance(...` statement here at all. This is because we don't want to suppress any of these exceptions in pika. This would prevent `CTL-C` from stopping the program. If user's code want's to suppress them, then user's code is free to do that on its own.

Let's have just this in `__exit__`:

```
def __exit__(self, *_args):
    self.close()
```
and add the test:
```
class TestConnectionContextManagerClosesConnectionAndPassesSystemException(BlockingTestCaseBase):
    def test(self):
        """BlockingConnection: connection context manager closes connection and passes system exception"""
        with self.assertRaises(SystemExit):
            with self._connect() as connection:
                self.assertTrue(connection.is_open)
                raise SystemExit()

        self.assertTrue(connection.is_closed)

```

reddec:
Done

vitaly-krugl:
Thanks @reddec, the code looks good. Please rebase against master and squash into a single commit to prepare for merging.

reddec:
[Done ](https://github.com/pika/pika/pull/629)

PR #571, opened 2015-05-14
[ACC-22] add tunable client props to pika

mffrench:
Hello,

I added a little functionality which help you to define the application client properties. eg : 

```
import pika

client_properties = {
    'product': 'my product',
    'information': 'my information',
    'my_other_props': 'my other property',
}

credentials = pika.PlainCredentials('guest', 'guest')
parameters = pika.ConnectionParameters("localhost", 5672, '/',
                                       credentials=credentials, client_props=client_properties)
connection = pika.BlockingConnection(parameters)
```

You'll be then able to differentiate python connections from each other more easily...

Cheers,

Mathilde
screen shot 2015-05-14 at 12 26 26

PR #628, opened 2015-07-20
Add the hostname of the server to client properties

markunsworth:
Passes the hostname of the client to the server when connecting. Useful for identifying connections when  multiple clients share the same username


mffrench:
Hello @markunsworth,

interesting solution but what if you have several connections from same client (which could happen as pika connections are not really thread safe) ?

Currently still waiting this pull request or some other solutions : 
https://github.com/pika/pika/pull/571

Cheers,

Mathilde

markunsworth:
We just wanted to know which connections were coming from which servers as the ip addresses shown in the `name` don't seem to match.

vitaly-krugl:
@markunsworth: what if the user needs to pass fqdn?

vitaly-krugl:
@markunsworth: please add tests to your PR.

markunsworth:
@vitaly-krugl yes, it's limited for now, maybe pull request #571 by @mffrench is a better place to do this, we just need to agree on an approach

gmr:
Closing for now due to lack of tests or agreed upon method of doing this. Thanks for the PR, please feel free to re-open if you add test coverage.

PR #630, opened 2015-07-23
Check the existence of a timeout-handler key before removing as the h…

shinji-s:
…andler may have deleted the key. (E.g. heartbeat checker)

Running the program below against pika@91a5f45bbff666754dd18f11d8a48ffa296e1824 to connect to a rabbitmq server running under VirtualBox and then suspending the VM will generate the following stack trace.

import pika
print pika.__file__

connection = pika.BlockingConnection(pika.ConnectionParameters(
    heartbeat_interval = 10,
    host='suzuki-VirtualBox'))

channel.queue_declare(queue='hello')

print ' [*] Waiting for messages. To exit press CTRL+C'

def callback(ch, method, properties, body):
    print " [x] Received %r" % (body,)

channel.basic_consume(callback,
                      queue='hello',
                      no_ack=True)

channel.start_consuming()

  File "test.py", line 26, in <module>
    channel.start_consuming()
  File "../pika/adapters/blocking_connection.py", line 1681, in start_consumin\
g
  File "../pika/adapters/blocking_connection.py", line 647, in process_data_ev\
ents
    self._flush_output(common_terminator)
  File "../pika/adapters/blocking_connection.py", line 411, in _flush_output
    self._impl.ioloop.process_timeouts()
  File "../pika/adapters/select_connection.py", line 279, in process_timeouts
    del self._timeouts[hash(frozenset(t.items()))]
KeyError: 5302635927768386976

This is because the timer is deleted at:

  File "test.py", line 26, in <module>
    channel.start_consuming()
  File "../pika/adapters/blocking_connection.py", line 1681, in start_consuming
  File "../pika/adapters/blocking_connection.py", line 647, in process_data_events
    self._flush_output(common_terminator)
  File "../pika/adapters/blocking_connection.py", line 411, in _flush_output
    self._impl.ioloop.process_timeouts()
  File "../pika/adapters/select_connection.py", line 278, in process_timeouts
    t['callback']()
  File "../pika/heartbeat.py", line 88, in send_and_check
    return self._close_connection()
  File "../pika/heartbeat.py", line 120, in _close_connection
    self._connection._adapter_disconnect()
  File "../pika/adapters/select_connection.py", line 96, in _adapter_disconnect
    super(SelectConnection, self)._adapter_disconnect()
  File "../pika/adapters/base_connection.py", line 152, in _adapter_disconnect
    self._remove_heartbeat()
  File "../pika/connection.py", line 1011, in _remove_heartbeat
    self.heartbeat.stop()
  File "../pika/heartbeat.py", line 110, in stop
    raise Exception('self._connection.remove_timeout(self._timer)')
Exception: self._connection.remove_timeout(self._timer)

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -28,22 +28,6 @@ WRITE = 0x0004
 ERROR = 0x0008
 
 
-if pika.compat.PY2:
-    _SELECT_ERROR = select.error
-else:
-    # select.error was deprecated and replaced by OSError in python 3.3
-    _SELECT_ERROR = OSError
-
-
-def _get_select_errno(error):
-    if pika.compat.PY2:
-        assert isinstance(error, select.error), repr(error)
-        return error.args[0]
-    else:
-        assert isinstance(error, OSError), repr(error)
-        return error.errno
-
-
 class SelectConnection(BaseConnection):
     """An asynchronous connection adapter that attempts to use the fastest
     event loop adapter for the given platform.
@@ -392,8 +376,8 @@ class SelectPoller(object):
                                                    self._fd_events[ERROR],
                                                    self.get_next_deadline())
                 break
-            except _SELECT_ERROR as error:
-                if _get_select_errno(error) == errno.EINTR:
+            except pika.compat._SELECT_ERROR as error:
+                if pika.compat._is_resumable_error(error):
                     continue
                 else:
                     raise
@@ -507,8 +491,8 @@ class KQueuePoller(SelectPoller):
                 kevents = self._kqueue.control(None, 1000,
                                                self.get_next_deadline())
                 break
-            except _SELECT_ERROR as error:
-                if _get_select_errno(error) == errno.EINTR:
+            except pika.compat._SELECT_ERROR as error:
+                if pika.compat._is_resumable_error(error):
                     continue
                 else:
                     raise
@@ -582,8 +566,8 @@ class PollPoller(SelectPoller):
             try:
                 events = self._poll.poll(self.get_next_deadline())
                 break
-            except _SELECT_ERROR as error:
-                if _get_select_errno(error) == errno.EINTR:
+            except pika.compat._SELECT_ERROR as error:
+                if pika.compat._is_resumable_error(error):
                     continue
                 else:
                     raise
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -276,7 +276,9 @@ class SelectPoller(object):
         # were only run periodically.
         for t in sorted(to_run, key=itemgetter('deadline')):
             t['callback']()
-            del self._timeouts[hash(frozenset(t.items()))]
+            k = hash(frozenset(t.items()))
+            if k in self._timeouts:
+                    del self._timeouts[k]
             self._next_timeout = None
 
     def add_handler(self, fileno, handler, events):
--- a/pika/compat.py
+++ b/pika/compat.py
@@ -1,5 +1,5 @@
 import sys as _sys
-
+import errno
 
 PY2 = _sys.version_info < (3,)
 PY3 = not PY2
@@ -73,6 +73,9 @@ if not PY2:
 
         return str(value)
 
+    _SELECT_ERROR = OSError
+    def _is_resumable_error(x): return x.errno==errno.EINTR
+
 else:
     from urllib import unquote as url_unquote, urlencode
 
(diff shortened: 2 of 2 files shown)


vitaly-krugl:
@shinji-s - thanks for the info about the timeout error.

However, the IOError commit caf0e4e0148a029fb212264f11f718ecc9f206d6 is incorrect. Let's deal with select errors in a separate PR, where you will need to add tests for each of SelectConnection's poller types (select, kqueue, poll, epoll) to validate that your change does the correct thing. We can't go on belief alone. One reason why caf0e4e0148a029fb212264f11f718ecc9f206d6 is incorrect is that on Python 2.7, select.select raises select.error, not IOError:

```
Python 2.7.6 (default, Sep  9 2014, 15:04:36) 
Type "copyright", "credits" or "license" for more information.

IPython 1.1.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import select

In [2]: try:         
    select.select([99], [], [])
except Exception as e:
    raise
   ...: 
---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
<ipython-input-2-16b13ecd4094> in <module>()
      1 try:
----> 2     select.select([99], [], [])
      3 except Exception as e:
      4     raise
      5 

error: (9, 'Bad file descriptor')

In [3]: repr(e)
Out[3]: "error(9, 'Bad file descriptor')"

In [4]: type(e)
Out[4]: select.error

In [5]: e.__class__.__bases__
Out[5]: (Exception,)
```

As you can see, it's not IOError. However, things may be different with select.kqueue, select.poll, and select.epoll (maybe?). That's why you need to do some more work on it and submit it in a separate PR.

pika/adapters/select_connection.py
```diff
@@ -276,7 +260,9 @@ def process_timeouts(self):
         # were only run periodically.
         for t in sorted(to_run, key=itemgetter('deadline')):
             t['callback']()
-            del self._timeouts[hash(frozenset(t.items()))]
+            k = hash(frozenset(t.items()))
+            if k in self._timeouts:
```
vitaly-krugl: @shinji-s: also consider the possibility that one timer callback can unregister additional timers. If timer A callback unregisters timers B and C that were also ready, then B and C callbacks should **not** be called. So you need something like this to really fix the bug:

```
for t in sorted(to_run, key=itemgetter('deadline')):
    k = hash(frozenset(t.items()))
    if k in self._timeouts:
        t['callback']()
        self._timeouts.pop(k, None)
```

vitaly-krugl: Another surprising aspect of the legacy implementation is the unnecessary re-computation of the key. The hash key is available in self._timeouts, so the `to_run = ...` statement could also save the key along with `timer` as a tuple in order to avoid the re-computation here.

vitaly-krugl:
@shinji-s: to summarize, please:
1. Remove the select/IOError commit from this PR, since it's incorrect (as described in my earlier comment), and also so we can focus on fixing a single problem.
2. Address feedback https://github.com/pika/pika/pull/630#discussion_r35355729 regarding checking `if k in self._timeouts:` before calling `t['callback']()`.
3. Add tests that verify your timer fix.

shinji-s:
@vitaly-krugl 
Thank you for your valuable feedback. I thought I've sent a PR for the timer-fix only but it seems the EINTR patch was piggy backed. Sorry for the extra review work that has caused. Following your advice I'll first work on renewing PR for the timer issue.

To me it seems forbidding timer removal within a timeout handler altogether  is a cleaner solution, which may be too big a change to bring in now and/or too difficult/costly to implement.  With that said, I would take your patch as is for the imminent fix.

As for EINTR issue, raised exception class varies depending on the polling mechanism being used. Looking at selectmodule.c in python 2.7.3, the mapping is as follows;
select => SelectError
poll => SelectError
epoll => IOError
kevent => OSError
kqueue => IOError
It seems the exception handler need to be prepared for more than one kind of exception class even though we hypothetically target 2.7.3 only. I'll work on it after preparing a renewed PR for the timer.

vitaly-krugl:
Thank you @shinji-s!

shinji-s:
@vitaly-krugl 
Will you please have a look on the new patch below before my sending in another PR and tell me if it's good enough for submission? (Key recomplutation optimization seemed less complex than I had thought. I bit the bullet. Please advise if you think that is too much to carry in the patch.)

https://github.com/shinji-s/pika/commit/e1d287bfc05da79c44aadfeb04843e03dd9fbaee

As I have not dealt with unittest beffore and  I don't understand how the test cases are organized in pika, I might be doing something terribly wrong. Please holler if I did.

PR #633, opened 2015-07-29
Fix KeyError exception caused by timeout handler(s) that remove thems…

shinji-s:
…elves during execution. (Notably the heartbeat checker.)

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -264,20 +264,37 @@ class SelectPoller(object):
         timeout = min((timeout, SelectPoller.POLL_TIMEOUT))
         return timeout * SelectPoller.POLL_TIMEOUT_MULT
 
+    @staticmethod
+    def is_ready_to_fire(to_run):
+        """For patching by unittest"""
+        return True
+
     def process_timeouts(self):
         """Process the self._timeouts event stack"""
 
         now = time.time()
-        to_run = [timer for timer in self._timeouts.values()
-                  if timer['deadline'] <= now]
-
         # Run the timeouts in order of deadlines. Although this shouldn't
         # be strictly necessary it preserves old behaviour when timeouts
         # were only run periodically.
-        for t in sorted(to_run, key=itemgetter('deadline')):
-            t['callback']()
-            del self._timeouts[hash(frozenset(t.items()))]
+        to_run = sorted([(k,timer) for (k,timer) in self._timeouts.items()
+                         if timer['deadline'] <= now],
+                        key = lambda (k,timer) : timer['deadline'])
+
+        if self.is_ready_to_fire(to_run):
+            # Force recalculation as the oldest timer is going to be removed.
             self._next_timeout = None
+
+            for k, timer in to_run:
+                if k not in self._timeouts:
+                    # Previous invocation(s) should have deleted the timer.
+                    continue
+                try :
+                    timer['callback']()
+                finally:
+                    # Don't do 'del self._timeout[k]' as the key might
+                    # have been deleted just now.
+                    self._timeouts.pop(k, None)
+
 
     def add_handler(self, fileno, handler, events):
         """Add a new fileno to the set to be monitored
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -8,7 +8,6 @@ import socket
 import select
 import errno
 import time
-from operator import itemgetter
 from collections import defaultdict
 import threading
 
@@ -272,15 +271,15 @@ class SelectPoller(object):
         # Run the timeouts in order of deadlines. Although this shouldn't
         # be strictly necessary it preserves old behaviour when timeouts
         # were only run periodically.
-        to_run = sorted([(k,timer) for (k,timer) in self._timeouts.items()
+        to_run = sorted([(k, timer) for (k, timer) in self._timeouts.items()
                          if timer['deadline'] <= now],
-                        key = lambda item : item[1]['deadline'])
+                        key=lambda item: item[1]['deadline'])
 
         for k, timer in to_run:
             if k not in self._timeouts:
                 # Previous invocation(s) should have deleted the timer.
                 continue
-            try :
+            try:
                 timer['callback']()
             finally:
                 # Don't do 'del self._timeout[k]' as the key might
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -264,10 +264,6 @@ class SelectPoller(object):
         timeout = min((timeout, SelectPoller.POLL_TIMEOUT))
         return timeout * SelectPoller.POLL_TIMEOUT_MULT
 
-    @staticmethod
-    def is_ready_to_fire(to_run):
-        """For patching by unittest"""
-        return True
 
     def process_timeouts(self):
         """Process the self._timeouts event stack"""
@@ -280,20 +276,17 @@ class SelectPoller(object):
                          if timer['deadline'] <= now],
                         key = lambda item : item[1]['deadline'])
 
-        if self.is_ready_to_fire(to_run):
-            # Force recalculation as the oldest timer is going to be removed.
-            self._next_timeout = None
-
-            for k, timer in to_run:
-                if k not in self._timeouts:
-                    # Previous invocation(s) should have deleted the timer.
-                    continue
-                try :
-                    timer['callback']()
-                finally:
-                    # Don't do 'del self._timeout[k]' as the key might
-                    # have been deleted just now.
-                    self._timeouts.pop(k, None)
+        for k, timer in to_run:
+            if k not in self._timeouts:
+                # Previous invocation(s) should have deleted the timer.
+                continue
+            try :
+                timer['callback']()
+            finally:
+                # Don't do 'del self._timeout[k]' as the key might
+                # have been deleted just now.
+                if self._timeouts.pop(k, None) is not None:
+                    self._next_timeout = None
 
 
     def add_handler(self, fileno, handler, events):
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -278,7 +278,7 @@ class SelectPoller(object):
         # were only run periodically.
         to_run = sorted([(k,timer) for (k,timer) in self._timeouts.items()
                          if timer['deadline'] <= now],
-                        key = lambda (k,timer) : timer['deadline'])
+                        key = lambda item : item[1]['deadline'])
 
         if self.is_ready_to_fire(to_run):
             # Force recalculation as the oldest timer is going to be removed.
--- a/tests/unit/select_connection_ioloop_tests.py
+++ b/tests/unit/select_connection_ioloop_tests.py
@@ -106,12 +106,12 @@ class IOLoopTimerTestSelect(IOLoopBaseTest):
         handle_holder = []
         self.timer_got_fired = False
         self.handle = self.ioloop.add_timeout(
-            0.1, partial(self._on_timer_deletes_itself, handle_holder))
+            0.1, partial(self._on_timer_delete_itself, handle_holder))
         handle_holder.append(self.handle)
         self.start()
         self.assertTrue(self.timer_got_called)
 
-    def _on_timer_deletes_itself(self, handle_holder):
+    def _on_timer_delete_itself(self, handle_holder):
         """ A timeout hanlder that tries to remove itself. """
         self.assertEqual(self.handle, handle_holder.pop())
         # This removal here should not raise exception by itself nor
--- a/tests/unit/select_connection_ioloop_tests.py
+++ b/tests/unit/select_connection_ioloop_tests.py
@@ -5,12 +5,6 @@ Tests for SelectConnection IOLoops
 """
 import logging
 try:
-    from unittest import mock
-    patch = mock.patch
-except ImportError:
-    import mock
-    from mock import patch
-try:
     import unittest2 as unittest
 except ImportError:
     import unittest
@@ -39,10 +33,8 @@ class IOLoopBaseTest(unittest.TestCase):
         self.ioloop.remove_timeout(self.fail_timer)
         self.ioloop = None
 
-    def start(self, callback = None):
-        if callback is None:
-            callback = self.on_timeout
-        self.fail_timer = self.ioloop.add_timeout(self.TIMEOUT, callback)
+    def start(self):
+        self.fail_timer = self.ioloop.add_timeout(self.TIMEOUT, self.on_timeout)
         self.ioloop.start()
 
     def on_timeout(self):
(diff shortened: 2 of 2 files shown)


gmr:
At the very minimum, your PR needs to pass the unit tests for all supported versions of Python. At the moment it appears to fail for Python 3. You can check out the details in the link from Travis-CI above.

shinji-s:
Thank you for the poiner to Travis-CI. The issue with Python3 should have been resolved.
For unit testing I need to have two timeouts fired in the same invocation of process_timeouts(). 
What do you have to say about wrapping the expression below in a method and patching it when testing?
      [(k,timer) for (k,timer) in self._timeouts.items() if timer['deadline'] <= now]

pika/adapters/select_connection.py
```diff
@@ -264,21 +264,38 @@ def get_next_deadline(self):
         timeout = min((timeout, SelectPoller.POLL_TIMEOUT))
         return timeout * SelectPoller.POLL_TIMEOUT_MULT
 
+    @staticmethod
+    def is_ready_to_fire(to_run):
```
vitaly-krugl: We shouldn't need to inject test code into production code.

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -81,7 +89,54 @@ def start_test(self):
...
+            self.ioloop.stop()
+
+    def test_normal(self):
+        self.start_test()
+
+    def test_timer_for_suicide(self):
+        self.timer_stack = list()
+        handle_holder = []
+        self.handle = self.ioloop.add_timeout(
+            0.1, partial(self.on_timer_suicide,handle_holder))
```
vitaly-krugl: Please check your changes with pylint.

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -81,7 +89,54 @@ def start_test(self):
...
     def on_timer(self, val):
-        self.assertEquals(val, self.timer_stack.pop())
+        self.assertEqual(val, self.timer_stack.pop())
+        if not self.timer_stack:
+            self.ioloop.stop()
+
+    def test_normal(self):
+        self.start_test()
+
+    def test_timer_for_suicide(self):
```
vitaly-krugl: Need a short docstring that describes what this is testing. "test_timer_for_suicide" is not meaningful to other people.

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -81,7 +89,54 @@ def start_test(self):
...
+        # in the caller SelectPoller.process_timeouts().
+        self.ioloop.remove_timeout( self.handle )
+        self.ioloop.stop()
+
+    # This patch delays the execution of the handlers so that they get fired
+    # in the same invocation of SelectPoller.process_timeouts().
+    @patch('pika.adapters.select_connection.SelectPoller.is_ready_to_fire',
+           side_effect= lambda x : 0<len(x) and
+                                   x[0][1]['deadline'] < time.time()-0.1 )
+    def test_timers_for_murder(self, mocked_is_ready):
```
vitaly-krugl: The use of "suicide" and "murder" in method names is a bit unpleasant. Could you use longer, more descriptive function names that don't involve those ominous words? In unittest, it's preferable to use longer, descriptive method names for test methods. For example: `test_timer_handler_deletes_own_timer` and `test_timer_handler_deletes_another_expired_timer`

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -33,8 +39,10 @@ def tearDown(self):
         self.ioloop.remove_timeout(self.fail_timer)
         self.ioloop = None
 
-    def start(self):
-        self.fail_timer = self.ioloop.add_timeout(self.TIMEOUT, self.on_timeout)
+    def start(self, callback = None):
```
vitaly-krugl: Let's revert this change (callback ...) in `start`. It's not necessary, and breaks the intended behavior of `raise AssertionError('Test timed out')` when the test times out. I will explain why it's not necessary in other comments in this code.

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -81,7 +89,54 @@ def start_test(self):
...
+    def on_timer_suicide(self, handle_holder):
+        self.assertEqual(self.handle, handle_holder.pop())
+        # This removal here should not raise exception by itself nor
+        # in the caller SelectPoller.process_timeouts().
+        self.ioloop.remove_timeout( self.handle )
+        self.ioloop.stop()
+
+    # This patch delays the execution of the handlers so that they get fired
+    # in the same invocation of SelectPoller.process_timeouts().
+    @patch('pika.adapters.select_connection.SelectPoller.is_ready_to_fire',
```
vitaly-krugl: Please get rid of `SelectPoller.is_ready_to_fire` from SelectPoller class, this patch, and the "This patch delays the execution..." comment above. They are not needed to test this functionality. The following should suffice to ensure that the timers will be ready in the same invocation of `SelectPoller.process_timeouts()`:

0. Initialize `self.deleted_another_timer = False`
1. Create timer 1: 0.01 seconds: callback is closure (function in function) `on_timer_that_deletes_another_ready_timer`. This callback should delete timer 2 and leave a breadcrumb in self (e.g., `self.deleted_another_timer = True`)
2. Create timer 2: 0.02 seconds: callback is closure `on_deleted_timer`. This callback should call `self.fail('deleted timer's callback called')`
3. Create timer 3: 0.03 seconds: callback is closure `on_conclude_another_timer_deletion`. This callback checks `self.assertTrue(self.deleted_another_timer)`, checks `self.assertNotIn(timer2_id, self.ioloop._timeouts)`, and calls `self.ioloop.stop()`.

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -81,7 +89,54 @@ def start_test(self):
...
+        self.timer_stack = list()
+        self.timer_a = self.ioloop.add_timeout(
+            0.05, partial(self.on_timer_murder, None))
+        self.timer_stack.append(None)
+
+        self.timer_b = self.ioloop.add_timeout(
+            0.04, partial(self.on_timer_murder, self.timer_a))
+        self.timer_stack.append(self.timer_a)
+
+        self.start( self.check_none_is_left )
```
vitaly-krugl: Change to `self.start()`

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -81,7 +89,54 @@ def start_test(self):
...
+
+    def check_none_is_left(self):
```
vitaly-krugl: Delete `check_none_is_left` - this timer callback is not needed any longer

pika/adapters/select_connection.py
```diff
@@ -264,21 +264,38 @@ def get_next_deadline(self):
...
-        for t in sorted(to_run, key=itemgetter('deadline')):
-            t['callback']()
-            del self._timeouts[hash(frozenset(t.items()))]
+        to_run = sorted([(k,timer) for (k,timer) in self._timeouts.items()
+                         if timer['deadline'] <= now],
+                        key = lambda item : item[1]['deadline'])
+
+        if self.is_ready_to_fire(to_run):
+            # Force recalculation as the oldest timer is going to be removed.
             self._next_timeout = None
```
vitaly-krugl: Bug: `self._next_timeout = None` used to be inside the `for t in sorted ...` loop. It's a bug to have it here outside of the loop, because we don't know whether `to_run` list is empty or not at this point. Make it the first statement in your new `for k, timer ...` loop to make sure that it will be executed if `to_run` is not empty:

```
for k, timer in to_run:
    self._next_timeout = None

    if k not in self._timeouts:
        # Prior invocation must have deleted this timer
        continue
    ...
```

pika/adapters/select_connection.py
```diff
@@ -264,21 +264,38 @@ def get_next_deadline(self):
...
         # be strictly necessary it preserves old behaviour when timeouts
         # were only run periodically.
-        for t in sorted(to_run, key=itemgetter('deadline')):
-            t['callback']()
-            del self._timeouts[hash(frozenset(t.items()))]
+        to_run = sorted([(k,timer) for (k,timer) in self._timeouts.items()
+                         if timer['deadline'] <= now],
+                        key = lambda item : item[1]['deadline'])
+
+        if self.is_ready_to_fire(to_run):
```
vitaly-krugl: Remove the `if self.is_ready_to_fire(to_run):` check. This pollutes the production code, and it's not necessary for making the unit tests work. See my comment on `test_timers_for_murder` that describes how to implement the test without this.

pika/adapters/select_connection.py
```diff
@@ -264,21 +264,38 @@ def get_next_deadline(self):
...
         """Process the self._timeouts event stack"""
 
         now = time.time()
-        to_run = [timer for timer in self._timeouts.values()
-                  if timer['deadline'] <= now]
-
         # Run the timeouts in order of deadlines. Although this shouldn't
         # be strictly necessary it preserves old behaviour when timeouts
         # were only run periodically.
-        for t in sorted(to_run, key=itemgetter('deadline')):
```
vitaly-krugl: Please run pylint and fix problems introduced by your change in both scripts. For example, since you replaced the only user of `itemgetter`, pylint will complain that import of `itemgetter` is unused.

vitaly-krugl:
@shinji-s: thanks for your latest round of changes and for implementing additional tests. Please see my new feedback.

pika/adapters/select_connection.py
shinji-s: How should I deal with the warnings like one below?
  Attribute 'timer_stack' defined outside __init__ (attribute-defined-outside-init)
If we initialize all attributes, that are need only temporarily during execution of certain test(s), then __init__ method gets too messy. Any suggestions?

Another thing is I don't see pylint complain about the use of assertEqual(). Will you please show me how you run pylint? Or the comment "Please check your changes with pylint and fix findings." isn't necessarily directed to the diff that the comment is attached to?

vitaly-krugl: Something like: `W0201: Attribute 'timer_stack' defined outside __init__` should be disabled globally in this script. With `unittest`, you typically don't declare your own `__init__` in the test case classes. Since this is normal behavior for unittest class implementations, disable this warning globally in the specific script. I tend to put the disable-warning pragmas between the module's docstring and the first import for visibility:

Using this script as an example:
```
# -*- coding: utf8 -*-
"""
Tests for SelectConnection IOLoops

"""
# pylint: disable=W0201

import logging

...
```

Please note that it is generally bad practice to disable pylint warnings globally like this, because that hides potential bugs. In most cases, you need to fix the code; in very special cases, where you really need to disable a pylint warning for a good reason, you would disable it on a case-by-case basis; you can put the pragma at the end of a statement to disable the warning on a specific line of code;  you can also use disable/enable pragmas around a block of code, if necessary.

vitaly-krugl: > Or the comment "Please check your changes with pylint and fix findings." isn't necessarily directed to the diff that the comment is attached to?

This comment was attached to the change:
```
+        self.handle = self.ioloop.add_timeout(
+            0.1, partial(self.on_timer_suicide,handle_holder))
```

I noticed lack of space after the comma here: `self.on_timer_suicide,handle_holder`

vitaly-krugl: @shinji-s: I am not sure what you're asking in this question:

> Another thing is I don't see pylint complain about the use of assertEqual().

Why would you expect pylint to coplain about `assertEqual().`?

shinji-s: @vitaly-krugl Thank you very much for your replies. I'll follow your advice of selectively disabling the warningings thought rather than using setattr() instead of regular assignment. As for my question on pylint complaining about assertEqual, it was my misunderstanding. My apology.

shinji-s:
@gmr , @vitaly-krugl 
Will you please give another consideration to the merge?

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -1,8 +1,13 @@
-# -*- coding: utf8 -*-
+# -*- coding: utf-8 -*-
 """
 Tests for SelectConnection IOLoops
 
 """
+# Disable warnings about initialization of members outside of __init__
+# pylint: disable=W0201
+# Disable warnings about too many public methods as they are in base classes
+# pylint: disable=R0904
```
vitaly-krugl: Thanks for adding the comments to explain the pylint pragmas.

pika/adapters/select_connection.py
```diff
@@ -264,20 +263,30 @@ def get_next_deadline(self):
...
-            t['callback']()
-            del self._timeouts[hash(frozenset(t.items()))]
-            self._next_timeout = None
+        to_run = sorted([(k, timer) for (k, timer) in self._timeouts.items()
```
vitaly-krugl: Unnecessary overhead: no need for the list `[(k, timer) ...`. `sorted` accepts generators/iterators, too. So, this would produce the equivalent result without the unnecessary overhead of instantiating/populating  the list:

```
to_run = sorted(((k, timer) for (k, timer) in self._timeouts.items()
                 if timer['deadline'] <= now),
                key=lambda item: item[1]['deadline'])
```

The original code had this problem, too, so fixing it in your PR is technically out of scope.

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -77,22 +84,89 @@ def set_timers(self):
...
+    def _on_timer_delete_itself(self, handle_holder):
+        """ A timeout hanlder that tries to remove itself. """
+        self.assertEqual(self.handle, handle_holder.pop())
+        # This removal here should not raise exception by itself nor
+        # in the caller SelectPoller.process_timeouts().
+        self.timer_got_called = True
+        self.ioloop.remove_timeout(self.handle)
+        self.ioloop.stop()
+
+    def test_timer_delete_anothoer(self):
```
vitaly-krugl: Spelling: `test_timer_delete_anothoer` --> `test_timer_delete_another`

vitaly-krugl:
@shinji-s: just a couple of nits, otherwise ready to go.

shinji-s:
@vitaly-krugl Thank you very much for your kind inspection. The typo is now fixed. The timer sorting code is left as it is.

vitaly-krugl:
LGTM

pika/adapters/select_connection.py
shinji-s: Thanks for pointing out sorted() can take a generator, which I was not aware of.
I think the code is better left as it is for now because it seems generator is not necessarily faster (especially because sorted() materialize generator as a list anyway?).
For our amusement, the script below generate following output. (The left side number is for generator and the right is for list.)

With threshold==0.01 (selectivity approx 1%)
Total = 0.4483 / 0.0092
Total = 0.3955 / 0.0099
With threshold==0.01 (selectivity approx 100%)
Total = 4.7233 / 2.2647
Total = 4.7220 / 2.2071

import time, random

def bench(l):
    start = time.time()
    sorted(l, key = lambda item: item[1]['k'])
    elapsed = time.time() - start
    return elapsed

def run(threshold):
    accum = {}
    for i in range(10):
        source = dict(map(lambda x: (random.random(), {'k':random.random()}),
                          range(0, 65536)))
        accum[1] = accum.get(1, 0.0) + \
            bench( (k,v) for (k,v) in source.items() if k<threshold )
        accum[2] = accum.get(2, 0.0) + \
            bench( [(k,v) for (k,v) in source.items() if k<threshold] )
    print 'Total = %.4f / %.4f' % (accum[1], accum[2])

    accum = {}
    for i in range(10):
        source = dict(map(lambda x: (random.random(), {'k':random.random()}),
                          range(0, 65536)))
        accum[2] = accum.get(2, 0.0) + \
            bench( [(k,v) for (k,v) in source.items() if k<threshold] )
        accum[1] = accum.get(1, 0.0) + \
            bench( (k,v) for (k,v) in source.items() if k<threshold )
    print 'Total = %.4f / %.4f' % (accum[1], accum[2])

if __name__=='__main__':
    print 'With threshold==0.01 (selectivity approx 1%)'
    run(0.01)
    print 'With threshold==0.01 (selectivity approx 100%)'
    run(1.0)

vitaly-krugl: Hmmm... interesting and somewhat unintuitive outcome :)

shinji-s:
@gmr, @vitaly-krugl Thank you guys for kindly guiding a github newbie.

gmr:
Thank you for the contribution!

PR #637, opened 2015-08-19
More comprehensive/relaxed handling of EINTR exception.

shinji-s:
Some polling mechanism on some platform (epoll on python2.7 on ubuntu14.04LTS) generates IOError rather than select.error or OSError.
Advice is needed as this patch lacks unit test and I can't figure out how I write test code that generates EINTR.

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -27,21 +27,36 @@ WRITE = 0x0004
 ERROR = 0x0008
 
 
-if pika.compat.PY2:
-    _SELECT_ERROR = select.error
-else:
-    # select.error was deprecated and replaced by OSError in python 3.3
-    _SELECT_ERROR = OSError
+# Reason for this unconventional dict initialization is the fact that on some
+# platforms select.error is an aliases for OSError. We don't want the lambda
+# for select.error to win over one for OSError.
+_SELECT_ERROR_CHECKERS = {}
+if pika.compat.PY3:
+    #InterruptedError is undefined in PY2
+    #pylint: disable=E0602
+    _SELECT_ERROR_CHECKERS[InterruptedError] = lambda e: True
+_SELECT_ERROR_CHECKERS[select.error] = lambda e: e.args[0] == errno.EINTR
+_SELECT_ERROR_CHECKERS[IOError] = lambda e: e.errno == errno.EINTR
+_SELECT_ERROR_CHECKERS[OSError] = lambda e: e.errno == errno.EINTR
 
 
-def _get_select_errno(error):
-    if pika.compat.PY2:
-        assert isinstance(error, select.error), repr(error)
-        return error.args[0]
+# We can reduce the number of elements in the list by looking at super-sub
+# class relationship because only the most generic ones needs to be caught.
+# For now the optimization is left out.
+# Following is better but still incomplete.
+#_SELECT_ERRORS = tuple(filter(lambda e: not isinstance(e, OSError),
+#                              _SELECT_ERROR_CHECKERS.keys())
+#                       + [OSError])
+_SELECT_ERRORS = tuple(_SELECT_ERROR_CHECKERS.keys())
+
+def _is_resumable(exc):
+    ''' Check if caught exception represents EINTR error.
+    :param exc: exception; must be one of classes in _SELECT_ERRORS '''
+    checker = _SELECT_ERROR_CHECKERS.get(exc.__class__, None)
+    if checker is not None:
+        return checker(exc)
     else:
-        assert isinstance(error, OSError), repr(error)
-        return error.errno
-
+        return False
 
 class SelectConnection(BaseConnection):
     """An asynchronous connection adapter that attempts to use the fastest
@@ -261,7 +276,7 @@ class SelectPoller(object):
             timeout = SelectPoller.POLL_TIMEOUT
 
         timeout = min((timeout, SelectPoller.POLL_TIMEOUT))
-        return timeout * SelectPoller.POLL_TIMEOUT_MULT
+        return timeout * self.POLL_TIMEOUT_MULT
 
 
     def process_timeouts(self):
@@ -399,8 +414,8 @@ class SelectPoller(object):
                                                    self._fd_events[ERROR],
                                                    self.get_next_deadline())
                 break
-            except _SELECT_ERROR as error:
-                if _get_select_errno(error) == errno.EINTR:
+            except _SELECT_ERRORS as error:
+                if _is_resumable(error):
                     continue
                 else:
                     raise
@@ -514,8 +529,8 @@ class KQueuePoller(SelectPoller):
                 kevents = self._kqueue.control(None, 1000,
                                                self.get_next_deadline())
                 break
-            except _SELECT_ERROR as error:
-                if _get_select_errno(error) == errno.EINTR:
+            except _SELECT_ERRORS as error:
+                if _is_resumable(error):
                     continue
                 else:
                     raise
@@ -589,8 +604,8 @@ class PollPoller(SelectPoller):
             try:
                 events = self._poll.poll(self.get_next_deadline())
                 break
-            except _SELECT_ERROR as error:
-                if _get_select_errno(error) == errno.EINTR:
+            except _SELECT_ERRORS as error:
+                if _is_resumable(error):
                     continue
                 else:
                     raise
--- a/pika/compat.py
+++ b/pika/compat.py
@@ -1,5 +1,5 @@
+import os
 import sys as _sys
-
 
 PY2 = _sys.version_info < (3,)
 PY3 = not PY2
@@ -103,3 +103,7 @@ def as_bytes(value):
         return value.encode('UTF-8')
     return value
 
+
+HAVE_SIGNAL = os.name == 'posix'
+
+EINTR_IS_EXPOSED = _sys.version_info[:2] <= (3,4)
--- a/tests/unit/select_connection_ioloop_tests.py
+++ b/tests/unit/select_connection_ioloop_tests.py
@@ -8,24 +8,34 @@ Tests for SelectConnection IOLoops
 # Disable warnings about too many public methods as they are in base classes
 # pylint: disable=R0904
 
-import logging
 try:
+    #pylint: disable=F0401
     import unittest2 as unittest
 except ImportError:
+    #pylint: disable=W0404
     import unittest
 
-import os
-import socket
+try:
+    #pylint: disable=F0401,E0611
+    from unittest.mock import patch as PATCH_
+except ImportError:
+    #pylint: disable=W0404
+    from mock import patch as PATCH_
+
 import errno
+import os
+import signal
+import socket
 import time
 import threading
 
 import pika
 from pika.adapters import select_connection
-from pika.adapters.select_connection import READ, WRITE, ERROR
+from pika.adapters.select_connection import READ, WRITE
 from functools import partial
 
 class IOLoopBaseTest(unittest.TestCase):
+    ''' Base for test classes in this uit. '''
 
     SELECT_POLLER = None
     TIMEOUT = 1.0
@@ -35,34 +45,43 @@ class IOLoopBaseTest(unittest.TestCase):
         self.ioloop = select_connection.IOLoop()
 
     def tearDown(self):
-        self.ioloop.remove_timeout(self.fail_timer)
         self.ioloop = None
 
     def start(self):
-        self.fail_timer = self.ioloop.add_timeout(self.TIMEOUT, self.on_timeout)
+        ''' Setup timeout handler for detecting 'no-activity'
+            and start polling. '''
+        fail_timer = self.ioloop.add_timeout(self.TIMEOUT, self.on_timeout)
+        self.addCleanup(self.ioloop.remove_timeout, fail_timer)
         self.ioloop.start()
 
     def on_timeout(self):
         """called when stuck waiting for connection to close"""
         # force the ioloop to stop
         self.ioloop.stop()
-        raise AssertionError('Test timed out')
+        self.fail('Test timed out')
 
 
 class IOLoopThreadStopTestSelect(IOLoopBaseTest):
+    ''' Test ioloop being stopped by another Thread. '''
     SELECT_POLLER = 'select'
     def start_test(self):
-        t = threading.Timer(0.1, self.ioloop.stop)
-        t.start()
+        ''' Starts a thread that stops ioloop after a while
+            and start polling '''
+        timer = threading.Timer(0.1, self.ioloop.stop)
+        self.addCleanup(timer.cancel)
+        timer.start()
         self.start()
 
 class IOLoopThreadStopTestPoll(IOLoopThreadStopTestSelect):
+    ''' Same as IOLoopThreadStopTestSelect but uses 'poll' syscall. '''
     SELECT_POLLER = 'poll'
 
 class IOLoopThreadStopTestEPoll(IOLoopThreadStopTestSelect):
+    ''' Same as IOLoopThreadStopTestSelect but uses 'epoll' syscall. '''
     SELECT_POLLER = 'epoll'
 
 class IOLoopThreadStopTestKqueue(IOLoopThreadStopTestSelect):
+    ''' Same as IOLoopThreadStopTestSelect but uses 'kqueue' syscall. '''
     SELECT_POLLER = 'kqueue'
 
 
@@ -160,12 +179,15 @@ class IOLoopTimerTestSelect(IOLoopBaseTest):
 
 
 class IOLoopTimerTestPoll(IOLoopTimerTestSelect):
+    ''' Same as IOLoopTimerTestSelect but uses 'poll' syscall '''
     SELECT_POLLER = 'poll'
 
 class IOLoopTimerTestEPoll(IOLoopTimerTestSelect):
+    ''' Same as IOLoopTimerTestSelect but uses 'epoll' syscall '''
     SELECT_POLLER = 'epoll'
 
 class IOLoopTimerTestKqueue(IOLoopTimerTestSelect):
+    ''' Same as IOLoopTimerTestSelect but uses 'kqueue' syscall '''
     SELECT_POLLER = 'kqueue'
 
 
@@ -174,29 +196,34 @@ class IOLoopSleepTimerTestSelect(IOLoopTimerTestSelect):
         fire in deadline order"""
 
     def start_test(self):
+        ''' Setup timers, sleep and start polling '''
         self.set_timers()
         time.sleep(self.NUM_TIMERS * self.TIMER_INTERVAL)
         self.start()
 
 
 class IOLoopSleepTimerTestPoll(IOLoopSleepTimerTestSelect):
+    ''' Same as IOLoopSleepTimerTestSelect but uses 'poll' syscall '''
     SELECT_POLLER = 'poll'
 
 class IOLoopSleepTimerTestEPoll(IOLoopSleepTimerTestSelect):
+    ''' Same as IOLoopSleepTimerTestSelect but uses 'epoll' syscall '''
     SELECT_POLLER = 'epoll'
 
 class IOLoopSleepTimerTestKqueue(IOLoopSleepTimerTestSelect):
+    ''' Same as IOLoopSleepTimerTestSelect but uses 'kqueue' syscall '''
     SELECT_POLLER = 'kqueue'
 
 class IOLoopSocketBaseSelect(IOLoopBaseTest):
-
+    ''' A base class for setting up a communicating pair of sockets. '''
     SELECT_POLLER = 'select'
     READ_SIZE = 1024
 
     def save_sock(self, sock):
-        fd = sock.fileno()
-        self.sock_map[fd] = sock
-        return fd
+        ''' Store 'sock' in self.sock_map and return the fileno.'''
+        fd_ = sock.fileno()
+        self.sock_map[fd_] = sock
+        return fd_
 
     def setUp(self):
         super(IOLoopSocketBaseSelect, self).setUp()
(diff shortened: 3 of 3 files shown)


pika/adapters/select_connection.py
```diff
@@ -27,20 +27,15 @@
 ERROR = 0x0008
 
 
-if pika.compat.PY2:
-    _SELECT_ERROR = select.error
-else:
-    # select.error was deprecated and replaced by OSError in python 3.3
-    _SELECT_ERROR = OSError
+_SELECT_ERRORS = (select.error, OSError, IOError)
```
vitaly-krugl: Do the changes pass all pylint checks?

pika/adapters/select_connection.py
```diff
@@ -27,20 +27,15 @@
...
+_select_error_checkers = {
```
vitaly-krugl: The use of spaces is non-standard. I would expect to see something like:

```
    select.error: lambda e: e.args[0] == errno.EINTR,
    OSError: lambda e: e.errno == errno.EINTR,
    IOError: lambda e: e.errno == errno.EINTR,
```

pika/adapters/select_connection.py
```diff
@@ -27,20 +27,15 @@
...
 
-
-def _get_select_errno(error):
-    if pika.compat.PY2:
-        assert isinstance(error, select.error), repr(error)
-        return error.args[0]
-    else:
-        assert isinstance(error, OSError), repr(error)
-        return error.errno
+def _is_resumable(exc):
```
vitaly-krugl: This could benefit from a docstring, taking care to document the arg; something like

```
:param exc: exception; must be of one of the exception classes in _SELECT_ERRORS
```

vitaly-krugl:
@shinji-s: Thanks for staying on top of this.

Regarding
> ... I can't write a test code that generates EINTR exception. Could you provide direction on how I proceed with the issue?

You would need to do something like this in the test methods:

1. Create a connected socket pair. `get_interrupt_pair` in `select_connection.py` is an example of how to do this in a portable way (you cannot pass a pipe to select on Windows). NOTE: Be mindful that some developers will be running the tests on Windows, so your SelectPoller EINTR test will need to be compatible with Windows, whatever that entails. You will use the poller on one of the sockets and your signal handler function will send a character via the other socket.
2. Register a handler for `signal.SIGUSR1`; your signal handler function should set a flag that it was called, then send a character on the "sending" socket of the above-created socket pair and return (see https://docs.python.org/2/library/signal.html). Sending the character will cause the poller's `poll()` method to call your socket event handler and return quickly. Wrap the rest of the test method's code in try/finally and unregister the signal handler in `finally` so that it won't interfere with other tests.
3. Instantiate the poller class to be tested
4. Call the poller's `add_handler` method to add a handler for the `fileno()` of one of the connected sockets from the above step (but don't send any data at this point, so that the `poll` method would block). Use `select_connection.READ` as the event mask. The handler function that you register via `add_handler` should assert to make sure that it was called with the expected `select_connection.READ` event and also set a flag to record that it was called.
5. Call the poller's `add_timeout` method to register a timeout function; e.g., 1 second. Your timeout handler function should raise an exception that causes the test to fail; the exception should have a meaningful text message that helps to debug.
6. Immediately **before calling the poller's `poll()` method**, start a thread that will use `os.kill()` to send the signal `signal.SIGUSR1` to the test's process (see `os.getpid()`). The thread function should sleep for a little bit (e.g., 0.1 sec) BEFORE calling `os.kill()` to give the main thread enough time to call the poller's `poll()` method. Wrap the `poll()` call in try/finally, and join the thread object in the `finally` to make sure that the thread will be stopped inside this test method and won't impact other tests.
7. Assert that your `signal.SIGUSR1` handler and  `select_connection.READ` handler functions were called.

gmr:
@shinji-s do you have time to address these items? We'd like to see this included with 0.10.0 and is the last gating item for that release.

shinji-s:
@gmr
Thanks to @vitaly-krugl 's direction, I've managed to cook up small code that demonstrate the issue.
Unfortunately, however, I'm on a ferry to another island to start my vacation there. I'm afraid I'm not going to make much progress in coming two weeks.

The output from the script is as follows. I think the issue has more to do with implementatiion select.epoll in python 2.7 than with ubuntu.

Testing <function select_poller at 0xb77203e4>
  File "test.py", line 70, in run_test_on
    poll_using(poller)
  File "test.py", line 35, in poll_using
    poller(sock)
  File "test.py", line 40, in select_poller
    select.select([sock], [], []) \
error: (4, 'Interrupted system call')


Testing <function epoll_poller at 0xb7720dbc>
  File "test.py", line 70, in run_test_on
    poll_using(poller)
  File "test.py", line 35, in poll_using
    poller(sock)
  File "test.py", line 46, in epoll_poller
    ep.poll()
IOError: [Errno 4] Interrupted system call

Here is the test code.

import threading, socket, time, signal, sys, select, os, traceback

PORT = 8181
SIGNAL_TO_USE = signal.SIGUSR1

class KillerThread(threading.Thread):
    def __init__(self, initial_wait, *args, **kw):
        self.initial_wait = initial_wait
        super(KillerThread, self).__init__(*args, **kw)
        
    def run(self):
        time.sleep(self.initial_wait)
        os.kill(os.getpid(), SIGNAL_TO_USE)

class ServerThread(threading.Thread):
    def run(self):
        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        sock.bind(('localhost', PORT))
        sock.listen(8)
        sk, addr = sock.accept()
        try:
            while 1:
                b = sk.recv(4096)
                if not b:
                    break
        finally:
            sk.close()

        
def poll_using(poller):
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    try:
        sock.connect(('localhost', PORT))
        poller(sock)
    finally:
        sock.close()

def select_poller(sock):
    select.select([sock], [], [])


def epoll_poller(sock):
    ep = select.epoll()
    ep.register(sock, select.EPOLLIN)
    ep.poll()

    
def signal_handler(signum, interrupted_stack):
    # print 'Got signal-%s. Ignoiring' % signum
    pass

        

def run_test_on(poller):
    th_list = []
    try:
        k = KillerThread(0.2)
        k.start()
        th_list.append(k)
        s = ServerThread()
        s.start()
        th_list.append(s)

        time.sleep(0.1)
        old_handler = signal.signal(SIGNAL_TO_USE, signal_handler)
        try:
            print 'Testing %s' % poller
            try:
                poll_using(poller)
            except KeyboardInterrupt, SystemExit:
                raise
            except:
                traceback.print_exc()
                print
                print
        finally:
            signal.signal(SIGNAL_TO_USE, old_handler)
    finally:
        for th in th_list:
            th.join()
    
if __name__=='__main__':
    run_test_on( select_poller )
    run_test_on( epoll_poller )

shinji-s:
@gmr
Thanks to @vitaly-krugl 's direction, I've managed to cook up small code that demonstrate the issue.
Unfortunately, however, I'm on a ferry to another island to start my vacation there. I'm afraid I'm not going to make much progress in coming two weeks.

The output from the script is as follows. I think the issue has more to do with implementatiion select.epoll in python 2.7 than with ubuntu.

Testing
File "test.py", line 70, in run_test_on
poll_using(poller)
File "test.py", line 35, in poll_using
poller(sock)
File "test.py", line 40, in select_poller
select.select([sock], [], []) \
error: (4, 'Interrupted system call')

Testing
File "test.py", line 70, in run_test_on
poll_using(poller)
File "test.py", line 35, in poll_using
poller(sock)
File "test.py", line 46, in epoll_poller
ep.poll()
IOError: [Errno 4] Interrupted system call

Here is the test code.
```python
import threading, socket, time, signal, sys, select, os, traceback

PORT = 8181
SIGNAL_TO_USE = signal.SIGUSR1

class KillerThread(threading.Thread):
    def __init__(self, initial_wait, *args, **kw):
        self.initial_wait = initial_wait
        super(KillerThread, self).__init__(*args, **kw)
        
    def run(self):
        time.sleep(self.initial_wait)
        os.kill(os.getpid(), SIGNAL_TO_USE)

class ServerThread(threading.Thread):
    def run(self):
        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        sock.bind(('localhost', PORT))
        sock.listen(8)
        sk, addr = sock.accept()
        try:
            while 1:
                b = sk.recv(4096)
                if not b:
                    break
        finally:
            sk.close()

        
def poll_using(poller):
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    try:
        sock.connect(('localhost', PORT))
        poller(sock)
    finally:
        sock.close()

def select_poller(sock):
    select.select([sock], [], [])


def epoll_poller(sock):
    ep = select.epoll()
    ep.register(sock, select.EPOLLIN)
    ep.poll()

    
def signal_handler(signum, interrupted_stack):
    # print 'Got signal-%s. Ignoiring' % signum
    pass

        

def run_test_on(poller):
    th_list = []
    try:
        k = KillerThread(0.2)
        k.start()
        th_list.append(k)
        s = ServerThread()
        s.start()
        th_list.append(s)

        time.sleep(0.1)
        old_handler = signal.signal(SIGNAL_TO_USE, signal_handler)
        try:
            print 'Testing %s' % poller
            try:
                poll_using(poller)
            except KeyboardInterrupt, SystemExit:
                raise
            except:
                traceback.print_exc()
                print
                print
        finally:
            signal.signal(SIGNAL_TO_USE, old_handler)
    finally:
        for th in th_list:
            th.join()
    
if __name__=='__main__':
    run_test_on( select_poller )
    run_test_on( epoll_poller )
```python

**Comment by shinji-s** [comment]
Output was messed up.

```
Testing <function select_poller at 0xb77203e4>
  File "test.py", line 70, in run_test_on
    poll_using(poller)
  File "test.py", line 35, in poll_using
    poller(sock)
  File "test.py", line 40, in select_poller
    select.select([sock], [], [])
error: (4, 'Interrupted system call')


Testing <function epoll_poller at 0xb7720dbc>
  File "test.py", line 70, in run_test_on
    poll_using(poller)
  File "test.py", line 35, in poll_using
    poller(sock)
  File "test.py", line 46, in epoll_poller
    ep.poll()
IOError: [Errno 4] Interrupted system call
```

**Comment by gmr** [comment]
@vitaly-krugl waiting on your +1 for merge.

**Comment by shinji-s** [comment]
Although I have pylint related fixes and test cases tha need a little brushing up pending in my workspace, I don't mind the fix above to be merged now if you guys sees it's ok. +1 from me. I'll commit&push pending changes and send another PR later (Perhaps comming Mon or Tue?).

**Comment by vitaly-krugl** [comment]
@gmr, we're still waiting for @shinji-s to add the tests to this PR.

**Comment by vitaly-krugl** [comment]
@shinji-s, let's get the tests into this PR before merging.

**Comment by shinji-s** [comment]
@gmr @vitaly-krugl  I will commit pending test cases asap when I get back home on Sep. 9. I may need your help to get tests to be Windows 'compatible'.

**Comment by vitaly-krugl** [comment]
@shinji-s - thanks!

**Comment by shinji-s** [comment]
Sorry, pushed the last commit too early. Now fixing the problem.

**Comment by shinji-s** [comment]
@gmr, @vitaly-krugl Turned out that InterruptedError as well needs to be handled in PY3. The test case should be ready for your review. Don't know yet how it behaves in  Windows envrionment though.
Please note that 300be19 is a fundamental change that requires special attention.

File: pika/compat.py [hunk 4]
```diff
@@ -1,5 +1,5 @@
 import sys as _sys
-
+import platform
```
vitaly-krugl: Should be `import os` and place it before `import sys ...` (alphabetically)

pika/compat.py
```diff
@@ -103,3 +103,5 @@ def as_bytes(value):
         return value.encode('UTF-8')
     return value
 
+
+HAVE_SIGNAL = platform.os.name == 'posix'
```
vitaly-krugl: Should be `os.name` instead of `platform.os.name`

pika/adapters/select_connection.py
```diff
@@ -27,20 +27,25 @@
...
-        assert isinstance(error, select.error), repr(error)
-        return error.args[0]
-    else:
-        assert isinstance(error, OSError), repr(error)
-        return error.errno
+_SELECT_ERROR_CHECKERS = {
+    select.error : lambda e: e.args[0] == errno.EINTR,
+    OSError      : lambda e: e.errno == errno.EINTR,
+    IOError      : lambda e: e.errno == errno.EINTR,
+    }
```
vitaly-krugl: On Python 3.x (tested on 3.5), they are all OSError:

```
In [12]: _SELECT_ERROR_CHECKERS = {
    select.error : "select.error",
    OSError      : "OSError",
    IOError      : "IOError"
    }
In [13]: _SELECT_ERROR_CHECKERS
Out[13]: {OSError: 'IOError'}
```

pika/adapters/select_connection.py
```diff
@@ -27,20 +27,25 @@
 ERROR = 0x0008
 
 
-if pika.compat.PY2:
-    _SELECT_ERROR = select.error
-else:
-    # select.error was deprecated and replaced by OSError in python 3.3
-    _SELECT_ERROR = OSError
 
+_SELECT_ERRORS = (select.error, OSError, IOError)
```
vitaly-krugl: To guarantee consistency, `_SELECT_ERRORS` should be based on `_SELECT_ERROR_CHECKERS`. E.g.,

```
_SELECT_ERRORS = pika.compat.dictkeys(_SELECT_ERROR_CHECKERS)
```

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -10,22 +10,33 @@
 
 import logging
 try:
+    #pylint: disable=F0401
```
vitaly-krugl: Need to put all these pylint pragmas on the same line as the imports, so they don't disable checks on all the other imports

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -35,10 +46,13 @@ def setUp(self):
         self.ioloop = select_connection.IOLoop()
 
     def tearDown(self):
-        self.ioloop.remove_timeout(self.fail_timer)
+        if hasattr(self, 'fail_timer'): # some test don't use this timer.
```
vitaly-krugl: Don't need any of this `remove_timeout` logic here at all. Instead, use addCleanup to register cleanup call after creating fail timer in `start()`. This make the code more readable. See my comment in `start()`

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -35,10 +46,13 @@ def setUp(self):
...
+            self.ioloop.remove_timeout(self.fail_timer)
         self.ioloop = None
 
     def start(self):
+        ''' Setup timeout handler for detecting 'no-activity'
+            and start polling. '''
         self.fail_timer = self.ioloop.add_timeout(self.TIMEOUT, self.on_timeout)
```
vitaly-krugl: Instead of the messy remove_timeout logic in `tearDown`, do this instead:

```
        fail_timer = self.ioloop.add_timeout(self.TIMEOUT, self.on_timeout)
        self.addCleanup(self.ioloop.remove_timeout, fail_timer)
```

fail_timer can no be a local variable and there is no need to create the instance member `self.fail_timer`

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -174,26 +197,31 @@ class IOLoopSleepTimerTestSelect(IOLoopTimerTestSelect):
...
         time.sleep(self.NUM_TIMERS * self.TIMER_INTERVAL)
         self.start()
 
 
 class IOLoopSleepTimerTestPoll(IOLoopSleepTimerTestSelect):
+    ''' Same as IOLoopSleepTimerTestSelect but uses 'poll' syscall '''
     SELECT_POLLER = 'poll'
 
 class IOLoopSleepTimerTestEPoll(IOLoopSleepTimerTestSelect):
+    ''' Same as IOLoopSleepTimerTestSelect but uses 'epoll' syscall '''
```
vitaly-krugl: Thanks for adding the docstrings

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -229,20 +259,31 @@ def create_write_socket(self, on_connected):
         return write_sock
 
     def do_accept(self, fd, events, write_only):
+        ''' Create socket from the given fd and setup 'read' handler '''
         self.assertEqual(events, READ)
+        self.assertTrue(write_only is not None)
```
vitaly-krugl: Instead: `self.assertIsNotNone(write_only)`

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -229,20 +259,31 @@ def create_write_socket(self, on_connected):
...
         listen_sock = self.sock_map[fd]
         read_sock, _ = listen_sock.accept()
         fd = self.save_sock(read_sock)
         self.ioloop.add_handler(fd, self.do_read, READ)
 
-    def connected(self, fd, events, write_only):
+    def connected(self, _fd, _events, write_only):
+        ''' Create socket from given _fd and respond to 'connected'.
+            Implemenation is subclass's responsibility. '''
+        self.assertTrue(self is not None) # To silence pylint
```
vitaly-krugl: Use a `#pylint: disable=...` pragma instead

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -252,39 +293,140 @@ def on_timeout(self):
...
 class IOLoopSimpleMessageTestCasetEPoll(IOLoopSimpleMessageTestCaseSelect):
+    ''' Same as IOLoopSimpleMessageTestCaseSelect but uses 'epoll' syscall '''
     SELECT_POLLER = 'epoll'
 
 class IOLoopSimpleMessageTestCasetKqueue(IOLoopSimpleMessageTestCaseSelect):
+    ''' Same as IOLoopSimpleMessageTestCaseSelect but uses 'kqueue' syscall '''
+    SELECT_POLLER = 'kqueue'
+
+
+class DoAfterDelayThread(threading.Thread):
```
vitaly-krugl: This appears to implement the same functionality as `threading.Timer` class. Use `threading.Timer` in stead of implementing this class. See https://docs.python.org/2/library/threading.html#timer-objects

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -252,39 +293,140 @@ def on_timeout(self):
...
+    def __init__(self, initial_wait, func, *args, **kw):
+        self.initial_wait = initial_wait
+        self.func = func
+        super(DoAfterDelayThread, self).__init__(*args, **kw)
+
+    def run(self):
+        time.sleep(self.initial_wait)
+        self.func()
+
+class IOLoopEintrTestCaseSelect(IOLoopBaseTest):
```
vitaly-krugl: Why derive from `IOLoopBaseTest`? This class is not using any of the core functionality from `IOLoopBaseTest` except the value of TIMEOUT, so it's not a "IOLoopBaseTest". Just derive from `unittest.TestCase` and reference `IOLoopBaseTest.TIMEOUT` where you need to use the same TIMEOUT.

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -252,39 +293,140 @@ def on_timeout(self):
...
+        is_resumable_mock.side_effect = is_resumable_raw
+        self.poller = self.ioloop._get_poller()
+        sockpair = self.poller.get_interrupt_pair()
+        self._eintr_read_handler_is_called = False
+        self.poller.add_handler(sockpair[0].fileno(), self._eintr_read_handler,
+                                READ)
+        self.poller.add_timeout(self.TIMEOUT, self._eintr_test_fail)
+        assert getattr(self, 'original_signal_handler', None) is None
+        original_signal_handler = \
+            signal.signal(signal.SIGUSR1, self.signal_handler)
```
vitaly-krugl: It's more obvious and reliable to schedule cleanup right after registering the signal instead of having the cleanup code somewhere else:

`self.addCleanup(signal.signal, signal.SIGUSR1, original_signal_handler)`

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -252,39 +293,140 @@ def on_timeout(self):
...
+        killer = writer = None
+        try:
+            killer = DoAfterDelayThread(
+                0.1, lambda: os.kill(os.getpid(), signal.SIGUSR1))
+            writer = DoAfterDelayThread(
+                0.2, lambda: sockpair[1].send(self.MSG_CONTENT))
+            killer.start()
+            writer.start()
```
vitaly-krugl: Use `threading.Timer` instead and no need for the `join` calls in finally:

```
killer = threading.Timer(0.1, lambda: os.kill(os.getpid(), signal.SIGUSR1))
self.addCleanup(killer.cancel)
writer = threading.Timer(0.2, lambda: sockpair[1].send(self.MSG_CONTENT))
self.addCleanup(writer.cancel)
killer.start()
writer.start()
```

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -252,39 +293,140 @@ def on_timeout(self):
...
+        mesg = sock.recv(256)
+        self.assertEqual(mesg, self.MSG_CONTENT)
+        self.poller.stop()
+        self._eintr_read_handler_is_called = True
+
+    def _eintr_test_fail(self):
+        '''This function gets called when eintr-test failed to get
+           _eintr_read_handler called.'''
+        self.poller.stop()
+        raise AssertionError('Eintr-test timed out')
```
vitaly-krugl: Use the following standard unittest functionality instead of raising AssertionError explicitly: `self.fail('Eintr-test timed out')`

vitaly-krugl:
@shinji-s: thank you for adding the tests and additional exceptions. I reviewed the changes and left feedback for you here.

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -252,39 +293,140 @@ def on_timeout(self):
...
+
+    #Disable warning about internal function access.
+    #pylint: disable=W0212
```
vitaly-krugl: Don't use the pylint pragma globally, because it disables the checks in other code. Use it on the affected functions or code lines directly instead.

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -252,39 +293,140 @@ def on_timeout(self):
...
+    @PATCH_('pika.adapters.select_connection._is_resumable')
+    def test_eintr(self, is_resumable_mock, is_resumable_raw=
+                   pika.adapters.select_connection._is_resumable):
+        '''Test that poll() is properly restarted after receiving EINTR error.
+           Class of an exception raised to signal the error differs in one
+           implementation of polling mechanism and another.'''
+        if not pika.compat.HAVE_SIGNAL:
+            return
```
vitaly-krugl: Use unittest.skipUnless decorator instead of this check/return. It will automatically print a nice message instead of quietly skipping the test:

```
@unittest.skipUnless(pika.compat.HAVE_SIGNAL, 'This platform doesn't support posix signals')
@PATCH_('pika.adapters.select_connection._is_resumable')
def test_eintr(self, is_resumable_mock, is_resumable_raw=
                   pika.adapters.select_connection._is_resumable):
```

pika/adapters/select_connection.py
shinji-s: And we are supposed to get no EINTR exceptions raised in Python 3.5 as the standard library performs retries. Some of the testes here should fail.

I'm afraid I don't get what changes you are suggesting here.
Since _SELECT_ERRORS_CHECKERS is indexed with most-derived class of a caught exception and on some platform the classes are different, we can't reduce _SELECT_ERROR_CHECKERS to the one-element dictionary. Also the values of _SELECT_ERROR_CHECKERS should be callable.

pika/adapters/select_connection.py
shinji-s: Thanks. Since I made the last push I realized that not all classes in _SELECT_ERROR_CHECKS needs to be in _SELECT_ERRORS if one is a subclass of another, only most generic ones need to be caught. I'll leave the optimization out for know and follow your suggestion.

tests/unit/select_connection_ioloop_tests.py
shinji-s: I understand that pylint error/warning suppression is effective only in the block it appears.
And I think it's ok or desirable to ignore warnings about failed imports after a failed import.

tests/unit/select_connection_ioloop_tests.py
shinji-s: Thanks. This made the code much lean and clean.

tests/unit/select_connection_ioloop_tests.py
shinji-s: It depends on IOLoopBaseTest.setUp() to have self.ioloop created based upon the mechanism specifid by select.SELECT_TYPE.

shinji-s:
@vitaly-krugl Thank you for your feedback. I think I've applied most of the fixes you suggested except one on changing _SELECT_ERROR_CHECKER and one on test class derivation. Please let me know if there are some that I missed.

pika/adapters/select_connection.py
vitaly-krugl: @shinji-s: sorry for the confusion. What I was concerned about is what would happen on python 3.x, where all those different error classes are reduced to the single OSError class. I don't want `select.error : lambda e: e.args[0] == errno.EINTR` to win and be the only one left at the end, because we should be looking at `e.errno` in OSError, not `e.args[0]`. I am hoping that python evaluates the dict properties in order so that in our case, the last one in `_SELECT_ERROR_CHECKERS` would win out.

The experiment in my comment was performed via https://www.python.org/shell/ (python 3.4.0) to demonstrate the collapsing of classes, and which property wins out.

If python guarantees the processing of the dict properties in the specified order, then we should probably have `OSError      : lambda e: e.errno == errno.EINTR` at the end to make that explicit and also add a comment in `_SELECT_ERROR_CHECKERS` that explains that "all these error classes are reduced to the single OSError class in 3.x"

vitaly-krugl: @shinji-s: Regarding
> And we are supposed to get no EINTR exceptions raised in Python 3.5 as the standard library performs retries. Some of the testes here should fail.

Let's add the appropriate `unittest.skip*` decorator to your test(s) to make sure that the one(s) that are not appropriate for Python 3.5 won't run on 3.5 and later.

tests/unit/select_connection_ioloop_tests.py
```diff
@@ -35,34 +46,43 @@ def setUp(self):
...
 class IOLoopThreadStopTestSelect(IOLoopBaseTest):
+    ''' Test ioloop being stopped by another Thread. '''
     SELECT_POLLER = 'select'
     def start_test(self):
-        t = threading.Timer(0.1, self.ioloop.stop)
-        t.start()
+        ''' Starts a thread that stops ioloop after a while
+            and start polling '''
+        # Pylint wants ids at least 3chars long. That's the reason for 't__'
+        t__ = threading.Timer(0.1, self.ioloop.stop)
```
vitaly-krugl: @shinji-s: `t__` is not a good idea; it's not conventional. If you really needed a longer variable, it would have been better to change it to `timer` in this case. However, since the variable in this case is only used to start the timer, you could get rid of it altogether: `threading.Timer(0.1, self.ioloop.stop).start()` is all you need here

tests/unit/select_connection_ioloop_tests.py
vitaly-krugl: Thanks @shinji-s. I stand corrected. I also discovered that with my pylint version 1.4.4, pylint is smart about the ImportError exception. In the following code block, it only complains F0401 about `import halbhalbhalb`, but doesn't complain about `import blahblahblah`

```
try:
  import blahblahblah
except ImportError:
  import datetime

import halbhalbhalb
```

vitaly-krugl:
@shinji-s, thank you for following up. I noted a few more issues. We're almost there!

pika/adapters/select_connection.py
```diff
@@ -27,20 +27,25 @@
...
-        return error.args[0]
-    else:
-        assert isinstance(error, OSError), repr(error)
-        return error.errno
+if pika.compat.PY3:
+    # InterruptedError is undefined in PY2
+    #pylint: disable=E0602
+    _SELECT_ERROR_CHECKERS[InterruptedError] = lambda e: True
+
+# We can reduce the number of elements
```
vitaly-krugl: @shinji-s: without our discussion, I don't think I would understand the implication of this comment. Perhaps this comment is not needed.

On 3.x, the number of elements will get reduced automatically, because `_SELECT_ERROR_CHECKERS` will get collapsed to just the two properties: `OSError` and `InterruptedError`

pika/adapters/select_connection.py
```diff
@@ -27,20 +27,25 @@
...
-    if pika.compat.PY2:
-        assert isinstance(error, select.error), repr(error)
-        return error.args[0]
-    else:
-        assert isinstance(error, OSError), repr(error)
-        return error.errno
+if pika.compat.PY3:
+    # InterruptedError is undefined in PY2
+    #pylint: disable=E0602
+    _SELECT_ERROR_CHECKERS[InterruptedError] = lambda e: True
```
vitaly-krugl: @shinji-s: 

```
In [2]: isinstance(InterruptedError(), OSError)
Out[2]: True
In [3]: InterruptedError.__bases__
Out[3]: (OSError,)
```

Do we need `InterruptedError` at all?

pika/adapters/select_connection.py
shinji-s: Thank you for the detailed explanation. Now I get what you meant and see that it is a valid concern.
I'll use a sequence of statement to initialize the dict so that precedence is more obvious.

pika/adapters/select_connection.py
shinji-s: I guess we do.
In python3.4 on Windows;
>>> e = InterruptedError()
>>> type(e)
<class 'InterruptedError'>
>>> isinstance(e, OSError)
True

Also please search for nose.proxy.KeyError: <class 'InterruptedError'> in the failed logs at
https://travis-ci.org/pika/pika/builds/79073409

pika/adapters/select_connection.py
shinji-s: Thanks for pointing it out. I've left the comment half cooked. I'll make it more detailed.

shinji-s: This discussion lead to me notice a bug in _is_resumable() implementation. It is possible that an exception which is a subclass of a classe in _SELECT_ERRORS but whose actual class is not listed in _SELECT_ERROR_CHECKERS. That exception needs to be reraised rather than causing KeyError when _SELECT_ERROR_CHECKERS is indexed.

tests/unit/select_connection_ioloop_tests.py
shinji-s: Changed the name to ```timer``` and installed a cleanup handler.

pika/adapters/select_connection.py
shinji-s: As for python 3.5 compatibility, the test (I wrote 'tests' before but there is only one test that exercise EINTR handling. my bad) should run except that EINTR exception does not get handled. cf. https://www.python.org/dev/peps/pep-0475/
I introduced pika.compat.EINTR_IS_EXPOSED flag and asserts on _is_resumable_mock.call_count accordingly.

pika/adapters/select_connection.py
vitaly-krugl: good point!

vitaly-krugl:
@shinji-s: awesome, thank you! Please rebase/squash against master, so that there will be a single commit instead of 15; and we should be good to go after that. Best.

shinji-s:
@vitaly-krugl Thank you for your directions.
I got new commit message for  'git rebase -i HEAD~15, sqush on all lines except the first and git puph'  but the old commit messages too are still in 'git log'. Did I do something wrong? Maybe it's better if I fork from pika/pika again and apply the diffs?

shinji-s:
The culprit must be the 'git pull' that I had to perform (without thinking about what it would do) in order to have push accepted.

shinji-s:
I think I've manged to fix the history.

vitaly-krugl:
@shinji-s: I usually run `git rebase -i master` if my branch's base was in local master or `git rebase -i upstream/master` if my branch's base was from upstream/master

vitaly-krugl:
@shinji-s: LGTM @ 59dfa06cb18ce53e12979547312941eba04b0a6d

CC @gmr

shinji-s:
> git rebase -i master
@vitaly-krugl  That looks more reliable than rebase -i HEAD~n.
Thank you for the tip and review of the PR.

vitaly-krugl:
@gmr, whenever you get a chance, I think we're good to merge this.

gmr:
Looks like this is broken in Python 3.5 -- if you can look into why, that'd be ideal.

shinji-s:
I guess the cause is python 3.5 absorbing EINTR and retries the system call within the library without raising an exception to the application. I'll have a closer look later today or tomorrow.

shinji-s:
There was a typo in a Python3.5 specific portion of the code.
I've sent a new PR with the fix. Please review.

gmr:
Thanks!

PR #644, opened 2015-09-02
Add constants for delivery mode options

steveYeah:
I know this is a small change, but I sometimes forget what `delivery_mode=2` means, and I have definitely been asked more time than I care to mention.

I think these constants would make things a lot clearer to those who don't visit their queue managers often, and newcomers alike.

### DIFF
--- a/pika/spec.py
+++ b/pika/spec.py
@@ -2067,6 +2067,9 @@ class BasicProperties(amqp_object.Properties):
     FLAG_APP_ID = (1 << 3)
     FLAG_CLUSTER_ID = (1 << 2)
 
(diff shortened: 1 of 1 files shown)


gmr:
I like the idea, but unfortunately this will be overwritten when the codegen is run. Please modify https://github.com/pika/pika/blob/master/utils/codegen.py#L283 to get your changes in.

gmr:
You'll need to reimplement this on top of the changes in master, so I'm going to close this and look for a 2nd PR.  Thanks :)

steveYeah:
OK, thanks. I haven't worked with codegen before, so not sure what to update, but will take a look and add another PR soon

PR #651, opened 2015-10-08
Use print function in examples and docstrings

mivade:


docs/examples/blocking_consume.rst
```diff
@@ -13,10 +13,10 @@ Example of consuming messages and acknowledging them::
...
 
         def on_message(channel, method_frame, header_frame, body):
-            print method_frame.delivery_tag
-            print body
-            print
-            channel.basic_ack(delivery_tag=method_frame.delivery_tag)
+            print(method_frame.delivery_tag)
+            print(body)
+            print(
+	        channel.basic_ack(delivery_tag=method_frame.delivery_tag))
```
vitaly-krugl: Why are you printing the result of basic_ack? Does it return anything interesting? No. That's not what the original code did.

vitaly-krugl:
@gmr: while this fixes up the examples to work for python 3.x, it breaks the examples for users of python 2.7.x, since the print function is not available there by default.

docs/examples/blocking_consume.rst
mivade: Thanks. Fixed.

mivade:
@vitaly-krugl: At least on the legacy Python builds I have tried, the print function Just Works:

```
Python 2.7.9 (default, Mar  1 2015, 12:57:24) 
[GCC 4.9.2] on linux2
>>> print('test')
test
```

```
Python 2.6.9 |Continuum Analytics, Inc.| (unknown, Aug 21 2014, 18:28:52) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux2
>>> print('test')
test
```

This is despite the [documentation](https://docs.python.org/2.6/library/functions.html#print) seemingly saying that you must add `from __future__ import print_statement`. Anecdotally, I seem to recall in earlier builds of 2.7 explicitly having to use this, but I don't remember the details of what specific version it was.

gmr:
This seems like the correct behavior to me. Thanks!

PR #656, opened 2015-10-21
Enh: Process _open_error_result before _flush_output

### LINKED ISSUE #655: BlockingConnection and connection times out + connection_attempts
I've been playing with BlockingConnection recently.

Have found the following : 

If the server isn't *available* (as with `sudo iptables -A INPUT -p tcp --destination-port amqp  -j DROP`) here is what we can see:

```
pika.adapters.base_connection - INFO - Connecting to 127.0.0.1:5672
pika.adapters.base_connection - ERROR - Connection to 127.0.0.1:5672 failed: timeout
```

the socket.connect method raise a socket.error which is catched by `SelectConnection._create_and_connect_to_socket()` and then the error is returned by that function, higher in the stack the error is finally given to `self._values = (self._value_class(*args, **kwargs),)` within `CallbackManager.process()`

but what we get as exception is `ConnectionClosed()` (this is raised from within `BlockingConnection._flush_output` (called from _process_io_for_connection_setup)).

Now, if I set a `connection_attemps > 1` to the BlockingConnection parameters then I get well the socket.error timeout exception..

Shouldn't we also get it for connection_attemps == 1 (which is actually the default) ? 

this would be more consistant.

gst:
This permit to correctly handle a timeout error (whatever is the connection_attemps parameter value), or, normally, any other kind of connection error.

This should fix #655

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -370,8 +370,9 @@ class BlockingConnection(object):  # pylint: disable=R0902
 
         :raises AMQPConnectionError: on connection open error
         """
-        self._flush_output(self._opened_result.is_ready,
-                           self._open_error_result.is_ready)
+        if not self._open_error_result.ready:
+            self._flush_output(self._opened_result.is_ready,
+                               self._open_error_result.is_ready)
 
         if self._open_error_result.ready:
             exception_or_message = self._open_error_result.value.error
--- a/tests/unit/blocking_connection_tests.py
+++ b/tests/unit/blocking_connection_tests.py
@@ -5,6 +5,8 @@ Tests for pika.adapters.blocking_connection.BlockingConnection
 """
 import socket
 
+from pika.exceptions import AMQPConnectionError
+
 try:
     from unittest import mock
     patch = mock.patch
@@ -203,4 +205,19 @@ class BlockingConnectionTests(unittest.TestCase):
                 spec_set=blocking_connection.BlockingConnection._flush_output):
             connection.sleep(0.00001)
 
-
+    def test_connection_attempts_with_timeout(self):
+        # for whatever conn_attempt we try:
+        for conn_attempt in (1, 2, 5):
+            # retry_delay of 0 to not wait uselessly during the retry process.
+            params = pika.ConnectionParameters(connection_attempts=conn_attempt, retry_delay=0)
+            with self.assertRaises(AMQPConnectionError) as ctx:
+                with mock.patch('socket.socket.connect',
+                                side_effect=socket.timeout) as connect_mock:
+                    pika.BlockingConnection(parameters=params)
+            # as any attempt will timeout (directly),
+            # at the end there must be exactly that count of socket.connect() method calls:
+            self.assertEqual(conn_attempt, connect_mock.call_count)
+            # and each must be with the following arguments (always the same):
+            connect_mock.assert_has_calls(conn_attempt * [mock.call(('127.0.0.1', 5672))])
+            # and the raised error must then looks like:
+            self.assertEqual('Connection to 127.0.0.1:5672 failed: timeout', str(ctx.exception))


gst:
damn, apparently it's not good (enough), now getting some AssertionError when using connection_attemps > 1..

gst:
There it is.

gst:
NB: actually I'm not completly sure this is the best way / place to handle this specific case..

vitaly-krugl:
@gst, please implement tests that validate this scenario and fix.

gst:
Should be good now I think..

gst:
damn, 4 green on the 5 travis configurations.. and the one which fail is python3.4 but I can't reproduce it on my side..

gst:
@vitaly-krugl 
could please restart the travis build on this : https://travis-ci.org/pika/pika/jobs/98017989  ?
want to see if it reproduces itself..

vitaly-krugl:
@gst: sorry, but I don't have admin privileges on this one, so can't restart the build. @gmr is the owner of this repo.

I noticed that there was an additional failure the 3.4 and the 3.5 builds that was logged, but suppressed by the io-loop, so that test was a false negative. The test timeout timer fires from the context of the io-loop, and a lot of the io-loop implementations suppress exceptions in the handlers. This also needs to be fixed to avoid false negatives:

1. in `AsyncTestCase.setUp`: add `self._timed_out = False`  
2. in `AsyncTestCase. on_timeout `: change `self.logger.debug` to `self.logger.error` and add `self._timed_out = True`
3. in `AsyncTestCase.start `: after the line `self.connection.ioloop.start()`, add `self.assertFalse(self._timed_out)`
```
Verify that access denied invokes on open error callback (TornadoConnection) ... ok
Publish a message and consume it (LibevConnection) ... Exception ignored in: <bound method LibevConnection._timer_callback of <pika.adapters.libev_connection.LibevConnection object at 0x7f7994a21dd8>>
  File "/home/travis/build/pika/pika/pika/adapters/libev_connection.py", line 242, in _timer_callback
    callback_method(**kwargs)
  File "/home/travis/build/pika/pika/tests/acceptance/async_test_base.py", line 85, in on_timeout
    raise AssertionError('Test timed out')
AssertionError: Test timed out
ok
```

gst:
damn, worse, and on my side I have all tests passing under python2.7 as well 3.4 and 3.5 :s

tests/acceptance/blocking_adapter_test.py
```diff
@@ -2393,5 +2401,23 @@ def test(self):
 ##        self.assertEqual(frame.method.consumer_count, 1)
 
 
+class TestConnectionAttemptsAgainstTimeout(BlockingTestCaseBase):
+
+    def test_connection_attempts_1(self):
+        params = pika.ConnectionParameters(connection_attempts=1)
+        with mock.patch('socket.socket.connect', side_effect=socket.timeout()):
```
vitaly-krugl: @gst, we do unit/mock tests in the `tests/unit/` directory. The acceptance tests are performed with real connections or connection attempts.  We already had a similar acceptance test `TestConnectWithDownedBroker`. I extended it to support 1 and 2 connection attempts: https://github.com/vitaly-krugl/pika/commit/81cbba61f4d8db1fd6c1db1650fbfe3acb0390df

vitaly-krugl:
@gst, sorry about getting you sidetracked into `AsyncTestCase` changes. Perhaps the `AsyncTestCase` changes could be a separate PR, along with a fix for that `Publish a message and consume it (LibevConnection)` failure.

vitaly-krugl:
@gst, also: the `BlockingChannel.basic_nack` failure in https://travis-ci.org/pika/pika/jobs/98017989 appears to be a race condition in the test. I just got a similar failure in `BlockingChannel.basic_reject` in my build https://travis-ci.org/pika/pika/jobs/98058578. When I implemented those tests, I assumed that the passive `queue_declare` calls would flush the data path to the broker and guarantee that the nack or reject would be handled before the passive `queue_declare` response. However, it looks like we need a retry loop (with a timeout) around those checks.

vitaly-krugl:
@gst, I discovered something else: those failures that you're getting in the Travis builds in "Publish a message and consume it (LibevConnection)" from `TestZ_PublishAndConsume` test class appear to be triggered by another test that runs before it. If I `@unittest.skip()` `TestZ_AccessDenied`, then `TestZ_PublishAndConsume` doesn't fail. Perhaps `TestZ_PublishAndConsume` is messing something up, and is the place to debug.

vitaly-krugl:
@gst, I finally cracked the LibevConnection test timeout problem: libev caches the current time at each event. LibevConnection uses a global default I/O loop by default. The cached time is frozen after the last event in the "Verify that access denied invokes on open error callback (LibevConnection" test. There are five more "Verify that access denied invokes on open error callback" tests after it using other ioloops (TornadoConnection, SelectConnection, etc.) at about 3 seconds each (about 15 seconds total). So, the next LibevConnection test "Publish a message and consume it (LibevConnection)" runs about 15 seconds later, which is roughly the value of `AsyncTestCase.TIMEOUT` and sets a test timeout of `AsyncTestCase.TIMEOUT` (15) seconds, but Libev's default ioloop's cached time is about 15 seconds in the past at that moment, so the test timeout timer fires almost immediately when the "Publish a message and consume it (LibevConnection)" test starts running.

In my experiment, I was able to fix this problem by adding a call to `self.ioloop.update()` in the default case within `LibevConnection.__init__` as follows:

```
        if custom_ioloop:
            self.ioloop = custom_ioloop
        else:
            with warnings.catch_warnings():
                warnings.simplefilter("ignore", RuntimeWarning)
                self.ioloop = pyev.default_loop()
                self.ioloop.update() # ZZZ added this to fix the erroneous test timeout
```

gst:
@vitaly-krugl 
sorry didn't see directly your comments yesterday.
Already I could reproduce some of the travis failures (simply by running the test with nosetest instead of pytest..).
Ah, and you have apparently a fix for that .. ? What I'm not sure is if it's related to the specific problem I'm trying to get fixed here. If not related: will you open a separate PR with it ? or else I'm not against doing it..
And, if I get all things correct, about the AsyncTestCase adaptation I made it's better to have it separately as well.. I'm gonna do that no problem :)

gst:
> I'm gonna do that no problem :)

just waiting confirmation..

pika/adapters/blocking_connection.py
```diff
@@ -370,14 +377,12 @@ def _process_io_for_connection_setup(self):  # pylint: disable=C0103
 
         :raises AMQPConnectionError: on connection open error
         """
+        self._check_open_error()
+
         self._flush_output(self._opened_result.is_ready,
```
vitaly-krugl: This method is small and the new `_check_open_error` method is only used from within this method. The code would be more readable to leave the old logic in place and simply make the call to `_flush_output` conditional on `self._open_error_result.ready`, like this:

```
if not self._open_error_result.ready:
    self._flush_output(self._opened_result.is_ready, ...

if self._open_error_result.ready:
    exception_or_message = self._open_error_result.value.error
    ...
```

vitaly-krugl:
Hi @gst, here is what I think is needed:

1. Move bd8a6ee8612d661be17e595af382b1b7a2cefbcb "Adapt AsyncTestCase for previous fix" to a separate PR, combined with the `self.ioloop.update()` fix described in my earlier comment. Otherwise, the travis build will keep failing for this PR.
2. See comment https://github.com/pika/pika/pull/656/files#r48185505 about the change in `_process_io_for_connection_setup`.
3. Move the mock tests introduced by a1b17d36d83fa1e931b3459d65423edaee17ac20 "add tests for connection_attempts scenarios" to unit tests https://github.com/pika/pika/blob/master/tests/unit/blocking_connection_tests.py
4. I think that it might be better to add the new acceptance tests from https://github.com/vitaly-krugl/pika/blob/81cbba61f4d8db1fd6c1db1650fbfe3acb0390df/tests/acceptance/blocking_adapter_test.py#L225-L259 as part of this PR, since it's tightly coupled to your change in this PR. Let me know if you prefer that I submit it via a separate PR.
5. I filed the issue #677 "Transient failures in acceptance test blocking_adapter_test.py" that can be implemented by someone later.

gst:
For point 4, I'm not sure what to do finally. I do think that my test case is better because I also assert that the socket.connect() is actually called the number of times desired ("connect_attempts" times) (and normally with the expected arguments though that's less "important" in this situation) + also, and that's the more important, that the raised exception is actually the timeout one..

vitaly-krugl:
@gst, regarding point 4: I think that we need both tests. The unit test cases that you implemented test that certain methods get called the correct number of times. The acceptance test cases help confirm that the BlockingConnection adapter is able to make a real live connection to the broker and does not get confused under the circumstances. 

Incidentally, I don't recall seeing a test like that in `async_adapter_tests.py`.

gst:
@vitaly-krugl  I'm not sure what must/should be done finally here.. :?

vitaly-krugl:
@gst, regarding
> I'm not sure what must/should be done finally here.. :? 

Are you referring just to point 4 or more broadly?

gst:
both I guess ;)

vitaly-krugl:
@gst, I described that in https://github.com/pika/pika/pull/656#issuecomment-166426610 and the clarified point 4.

pika/adapters/blocking_connection.py
```diff
@@ -363,6 +363,13 @@ def _acquire_event_dispatch(self):
             # __exit__ part
             self._event_dispatch_suspend_depth -= 1
 
+    def _check_open_error(self):
```
vitaly-krugl: @gst, this is not used now, please remove `_check_open_error`.

vitaly-krugl:
LGTM after the unused method `_check_open_error` is removed.

pika/adapters/blocking_connection.py
gst: I forgot it, will remove it asap

gst:
done and rebased.

vitaly-krugl:
LGTM

Issue #665, opened 2015-11-24
Pick the max of the heartbeat timeouts as result of negotiation 


vitaly-krugl:
Presently, there is no way for the user of pika to override the broker's default heartbeat timeout with a bigger one. This has become particularly problematic since RabbitMQ reduced its default heartbeat timeout in release 3.5.5 from 580 to 60 seconds.

The official java RabbitMQ client picks the max, and Michael Klishin (Pivotal/RabbitMQ) recommends that other clients follow that approach; see https://groups.google.com/forum/#!topic/rabbitmq-users/FpUdTaIr8CI and https://github.com/rabbitmq/rabbitmq-java-client/blob/30a9603a9fb9aaf02b518670e4b1c19e1ac8953c/src/com/rabbitmq/client/impl/AMQConnection.java#L367-L369

vitaly-krugl:
Fixed via #666 - in master

PR #666, opened 2015-11-24
Pick the max of the heartbeat timeouts as result of negotiation

nickcash:
For issue #665

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,8 @@
+languages:
+  - python
+exclude_paths:
+  - docs/*
+  - tests/*
+  - utils/*
+  - pika/examples/*
+  - pika/spec.py
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -360,7 +360,7 @@ class ConnectionParameters(Parameters):
         :param int channel_max: Maximum number of channels to allow
         :param int frame_max: The maximum byte size for an AMQP frame
         :param int heartbeat_interval: How often to send heartbeats.
-                                  Min between this value and server's proposal
+                                  Max between this value and server's proposal
                                   will be used. Use 0 to deactivate heartbeats
                                   and None to accept server's proposal.
         :param bool ssl: Enable SSL
@@ -1338,8 +1338,8 @@ class Connection(object):
         if self.params.heartbeat is None:
             self.params.heartbeat = method_frame.method.heartbeat
         elif self.params.heartbeat != 0:
-            self.params.heartbeat = self._combine(self.params.heartbeat,
-                                                  method_frame.method.heartbeat)
+            self.params.heartbeat = max(self.params.heartbeat,
+                                        method_frame.method.heartbeat)
 
         # Calculate the maximum pieces for body frames
         self._body_max_length = self._get_body_frame_max_length()


vitaly-krugl:
Thanks, @nickcash for the quick follow up. Would you mind adding a few simple unit tests in tests/unit/connection_tests.py for the change? But please hold off on that until @gmr agrees to this change in behavior (see my note below). If he does, then the following test cases would be good to cover:
1. client passes 0, so `_send_connection_tune_ok` should send 0 heartbeat timeout to broker in `spec.Connection.TuneOk`
2. Broker passes bigger value, so  `_send_connection_tune_ok` should send the server's heartbeat timeout to broker in `spec.Connection.TuneOk`
3. User passes bigger value, so  `_send_connection_tune_ok` should send the user's heartbeat timeout to broker in `spec.Connection.TuneOk`
4. Same values from both broker and user


@gmr, could you please take a look? This change was recommended by Michael Klishin in https://groups.google.com/forum/#!topic/rabbitmq-users/FpUdTaIr8CI.

nickcash:
After looking more closely at the Java client's code that [MK linked to](https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/com/rabbitmq/client/impl/AMQConnection.java#L357), I'm wondering if this change goes far enough. In addition to the heartbeat, the Java client sets the channel max and frame max as the max of the client request and server response, whereas pika is using the min for those as well.

So maybe a better change would be to switch _combine() to use max instead? If matching the Java behavior is desired, that is.

vitaly-krugl:
@nickcash, I was wondering that myself. I would like to hear @gmr's opinion on it.

vitaly-krugl:
Gavin, what's your opinion regarding @nickcash comment concerning the channel max, etc.?

vitaly-krugl:
@nickcash, regarding
> So maybe a better change would be to switch _combine() to use max instead? If matching the Java behavior is desired, that is.

Max is not appropriate for resource limits, such as frame size, max channels, etc. See the following excerpt from AMQP 0-9-1 spec and the last line, in particular.

Per https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf:
> 2.3.3 Protocol Negotiation
The AMQP client and server negotiate the protocol. This means that when the client connects, the server
proposes certain options that the client can accept, or modify. When both peers agree on the outcome, the
connection goes ahead. Negotiation is a useful technique because it lets us assert assumptions and
preconditions.
In AMQP, we negotiate a number of specific aspects of the protocol:
 The actual protocol and version. The server MAY host multiple protocols on the same port.
 Encryption arguments and the authentication of both parties. This is part of the functional layer,
explained previously.
 Maximum frame size, number of channels, and other operational limits.
Agreed limits MAY enable both parties to pre-allocate key buffers, avoiding deadlocks. Every incoming
frame either obeys the agreed limits, and so is "safe", or exceeds them, in which case the other party IS
faulty and MUST be disconnected. This is very much in keeping with the “it either works properly or it
doesn't work at all” philosophy of AMQP.
Both peers negotiate the limits to the lowest agreed value as follows:
 The server MUST tell the client what limits it proposes.
 **The client responds and MAY reduce those limits for its connection.**

nam4dev:
:+1:

gmr:
I'm good to merge this if the test coverage is improved to cover the change.

vitaly-krugl:
@nickcash: +1 for this change; could you please add test coverage to enable merging? Thank you.

nickcash:
I fully intended to write tests, but between holidays and workload I haven't had much time. I'll get to them ASAP!

vitaly-krugl:
Thant's great, @nickcash, thank you.

vitaly-krugl:
Hi @nickcash, I implemented the new heartbeat tuning test logic that corresponds to your changes in this PR. This is the commit that contains the test change: https://github.com/vitaly-krugl/pika/commit/6ef71f2d12d19128075624587d5fc21f6b4b8644. I wasn't able to submit a PR into your fork, but you can probably pull my commit into your branch.

If you like my test logic commit, please pull it into your pull request so that we can merge your PR.

Is there a github workflow that permits me to submit a PR into your fork?

Thanks,
Vitaly

vitaly-krugl:
@nickcash: UPDATE - I created a new commit https://github.com/vitaly-krugl/pika/commit/41b4f3230852368c4a562ac2e64984db5c589a53 that treats 0 from client and server the same and documented the reasoning behind it and how it applies to heartbeat negotiation per AMQP 0.9.1 specification. Please consider merging it. Thx.

vitaly-krugl:
@nickcash, please see https://github.com/nickcash/pika/pull/2 for tests and other changes related to this PR instead of the previously-mentioned commit.

nickcash:
@vitaly-krugl I've merged your pull request to my fork. Thanks for doing all the actual work while I was being lazy :smile:

vitaly-krugl:
Thanks, and I appreciate your initiative in laying the groundwork.

vitaly-krugl:
@nickcash, please take a look at the new PR https://github.com/nickcash/pika/pull/3. It should resolve the merge conflict against master.

nickcash:
@vitaly-krugl, looks good to me. Thanks again!

vitaly-krugl:
Thanks @nickcash, merging ...

avihoo:
Is it possible that even though the PR has merged it can't be found in current version `0.10.0`?

vitaly-krugl:
@avihoo - it's possible, if the PR was merged after release 0.10.0.

avihoo:
So, any estimation on a new version to include that fix? Really needs it.

avihoo:
I'm using this in a production environment so `pip` install directly from `master` is probably not the greatest idea for now. Is there anyone that might know when the next version is to be released?

anthonyserious:
I'm on 0.11.2 (via pip), but I'm still seeing the server default chosen over mine.  Has there been any progress getting this into a release?  Or am I possibly facing a different problem?

```    self.conn = pika.BlockingConnection(pika.ConnectionParameters(
        host=host,
        port=port,
        blocked_connection_timeout=360,
        heartbeat=600
    ))
```

Yet on the server, `missed heartbeats from client, timeout: 60s`.  Also passing `heartbeat=0` has no effect - I would expect heartbeating to be disabled, but I still get the server-logged 60s timeout message plus a connectionClosed exception:

```  File "/usr/local/lib/python3.6/site-packages/pika/adapters/blocking_connection.py", line 2077, in basic_publish
    mandatory, immediate)
  File "/usr/local/lib/python3.6/site-packages/pika/adapters/blocking_connection.py", line 2164, in publish
    self._flush_output()
  File "/usr/local/lib/python3.6/site-packages/pika/adapters/blocking_connection.py", line 1250, in _flush_output
    *waiters)
  File "/usr/local/lib/python3.6/site-packages/pika/adapters/blocking_connection.py", line 474, in _flush_output
    result.reason_text)
pika.exceptions.ConnectionClosed: (-1, 'EOF')
```

lukebakken:
@anthonyserious this is resolved in #966 and will be released in version `1.0.0`

anthonyserious:
@lukebakken thanks for the quick response.  I can override the server-side heartbeat for now as a workaround (dedicated instance for 1 or 2 queues).

Issue #649, opened 2015-09-16
ProbableAuthenticationError not turned into callback in TornadoConnection


gmr:
```python
ERROR      2015-09-16 14:45:22 2184   async-1         tornado.application  handle_callback_exception: Exception in callback None
Traceback (most recent call last):
  File "/Users/gavinr/Source/rejected/env/lib/python2.7/site-packages/tornado/ioloop.py", line 866, in start
    handler_func(fd_obj, events)
  File "/Users/gavinr/Source/rejected/env/lib/python2.7/site-packages/tornado/stack_context.py", line 275, in null_wrapper
    return fn(*args, **kwargs)
  File "/Users/gavinr/Source/rejected/env/lib/python2.7/site-packages/pika/adapters/base_connection.py", line 364, in _handle_events
    self._handle_read()
  File "/Users/gavinr/Source/rejected/env/lib/python2.7/site-packages/pika/adapters/base_connection.py", line 412, in _handle_read
    return self._handle_disconnect()
  File "/Users/gavinr/Source/rejected/env/lib/python2.7/site-packages/pika/adapters/base_connection.py", line 288, in _handle_disconnect
    self._adapter_disconnect()
  File "/Users/gavinr/Source/rejected/env/lib/python2.7/site-packages/pika/adapters/tornado_connection.py", line 73, in _adapter_disconnect
    super(TornadoConnection, self)._adapter_disconnect()
  File "/Users/gavinr/Source/rejected/env/lib/python2.7/site-packages/pika/adapters/base_connection.py", line 154, in _adapter_disconnect
    self._check_state_on_disconnect()
  File "/Users/gavinr/Source/rejected/env/lib/python2.7/site-packages/pika/adapters/base_connection.py", line 173, in _check_state_on_disconnect
    raise exceptions.ProbableAuthenticationError
ProbableAuthenticationError
```

sjlongland:
I've occasionally butted heads with this exception myself.  Is there a workaround?  (Sometimes it isn't an authentication error, just the AMQP broker being slow.)

sjlongland:
@dave-shawley: That indeed does cure this problem.  Thank-you.

vitaly-krugl:
fixed

vitaly-krugl:
Fixed in Pika master, closing

PR #672, opened 2015-12-03
Call on open error callback for probable auth/access failures

dave-shawley:
This PR invokes the `on_open_error_callback` when the broker connection is lost between the time that the socket connection is established and the OpenOK comes back.  The current implementation does not invoke any of the connection related callbacks.  I ran into this using the Tornado async adapter which exhibits itself as creating the `TornadoConnection` object and never receiving a callback until my application was closing.

### DIFF
--- a/docs/version_history.rst
+++ b/docs/version_history.rst
@@ -1,6 +1,13 @@
 Version History
 ===============
 
+Next Release
+------------
+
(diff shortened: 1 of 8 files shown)


dave-shawley:
I think that this might fix #649 as well

gmr:
Thanks :)

PR #673, opened 2015-12-08
Connection._next_channel_number optimization

eugals:
When we have many (hundreds) channels opened within a connection, the current `Connection._next_channel_number` implementation may become a bottleneck sometimes.
```Python
        ckeys = set(self._channels.keys())
        if not ckeys:
            return 1
        return [x + 1 for x in sorted(ckeys) if x + 1 not in ckeys][0]
```
I see 3 problems with this code.
1. This `ckeys = set(self._channels.keys())` call is effectively useless here. The rest of the code would work just as fine if there was simply `ckeys = self._channels`
2. Not the best idea to use a `[list-comprehension][0]` when we only need to find a first item of the resulting list
3. The `x + 1` part is all right, but why don't we allow to reuse `x - 1` numbers too?

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -1207,13 +1207,10 @@ class Connection(object):
         if len(self._channels) >= limit:
             raise exceptions.NoFreeChannels()
 
-        if self._channels:
-            n = min(self._channels)
-            if n <= 1:
-                for n in xrange(n + 1, n + len(self._channels) + 1):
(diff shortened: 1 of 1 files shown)


gmr:
Thanks

vitaly-krugl:
@gmr: could you please revert this PR; it breaks all the Python 3 builds

PR #675, opened 2015-12-19
Fix issue 663: don't block channel when making RPC call for nowait=True

### LINKED ISSUE #663: Regression: Cannot receive long running messages when in use in tornado app
Hi there,

we just noticed a regression from pika 0.9.14 to 0.10.0. We haven't been able to track down the problem further than this (sorry) but will instead stop for now by pinning the old version.

The problem seems to be that pika (used via tornado-celery) is not able to receive answer messages from rabbitmq for longish running (>4-5 seconds) requests to a background worker.

The answer message instead seems to time out after the message timeout period in rabbitmq after which it is lost.

Do you need more information to fix this?

vitaly-krugl:
Fixes #663
Fix Channel._rpc to place channel in blocking state only if acceptable_replies arg is not empty to ensure that _on_synchronous_complete will be called to process subsequently blocked requests.

### DIFF
--- a/pika/adapters/libev_connection.py
+++ b/pika/adapters/libev_connection.py
@@ -127,7 +127,7 @@ class LibevConnection(BaseConnection):
             if self._on_signal_callback and not global_sigterm_watcher:
                 global_sigterm_watcher = \
                     self.ioloop.signal(signal.SIGTERM,
-                                                                                  self._handle_sigterm)
+                                       self._handle_sigterm)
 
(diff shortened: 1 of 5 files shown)


vitaly-krugl:
CC @gmr: Hi Gavin, please review this fix.

vitaly-krugl:
@gmr, the failed test `BlockingChannel.basic_nack single message` is unrelated to the changes in this PR; it's a known issue described in #677. 

Please re-start the build. Thanks!

PR #676, opened 2015-12-21
Both int and long are valid types for delivery_tag

quantum5:
When calling `basic_cancel`, `pika` will attempt to use call `basic_reject` on all pending messages. However, sometimes, `delivery_tag` is a `long`, and it fails the `if not isinstance(delivery_tag, int)` test, resulting in an exception.

On the other hand, `basic_nack` has no such problems.

```pytb
  File ".../site-packages/pika/adapters/blocking_connection.py", line 1598, in basic_cancel
    requeue=True)
  File ".../site-packages/pika/channel.py", line 394, in basic_reject
    raise TypeError('delivery_tag must be an integer')
TypeError: delivery_tag must be an integer
```

### DIFF
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -146,7 +146,7 @@ class Channel(object):
         confirm mode. The acknowledgement can be for a single message or a
         set of messages up to and including a specific message.
 
-        :param int delivery_tag: The server-assigned delivery tag
+        :param integer delivery_tag: int/long The server-assigned delivery tag
         :param bool multiple: If set to True, the delivery tag is treated as
                               "up to and including", so that multiple messages
                               can be acknowledged with a single method. If set
@@ -286,7 +286,7 @@ class Channel(object):
         It can be used to interrupt and cancel large incoming messages, or
         return untreatable messages to their original queue.
 
-        :param int delivery-tag: The server-assigned delivery tag
+        :param integer delivery-tag: int/long The server-assigned delivery tag
         :param bool multiple: If set to True, the delivery tag is treated as
                               "up to and including", so that multiple messages
                               can be acknowledged with a single method. If set
@@ -381,7 +381,7 @@ class Channel(object):
         message. It can be used to interrupt and cancel large incoming messages,
         or return untreatable messages to their original queue.
 
-        :param int delivery-tag: The server-assigned delivery tag
+        :param integer delivery-tag: int/long The server-assigned delivery tag
         :param bool requeue: If requeue is true, the server will attempt to
                              requeue the message. If requeue is false or the
                              requeue attempt fails the messages are discarded or
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -436,7 +436,10 @@ class Channel(object):
         For more information see:
             http://www.rabbitmq.com/extensions.html#confirms
 
-        :param method callback: The callback for delivery confirmations
+        :param method callback: The callback for delivery confirmations that has
+          the following signature: callback(pika.frame.Method), where
+          method_frame contains either method `spec.Basic.Ack` or
+          `spec.Basic.Nack`
         :param bool nowait: Do not send a reply frame (Confirm.SelectOk)
 
         """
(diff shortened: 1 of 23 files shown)


vitaly-krugl:
@quantum5, the tests are failing for python 3.x, because it only defines a single integer type: `int`. You will need to solve this via the https://github.com/pika/pika/blob/master/pika/compat.py module, probably by defining `is_interger(value)`: in the `if not PY2:` block, it should check only for `int`, and in the other block (python 2.x) it should check for `(int, long)`.

vitaly-krugl:
@quantum5, that's a good change; https://www.rabbitmq.com/amqp-0-9-1-reference.html#domain.delivery-tag defines it as `longlong`, which is defined as `64-bit integer`. The code in pika's `spec.py` uses `>Q` to pack and unpack the tag: `pieces.append(struct.pack('>Q', self.delivery_tag))`, with `Q` defined as "unsigned long long" in https://docs.python.org/2/library/struct.html. Hopefully, for this purpose, the `Q` is compatible with AMQP's `64-bit integer`.

vitaly-krugl:
@gmr, the Travis status build of this PR appears to be missing.

pika/channel.py
```diff
@@ -390,7 +390,7 @@ def basic_reject(self, delivery_tag, requeue=True):
         """
         if not self.is_open:
             raise exceptions.ChannelClosed()
-        if not isinstance(delivery_tag, int):
+        if not is_integer(delivery_tag):
```
vitaly-krugl: @quantum5, I found a builtin module that may provide the same functionality in a portable way without necessitating any `pika.compat` logic:

```
import numbers

if not isinstance(delivery_tag, numbers.Intergral):
    . . .
```

quantum5: I thought about using it, but decided against it because:

```py
>>> import numbers
>>> numbers.Integral.register(type(None))
>>> isinstance(None, numbers.Integral)
True
```

vitaly-krugl: @quantum5, I get False on python 2.7.10 and 3.5

```
2.7.10 (default, Jul 14 2015, 19:46:27) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)]
Python Type "help", "copyright", "credits" or "license" for more information.
import numbers
numbers.Integral
<class 'numbers.Integral'>
isinstance(None, numbers.Integral)
False
```

quantum5: My point is that people can register any type to be an `Integral`, and so I decided to not risk it. `NoneType` was just an extreme example.

vitaly-krugl: Thx

vitaly-krugl:
Please add tests to the PR. Thx

pika/channel.py
vitaly-krugl: Interestingly, this is somewhat messed-up, too:

```
isinstance(True, int)
True
```

quantum5: `True` and `False` are basically `1` and `0`. They even work in arithmetic:

```py
>>> True + True
2
>>> True + False
1
>>> False + False
0
>>> 5 + True
6
>>> False - 100
-100
```

vitaly-krugl:
@quantum5, I implemented tests for your PR; please consider merging the tests from https://github.com/vitaly-krugl/pika/commit/25d917b8a0dfa322aa26debb4dc72e226ef0f7b1 into your PR.

I was trying to create a pull request into your pika fork, but couldn't get DMOJ/pika for the base fork.

vitaly-krugl:
@quantum5, I saw the breakage on py3 from my tests. I will submit a PR into your fork when I have a confirmed fix. Thx.

vitaly-krugl:
PR https://github.com/DMOJ/pika/pull/1 should fix the build breakage

vitaly-krugl:
@quantum5 Something went wrong with the merge - the PR picked up a lot of changes: 23 files and 35 commits. I created a clean test PR that has both your and my commits combined in it and it passes all checks: see PR #705

vitaly-krugl:
@quantum5, I am closing this PR via PR #705 that I created from yours and which is rebased against current Pika master branch. I credited you here: https://github.com/pika/pika/pull/705/files#diff-cfc895343b31cee85bb3efb82bc19fb2R20

PR #660, opened 2015-10-26
Fix: Improper KeyError when channel created on a BlockingConnection whose server was shutdown

gst:
More specifically:

    return self._channels[channel_number]

will raise KeyError in such case because the channel will have already been removed from the connection _channels dict.

Using this fix you'll more correctly get ConnectionClosed exception.

PR #679, opened 2015-12-28
Fixes #659 - prevent execution of unintended code paths after connection failure

### LINKED ISSUE #659: BlockingConnection + create new channel raise KeyError if server stopped
pika 0.10 : 

if the underlying connection has been closed by rabbit.

Steps to reproduce : 

1) create a new BlockingConnection, make sure it connects to the server : 
`conn = pika.BlockingConnection()`
2) Stop the server.
3) using the connection object from 1 try to create a new channel:
    `conn.channel()`

this actually raises a KeyError..

Here is the stack trace showing the actual problem:

image

We end up, from `channel.open()` (from within channel() method from Connection (connection.py)), in `_init_connection_state()` where the connection _channels attribute is reset with a new dict.

Then, when the call chain return from channel.open(), the next line will raise KeyError:

`return self._channels[channel_number]`

because the _channels attribute was reset/rebound to a new fresh empty dict in `_init_connection_state()`.

vitaly-krugl:
Fixes #659

Note that #659 cites a bug in `BlockingConnection`. This PR fixes the core bug that affects **all adapters**.

Methods in pika core and adapters do not deal gracefully with connection errors occurring in their context; e.g., `Connection.channel` (pika issue #659), `Connection._on_connection_tune` (if connection loss is detected in `_send_connection_tune_ok`, before `_send_connection_open` is called), etc.

Undoes part of a breaking change from PR #578 to fix issue #659 and others like it, because pika code was not designed to perform I/O from any methods other than ioloop handler callbacks.

Support SO_LINGER in forward_server test utility that helps verify the fix.

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -276,10 +276,16 @@ class BaseConnection(connection.Connection):
         return None
 
     def _flush_outbound(self):
-        """write early, if the socket will take the data why not get it out
-        there asap.
+        """Have the state manager schedule the necessary I/O.
         """
-        self._handle_write()
+        # NOTE: We don't call _handle_write() from this context, because pika
+        # code was not designed to be writing to (or reading from) the socket
+        # from any methods, except from ioloop handler callbacks. Many methods
+        # in pika core and adapters do not deal gracefully with connection
+        # errors occurring in their context; e.g., Connection.channel (pika
+        # issue #659), Connection._on_connection_tune (if connection loss is
+        # detected in _send_connection_tune_ok, before _send_connection_open is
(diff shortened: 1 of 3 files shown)


vitaly-krugl:
@gst, this is a more general solution that is intended to replace PR #660. Please review.

vitaly-krugl:
@wjps: Hi Will, it turns out that the "tornado write starvation" PR #578 broke a number of code paths in pika. This PR undoes the breaking change in `BaseConnection._flush_outbound()`. The comment https://github.com/pika/pika/pull/660#issuecomment-167213898 has a more detailed analysis of the problem.

PR #680, opened 2015-12-28
Fix leakage of ioloop pollers in SelectConnection and BlockingConnection

vitaly-krugl:


pika/adapters/select_connection.py
```diff
@@ -155,32 +157,34 @@ def _get_poller(self):
         return poller
 
 
-class SelectPoller(object):
-    """Default behavior is to use Select since it's the widest supported and has
-    all of the methods we need for child classes as well. One should only need
-    to override the update_handler and start methods for additional types.
+class _PollerBase(object):  # pylint: disable=R0902
+    """Base class for select-based IOLoop implementations"""
+    __metaclass__ = abc.ABCMeta
```
gst: this (_metaclass_ definition/assignment in class body) is py2 only

vitaly-krugl: @gst, thank you for pointing this out. I had no idea about this change in Python v3.x.

vitaly-krugl: @gst, I fixed this using `exec` on python 3. If Gavin allows dependency on six, then I will do it the six way.

gmr: I'd prefer not to use six, but rather to start focusing on Python 3 first with backwards compatibility patched in for two.

Is the desire to use six to just bypass having to deal with the differences?

vitaly-krugl: @gmr, in order to make use of `abc.ABCMeta` in Python 3, we need to use the new Python 3 syntax that's not supported in Python 2. That new syntax (`MyClass(metaclass=abc.ABCMeta)`) prevents the module from loading when running in Python 2. To make it work without `six`, I had to wrap the Python 3 variant of the code in the `exec()` function so it won't break loads under Python 2, which isn't ideal (https://github.com/vitaly-krugl/pika/blob/897624da1fa6c9aeeb854aca3a9abf5dd4dcf148/pika/adapters/select_connection.py#L170). I am not sure how else to overcome this compatibility issue.

pika/adapters/select_connection.py
```diff
@@ -155,32 +157,46 @@ def _get_poller(self):
...
+# between python 2 and 3 (metaclass syntax changed in Python 3). Ideally, would
+# use `@six.add_metaclass` or `six.with_metaclass`, but pika traditionally has
+# resisted external dependencies in its core code.
+if pika.compat.PY2:
+    class _AbstractBase(object):
+        __metaclass__ = abc.ABCMeta
+else:
+    # NOTE: Wrapping in exec, because `metaclass=abc.ABCMeta` fails to load
+    # on python 2.
+    exec('class _AbstractBase(metaclass=abc.ABCMeta): pass')
```
gst: strange, *as far as I know* it shouldn't fail to load without the exec..

vitaly-krugl:
It looks like Travis CI is still having issues: https://www.traviscistatus.com/. All of the failures are in the initialization: `The command "sudo apt-get update -qq" failed and exited with 100 during .`

vitaly-krugl:
Resolved merge conflict and rebased against new master

vitaly-krugl:
Resolved merge conflict

PR #678, opened 2015-12-22
Fix/Enh: correctly handle ioloop timeouts with AsyncTestCase

gst:

PR #683, opened 2016-01-01
Fix segfault in libevent connection that was introduced by PR 675

vitaly-krugl:
@gmr: this should fix the segfault from my earlier PR #675. This small change makes sure that timers are not deleted more than once by `async_test_base.AsyncTestCase` and also makes the logic in `LibevConnection` that removes timers more robust.

This PR should be merged ahead of other open PRs so that its code may be merged into the remaining PRs to permit tests to pass.

Root cause: #675 changed a test in a way that exposed bugs in `async_test_base. AsyncTestCase ` and `libev_connection.py`. Strangely, the #675 build got "lucky" and didn't trigger that failure in any of the platform builds (as have one or two other PRs since then).

### DIFF
--- a/pika/adapters/libev_connection.py
+++ b/pika/adapters/libev_connection.py
@@ -99,6 +99,7 @@ class LibevConnection(BaseConnection):
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", RuntimeWarning)
                 self.ioloop = pyev.default_loop()
+                self.ioloop.update()
 
         self.async = None
         self._on_signal_callback = on_signal_callback
@@ -139,8 +140,11 @@ class LibevConnection(BaseConnection):
                                    self._PIKA_TO_LIBEV_ARRAY[self.event_state],
                                    self._handle_events)
 
+            # NOTE: if someone knows why this async is needed here, please add
+            # a comment in the code that explains it.
             self.async = pyev.Async(self.ioloop, self._noop_callable)
             self.async.start()
+
             if self._on_signal_callback:
                 global_sigterm_watcher.start()
             if self._on_signal_callback:
@@ -237,12 +241,12 @@ class LibevConnection(BaseConnection):
             (callback_method, callback_timeout,
              kwargs) = self._active_timers[timer]
 
+            self.remove_timeout(timer)
+
             if callback_timeout:
                 callback_method(timeout=timer, **kwargs)
             else:
                 callback_method(**kwargs)
-
-            self.remove_timeout(timer)
         else:
             LOGGER.warning('Timer callback_method not found')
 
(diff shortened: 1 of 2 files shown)


pika/adapters/libev_connection.py
```diff
@@ -99,6 +99,7 @@ def __init__(self,
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", RuntimeWarning)
                 self.ioloop = pyev.default_loop()
+                self.ioloop.update()
```
vitaly-krugl: @gst, this line will likely have a minor merge conflict with the same line in your PR #678. I needed this here in order to get a clean build.

pika/adapters/libev_connection.py
```diff
@@ -237,12 +241,12 @@ def _timer_callback(self, timer, libev_events):
             (callback_method, callback_timeout,
              kwargs) = self._active_timers[timer]
 
+            self.remove_timeout(timer)
```
vitaly-krugl: Moved `remove_timeout` ahead of the callback to make sure it gets cleaned up, even if callback raises; eliminates need for `try/finally`

pika/adapters/libev_connection.py
```diff
@@ -284,9 +288,13 @@ def remove_timeout(self, timer):
 
         """
         LOGGER.debug('stop')
-        self._active_timers.pop(timer, None)
-        timer.stop()
-        self._stopped_timers.append(timer)
```
vitaly-krugl: this old code could lead to the same timer ending up multiple times on `self._stopped_timers` deque if user removed it multiple times or if user removed it after the timer callback was called (which also removes it)

tests/acceptance/async_test_base.py
```diff
@@ -91,11 +92,10 @@ def on_open_error(self, connection, error):
 
     def on_timeout(self):
         """called when stuck waiting for connection to close"""
-        self.logger.info('on_timeout at %s', datetime.utcnow())
-        # force the ioloop to stop
-        self.logger.debug('on_timeout called')
-        self.connection.ioloop.stop()
-        raise AssertionError('Test timed out')
```
vitaly-krugl: Assertion here doesn't help, since ioloops in some adapters, such as `LibevConnection`, suppress the exception. PR #678 will handle test timeout in a robust way, once it's merged.

tests/acceptance/async_test_base.py
```diff
@@ -91,11 +92,10 @@ def on_open_error(self, connection, error):
...
+        self.logger.info('on_timeout called at %s', datetime.utcnow())
+        # Initiate cleanup
+        self.timeout = None  # the dispatcher should have removed it
+        self.stop()
```
vitaly-krugl: @gst, please be careful to preserve these changes in `on_timeout`, too, when merging #678 that modifies the same method. I asked @gmr to merge this PR #683 ahead of the other PRs so that the other PRs may be re-merged with master after #683 to verify clean builds in all of them.

gst: heuh yes.. I had done this #678 totally based on your input if I remember correctly ;)

vitaly-krugl: @gst: I meant that if github forces you to merge, we will need **both**: your changes from #678 and my changes from this PR.

PR #685, opened 2016-01-04
Handle Connection.Close from broker at any time, including during connection setup, per RabbitMQ 3.6.0

### LINKED ISSUE #534: SelectConnection passes reason_code=0 to on_close_callback upon failed socket connection
SelectConnection passes `reason_code=0` to `on_close_callback` (that was registered via SelectConnection constructor) upon failed socket connection. This creates a problem for discerning between a user-initiated close vs. a socket connection failure.

`BaseConnection._handle_disconnect` has no method frame to pass to `self._on_connection_closed`, so `on_close_callback` ends up being emitted with `reason_code=0, reason_text='Not specified'`, which doesn't make sense for an error condition like lost socket connection.

### LINKED ISSUE #631: Socket Error handling problematic
We have a complex application that keeps failing due to unhandled socket errors. I have been able to build a simple case to reproduce the issue. I have used the sample select consumer and select publisher from the readthedocs examples. The ExamplePublisher has been modified to read a queue and publish any message that shows up. I have set the heartbeat interval to 10 seconds and then wait 20+ seconds between publishes. This creates a socket closure.  Pika processes the error and cleans up the connection (destroying my timeout scheduled for connection restart).  

Here are the log messages from the critical part:
INFO       2015-07-23 08:25:35,727 pika_publisher                 enable_delivery_confirmations        258 : Issuing Confirm.Select RPC command
DEBUG      2015-07-23 08:25:35,733 pika.channel                   _rpc                                 1036: Adding in on_synchronous_complete callback
DEBUG      2015-07-23 08:25:35,734 pika.callback                  add                                  152 : Incremented callback reference counter: {'callback': <bound method Channel._on_synchronous_complete of <pika.channel.Channel object at 0x7ff72ac6fad0>>, 'only': None, 'one_shot': True, 'arguments': None, 'calls': 5}
DEBUG      2015-07-23 08:25:35,734 pika.channel                   _rpc                                 1041: Adding passed in callback
DEBUG      2015-07-23 08:25:35,735 pika.callback                  add                                  152 : Incremented callback reference counter: {'callback': <bound method Channel._on_selectok of <pika.channel.Channel object at 0x7ff72ac6fad0>>, 'only': None, 'one_shot': True, 'arguments': None, 'calls': 5}
DEBUG      2015-07-23 08:25:35,737 pika.adapters.select_connection poll                                 477 : Calling <bound method SelectConnection._handle_events of <pika.adapters.select_connection.SelectConnection object at 0x7ff72ac6f5d0>>

=== pika sees socket is closed here when trying to send a frame ===
ERROR      2015-07-23 08:25:35,738 pika.adapters.base_connection  _handle_error                        279 : Socket Error on fd 5: 32

WARNING    2015-07-23 08:25:35,739 pika.adapters.base_connection  _check_state_on_disconnect           145 : Socket closed when connection was open
DEBUG      2015-07-23 08:25:35,741 pika.callback                  add                                  161 : Added: {'callback': <bound method SelectConnection._on_connection_start of <pika.adapters.select_connection.SelectConnection object at 0x7ff72ac6f5d0>>, 'only': None, 'one_shot': True, 'arguments': None, 'calls': 1}
WARNING    2015-07-23 08:25:35,742 pika.connection                _on_disconnect                       1284: Disconnected from RabbitMQ at 15.244.209.6:8080 (0): Not specified
DEBUG      2015-07-23 08:25:35,743 pika.callback                  process                              217 : Processing 0:_on_connection_closed
DEBUG      2015-07-23 08:25:35,744 pika.callback                  process                              231 : Calling <bound method ExamplePublisher.on_connection_closed of <ExamplePublisher(test_publisher, started 140699551250176)>> for "0:_on_connection_closed"
WARNING    2015-07-23 08:25:35,757 pika_publisher                 on_connection_closed                 104 : Connection closed, reopening in 5 seconds: (0) Not specified
ERROR      2015-07-23 08:25:35,758 pika.adapters.base_connection  _handle_events                       305 : Error event 29, None
CRITICAL   2015-07-23 08:25:35,758 pika.adapters.base_connection  _handle_error                        255 : Tried to handle an error where no error existed

There are two issues here:
1. Pika does not cleanly handle the socket issues or provide a simple mechanism to recover from them.
2. Pika fails to carry the necessary metadata to handle the error in the _handle_events/_handle_error methods to correctly reflect what error actually occurred.

At this point we are having to wrap large amounts of error handling around pika to handle these conditions that should be invisible to us.

Suggested approach to a fix:
Anytime the socket gets closes for any reason (not just a few), have a parameter that says "attempt to autostart the socket" and a registed callback for the error code to get a boolean from the publisher indicating whether or not to try to recover and continue.

If I get some more time, I will try to suggest a patch/fix that has been tested.

### LINKED ISSUE #634: _adapter_disconnect is not called properly and the socket remain enabled for select adapaters
I came across with this bug when I saw duplicated connections, or more, in my rabbitmq sever using the `rabbitmqctl list_connections` command where these connections belong to the same Pika AMQP process.

After spend a lot of time guessing which could be the origin of the bug, at last I reproduced it and fixed calling by hand the `_adapter_disconnect` connection method 

```python
    def _on_connection_closed(self, connection_unused, reply_code, reply_text):
        logging.warning('{0} Server closed connection, reopening: ({1}) {2}'.format(self.host,
                                                                                    reply_code,
                                                                                    reply_text))

        # PIKA doesn't call the remove handler of current fd belonging to the socket when
        # there is a communication issue between the server and client, therefore the socket
        # remains open and future heartbeats sent to outdated connection can provoke duplicate
        # duplicated conections scenarios. We call it by hand to avoid that
        connection_unused._adapter_disconnect()

        # after here some strategy to reconnect the client
```

The connection is configured to use the heartbeat strategy to figure out when there are communications issues, to allow to the client runs a strategy to connect again after N seconds. 

```python
    @gen.coroutine
    def _connect(self):
        coroutiner = CoroutineMaker({'TornadoConnection': 'on_open_callback'})
        try:
            credentials = pika.PlainCredentials(self.username, self.password)
            connection_parameters = pika.ConnectionParameters(
                host=self.host,
                credentials=credentials,
                connection_attempts=1,#  reconnection strategy is made by our own
                heartbeat_interval=60)
            self._connection = yield coroutiner.convert(pika.adapters.TornadoConnection)(
                connection_parameters,
                on_open_error_callback=self._on_connection_error)

```

The issue comes when the communications between the server and the client is stalled because of network issues. When the heartbeat module tries to close the connection [1] it calls first of all the method to close the connection and after it the `on_disconnect` method that will fire our callback.

The `connection.close` method tries to close the connection using a set of good practices, such as close the channels, send the properly Close frame, etc. And at last delegating to the callback , ,previously registered to attend the response of the Close frame, some other stuff like call the `_adapter_disconnect` method.

As we saw before the scenario can not grantee that the Close frame is delivered and the most important that the response of this frame is always received.

In that situation the map of the file descriptors that handles the Tornado adapter has still, because the unregistered was called never, the file descriptor belonging to one connection that has been apparently closed.

Because of that, if the network issues vanish the Rabbitmq sever, or SO buffers, can still send data to the Pika AMQP process. By default Pika will show the next error because of this data:

```
2015-08-07 16:39:19,952 - pika.adapters.base_connection - ERROR - Read empty data, calling disconnect
2015-08-07 16:39:19,952 - pika.adapters.base_connection - WARNING - Connection is closed but not stopping IOLoop
2015-08-07 16:39:19,953 - pika.connection - WARNING - Disconnected from RabbitMQ at manager.miimetiq.m2mcloudfactory.com:5672 (0): Not specified
```

And pika will call again the `_on_connection_closed` when in fact the connection is open and running.

[1] https://github.com/pika/pika/blob/0.9.14/pika/heartbeat.py#L118

### LINKED ISSUE #648: No any error information in case of unexpected interrupt connection
Hi guys,

I do not have any information about any errors in pika if connection closed unexpectedly with error (I mean network connection). Connection closes like we do not have any errors:

1) connect to rabbit mq (SelectConnection)
2) unplug Ethernet

Result of pika.SelectConnection::on_close_callback :

    reply_code = 0
    reply_text = 'Not specified'

Like ALL IS OK. May be I did not noticed but I did not find any information in documetation how to detect error information on case of unexpected close (e.g. callbacks, veriables)

Is it posible to have any information why connection closed unexpectedly? How I see by source code (may be I something lost) on these cases you cleanup any information about last errors, closes socket and reset all befere call of callback :( If this is not posible will this posible in future? 

    BaseConnection:

        def _on_connection_closed(self, method_frame, from_adapter=False):

            if method_frame and self._is_connection_close_frame(method_frame):
                self.closing = (method_frame.method.reply_code,

            if self.heartbeat:
                self.heartbeat.stop()

            # after this no any info about socket error
            # since: "self.socket = None" on _adapter_disconnect
            # without any saving of last error information :(((((((((((((
            #
            if not from_adapter:
                self._adapter_disconnect()  

            self._on_disconnect(self.closing[0], self.closing[1])

How I see by logger, you have like this:  "WARNING:pika.adapters.base_connection:Socket closed when connection was open" but no any callbacks or variables where I can get this error information without parsing of logger output.

### LINKED ISSUE #682: Pika drops Connection.Close from broker before Connection.Open-Ok
When running in Fedora with RabbitMQ 3.6.0, I noticed that the test async_adapter_tests.TestZ_AccessDenied times out. This test attempts to open a connection using a non-existent vhost. In this scenario, RabbitMQ eventually closes the socket connection (about 30 seconds after Connection.Open with the non-existent vhost from the client).

It turns out that RabbitMQ 3.6.0 sends a Connection.Close immediately after receiving Connection.Open with bad vhost from the client. However, there is a bug in pika: pika doesn't register to handle Connection.Close from the broker until it receives Connection.Open-Ok, so pika just drops the Connection.Close frame in this scenario. Pika eventually detects that the connection is closed, subject to whenever RabbitMQ decides to close the connection.

pika needs to register its handler for Connection.Close before initiating the connection. `Connection._init_connection_state` might be a good place for it, right next to the call `self._add_connection_start_callback()`.

vitaly-krugl:
One of the changes in RabbitMQ 3.6.0 was for the broker to send Connection.Close to client upon connection setup error, instead of just dropping the connection. Thus, the broker is now able to communicate a specific error code and error message to the client.

One visible side-effect of this RabbitMQ change is that the broker waits for about 30 seconds to close the socket connection after sending Connection.Close to the client. I observed this on both Fedora Linux and Mac OS X. Prior to this PR, pika would drop Connection.Close that was received before completion of the AMQP connection handshake, and thus would not discover the connection error for about 30 seconds. Among other obvious issues with not handling Connection.Close, it was also resulting in failure of the acceptance test that deals with connection-establishment error (the one that uses an invalid vhost).

Fixes #682
Fixes #648 
Fixes #631
Fixes #534
Fixes #634

1. Handle Connection.Close from broker at any time, including during connection setup, per RabbitMQ 3.6.0
2. Refactored and cleaned up handling of connection termination;
3. Facilitated passing of error information from failed socket to user callbacks. (fixes #648, #631, #534)
4. Ensure that socket connection is dropped when heartbeat module detects heartbeat timeout (fixes #634)

### DIFF
--- a/docs/version_history.rst
+++ b/docs/version_history.rst
@@ -10,6 +10,8 @@ Next Release
  - In BaseConnection.close, call _handle_ioloop_stop only if the connection is
    already closed to allow the asynchronous close operation to complete
    gracefully.
+ - Pass error information from failed socket connection to user callbacks
+   on_open_error_callback and on_close_callback with result_code=-1.
 
 0.10.0 2015-09-02
 -----------------
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -10,7 +10,6 @@ import ssl
 
 import pika.compat
 from pika import connection
-from pika import exceptions
 
 try:
     SOL_TCP = socket.SOL_TCP
@@ -52,10 +51,10 @@ class BaseConnection(connection.Connection):
 
         :param pika.connection.Parameters parameters: Connection parameters
         :param method on_open_callback: Method to call on connection open
-        :param on_open_error_callback: Method to call if the connection cant
-                                       be opened
-        :type on_open_error_callback: method
-        :param method on_close_callback: Method to call on connection close
+        :param method on_open_error_callback: Called if the connection can't
+            be established: on_open_error_callback(connection, str|exception)
+        :param method on_close_callback: Called when the connection is closed:
+            on_close_callback(connection, reason_code, reason_text)
         :param object ioloop: IOLoop object to use
         :param bool stop_ioloop_on_close: Call ioloop.stop() if disconnected
         :raises: RuntimeError
@@ -152,38 +151,9 @@ class BaseConnection(connection.Connection):
     def _adapter_disconnect(self):
         """Invoked if the connection is being told to disconnect"""
         try:
-            self._remove_heartbeat()
             self._cleanup_socket()
-            self._check_state_on_disconnect()
         finally:
-            # Ensure proper cleanup since _check_state_on_disconnect may raise
-            # an exception
             self._handle_ioloop_stop()
-            self._init_connection_state()
-
-    def _check_state_on_disconnect(self):
-        """Checks to see if we were in opening a connection with RabbitMQ when
-        we were disconnected and raises exceptions for the anticipated
-        exception types.
-
-        """
-        if self.connection_state == self.CONNECTION_PROTOCOL:
-            LOGGER.error('Incompatible Protocol Versions')
-            raise exceptions.IncompatibleProtocolError
-        elif self.connection_state == self.CONNECTION_START:
-            LOGGER.error("Socket closed while authenticating indicating a "
-                         "probable authentication error")
-            raise exceptions.ProbableAuthenticationError
-        elif self.connection_state == self.CONNECTION_TUNE:
-            LOGGER.error("Socket closed while tuning the connection indicating "
-                         "a probable permission error when accessing a virtual "
-                         "host")
-            raise exceptions.ProbableAccessDeniedError
-        elif self.is_open:
-            LOGGER.warning("Socket closed when connection was open")
-        elif not self.is_closed and not self.is_closing:
-            LOGGER.warning('Unknown state on disconnect: %i',
-                           self.connection_state)
 
     def _cleanup_socket(self):
         """Close the socket cleanly"""
@@ -272,11 +242,14 @@ class BaseConnection(connection.Connection):
         """
         if not error_value:
             return None
+
         if hasattr(error_value, 'errno'):  # Python >= 2.6
             return error_value.errno
-        elif error_value is not None:
+        else:
+            # TODO: this doesn't look right; error_value.args[0] ??? Could
+            # probably remove this code path since pika doesn't test against
+            # Python 2.5
             return error_value[0]  # Python <= 2.5
-        return None
 
     def _flush_outbound(self):
         """Have the state manager schedule the necessary I/O.
@@ -290,21 +263,6 @@ class BaseConnection(connection.Connection):
         # detected in _send_connection_tune_ok, before _send_connection_open is
         # called), etc., etc., etc.
         self._manage_event_state()
-
-    def _handle_disconnect(self):
-        """Called internally when the socket is disconnected already
-        """
-        try:
-            self._adapter_disconnect()
-        except (exceptions.ProbableAccessDeniedError,
-                exceptions.ProbableAuthenticationError) as error:
-            LOGGER.error('disconnected due to %r', error)
-            self.callbacks.process(0,
-                                   self.ON_CONNECTION_ERROR,
-                                   self,
-                                   self, error)
-
-        self._on_connection_closed(None, True)
 
     def _handle_ioloop_stop(self):
         """Invoked when the connection is closed to determine if the IOLoop
@@ -323,9 +281,10 @@ class BaseConnection(connection.Connection):
         :param int|object error_value: The inbound error
 
         """
-        if 'timed out' in str(error_value):
-            raise socket.timeout
+        # TODO: doesn't seem right: docstring defines error_value as int|object,
+        # but _get_error_code expects a falsie or an exception-like object
         error_code = self._get_error_code(error_value)
+
         if not error_code:
             LOGGER.critical("Tried to handle an error where no error existed")
             return
@@ -342,6 +301,8 @@ class BaseConnection(connection.Connection):
         elif self.params.ssl and isinstance(error_value, ssl.SSLError):
 
             if error_value.args[0] == ssl.SSL_ERROR_WANT_READ:
+                # TODO: doesn't seem right: this logic updates event state, but
+                # the logic at the bottom unconditionaly disconnects anyway.
                 self.event_state = self.READ
             elif error_value.args[0] == ssl.SSL_ERROR_WANT_WRITE:
                 self.event_state = self.WRITE
@@ -353,20 +314,21 @@ class BaseConnection(connection.Connection):
             LOGGER.error("Socket Error: %s", error_code)
 
         # Disconnect from our IOLoop and let Connection know what's up
-        self._handle_disconnect()
+        self._on_terminate(-1, repr(error_value))
 
     def _handle_timeout(self):
         """Handle a socket timeout in read or write.
         We don't do anything in the non-blocking handlers because we
         only have the socket in a blocking state during connect."""
-        pass
+        LOGGER.warning("Unexpected socket timeout")
 
     def _handle_events(self, fd, events, error=None, write_only=False):
         """Handle IO/Event loop events, processing them.
 
         :param int fd: The file descriptor for the events
         :param int events: Events from the IO/Event loop
-        :param int error: Was an error specified
+        :param int error: Was an error specified; TODO none of the current
+          adapters appear to be able to pass the `error` arg - is it needed?
         :param bool write_only: Only handle write events
 
         """
@@ -382,10 +344,11 @@ class BaseConnection(connection.Connection):
             self._handle_read()
 
         if (self.socket and write_only and (events & self.READ) and
-            (events & self.ERROR)):
-            LOGGER.error('BAD libc:  Write-Only but Read+Error. '
+                (events & self.ERROR)):
+            error_msg = ('BAD libc:  Write-Only but Read+Error. '
                          'Assume socket disconnected.')
-            self._handle_disconnect()
+            LOGGER.error(error_msg)
+            self._on_terminate(-1, error_msg)
 
         if self.socket and (events & self.ERROR):
             LOGGER.error('Error event %r, %r', events, error)
@@ -427,7 +390,7 @@ class BaseConnection(connection.Connection):
         # Empty data, should disconnect
         if not data or data == 0:
             LOGGER.error('Read empty data, calling disconnect')
-            return self._handle_disconnect()
+            return self._on_terminate(-1, "EOF")
 
         # Pass the data into our top level frame dispatching method
         self._on_data_available(data)
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -394,7 +394,7 @@ class BlockingConnection(object):  # pylint: disable=R0902
                         returning true when it's time to stop processing.
                         Their results are OR'ed together.
         """
-        if self._impl.is_closed:
+        if self.is_closed:
             raise exceptions.ConnectionClosed()
 
         # Conditions for terminating the processing loop:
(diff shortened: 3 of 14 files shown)


vitaly-krugl:
Can anyone help me figure out why `assert_any_call` is failing in the python 2.6 build https://travis-ci.org/pika/pika/jobs/100035773 ?  Many thanks

CC @gst, @gmr

gst:
having a look into that..

gst:
can't reproduce directly on my side..

gst:
the only error I can trigger (in python2.6 but I think it's the same with others versions), is this one : 

```
Test that poll() is properly restarted after receiving EINTR error. ... FAIL

======================================================================
FAIL: Test that poll() is properly restarted after receiving EINTR error.
----------------------------------------------------------------------
  File "/usr/local/lib/python2.7/dist-packages/mock.py", line 1201, in patched
    return func(*args, **keywargs)
  File "/home/gstarck/work/public/python/pika/tests/unit/select_connection_ioloop_tests.py", line 396, in test_eintr
    self.assertEqual(is_resumable_mock.call_count, 1)
AssertionError: 0 != 1
-------------------- >> begin captured logging << --------------------
pika.adapters.select_connection: DEBUG: Using SelectPoller
pika.adapters.select_connection: DEBUG: Using SelectPoller
pika.adapters.select_connection: DEBUG: Starting IOLoop
pika.adapters.select_connection: DEBUG: Stopping IOLoop
--------------------- >> end captured logging << ---------------------
```

I can trigger it, "simply" by executing many instances of the test at the same time, with something like this : 

`$ for i in $(seq 100) ; do ( nosetests -x tests/acceptance/blocking_adapter_test.py:TestUnroutableMessagesReturnedInNonPubackMode  &>/tmp/res$i || echo res$i failed) & done`

this also triggers quite a lot of Timed out errors which is expected given the load generated by the execution of so many tests in // ..

gst:
strange, when you give a look at the debug output : 
```
root: DEBUG: ZZZ self.connection.callbacks.process.call_args_list: [call(0, '_on_connection_error', <pika.connection.Connection object at 0x2f92dd0>, <pika.connection.Connection object at 0x2f92dd0>, Client was disconnected at a connection stage indicating a probable denial of access to the specified virtual host: (1, 'error text')),
```
you see the mock has well been called with the, as far as i see, expected arguments.. 

strange..

gst:
For refs/archive the actual error:
```
======================================================================
FAIL: on_disconnect invokes `ON_CONNECTION_ERROR` with `ProbableAccessDeniedError` and `ON_CONNECTION_CLOSED` callbacks
----------------------------------------------------------------------
  File "/home/travis/build/pika/pika/tests/unit/connection_tests.py", line 228, in test_on_disconnect_invokes_access_on_connection_error_and_closed
    mock.ANY)
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/mock/mock.py", line 999, in assert_any_call
    ), cause)
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/six.py", line 718, in raise_from
    raise value
AssertionError: mock(0, '_on_connection_error', <pika.connection.Connection object at 0x2ec4dd0>, <pika.connection.Connection object at 0x2ec4dd0>, <ANY>) call not found
-------------------- >> begin captured logging << --------------------
pika.callback: DEBUG: Added: {'callback': <bound method Connection._on_connection_error of <pika.connection.Connection object at 0x2ec4dd0>>, 'only': None, 'one_shot': False, 'arguments': None}
pika.callback: DEBUG: Added: {'callback': <bound method Connection._on_connection_start of <pika.connection.Connection object at 0x2ec4dd0>>, 'only': None, 'one_shot': True, 'arguments': None, 'calls': 1}
pika.callback: DEBUG: Added: {'callback': <bound method Connection._on_connection_closed of <pika.connection.Connection object at 0x2ec4dd0>>, 'only': None, 'one_shot': True, 'arguments': None, 'calls': 1}
pika.connection: WARNING: Disconnected from RabbitMQ at localhost:5672 from_adapter=True (1): error text
pika.connection: ERROR: Socket closed while tuning the connection indicating a probable permission error when accessing a virtual host
pika.connection: ERROR: Connection setup failed due to Client was disconnected at a connection stage indicating a probable denial of access to the specified virtual host: (1, 'error text')
pika.callback: DEBUG: Incremented callback reference counter: {'callback': <bound method Connection._on_connection_start of <pika.connection.Connection object at 0x2ec4dd0>>, 'only': None, 'one_shot': True, 'arguments': None, 'calls': 2}
pika.callback: DEBUG: Incremented callback reference counter: {'callback': <bound method Connection._on_connection_closed of <pika.connection.Connection object at 0x2ec4dd0>>, 'only': None, 'one_shot': True, 'arguments': None, 'calls': 2}
root: DEBUG: ZZZ self.connection.callbacks.process.call_args_list: [call(0, '_on_connection_error', <pika.connection.Connection object at 0x2ec4dd0>, <pika.connection.Connection object at 0x2ec4dd0>, Client was disconnected at a connection stage indicating a probable denial of access to the specified virtual host: (1, 'error text')),
 call(0, '_on_connection_closed', <pika.connection.Connection object at 0x2ec4dd0>, <pika.connection.Connection object at 0x2ec4dd0>, 1, 'error text')]
--------------------- >> end captured logging << ---------------------
```

does it reproduce on travis ?

gst:
could be due to mock version..

gst:
@vitaly-krugl 
I'd like to know the mock version used by travis, how can I know ?

in fact: it would be good if before the execution of the tests there was a "pip freeze" executed, so that we can know all versions which are in use by travis..

vitaly-krugl:
@gst, I was off the grid for a few days. Just got back today. Thanks for looking at this problem.

I don't know which version of mock it uses, but https://github.com/pika/pika/blob/master/test-requirements.txt specifies `mock` without a version. This would imply that it would load the most recent version. I like your idea about adding `pip freeze`. Perhaps this is something that could be added via https://github.com/pika/pika/blob/master/.travis.yml?

vitaly-krugl:
@gst, I am able to reproduce these problems with mock 1.3.0 on python 2.7.10.

I've had nothing but trouble with the latest mock.

vitaly-krugl:
@gst, getting rid of the spec arg in the mock allows the tests to pass now. I also switched to a context manager, but for patching, but I don't think that the context manager is important in this case. However, the spec/autospec/spec_set are really valuable, and it's a shame that I can't get the tests to pass using them.

gst:
@vitaly-krugl : good to know about the reproduce :)

and yes for having pip freeze output it's effectively within `.travis.yml` ; could be put after the `  - pip install -r test-requirements.txt` in install section..

vitaly-krugl:
@gst, would you like to submit a PR with the `pip freeze` change? thx

gst:
here it is.. https://github.com/pika/pika/pull/689

vitaly-krugl:
Resolved merge conflict and rebased

PR #689, opened 2016-01-08
Add pip freeze at the end of travis install target

gst:
so that we'll know the exact python package versions in use by travis.


vitaly-krugl:
LGTM, @gst

PR #694, opened 2016-01-18
adapter for Asyncio 

appetito:
Hi!

 I made port of Twisted adapter for asyncio. Currently it is not testet well, but works for me.

### DIFF
--- a/pika/adapters/asyncio_connection.py
+++ b/pika/adapters/asyncio_connection.py
@@ -391,9 +391,10 @@ class AsyncioProtocolConnection(base_connection.BaseConnection):
 
     """
 
-    def __init__(self, parameters, loop=None):
+    def __init__(self, parameters, loop=None, connection_lost_callback=None):
         self.ready = asyncio.Future()
         loop = loop or asyncio.get_event_loop()
+        self.connection_lost_callback = connection_lost_callback
         super().__init__(
             parameters=parameters,
             on_open_callback=self.connection_ready,
@@ -454,6 +455,10 @@ class AsyncioProtocolConnection(base_connection.BaseConnection):
         d, self.ready = self.ready, None
         if d:
             d.set_exception(reason)
+        else:
+            # Let client to handle disconnect
+            if self.connection_lost_callback:
+                self.connection_lost_callback(reason)
 
     def connection_made(self, transport):
         # Tell everyone we're connected
--- a/pika/adapters/asyncio_connection.py
+++ b/pika/adapters/asyncio_connection.py
@@ -229,7 +229,7 @@ class IOLoopAdapter:
 
         :param int deadline: The number of seconds to wait to call callback
         :param method callback_method: The callback method
-        :rtype: twisted.internet.interfaces.IDelayedCall
+        :rtype: asyncio.Handle
 
         """
         return self.loop.call_later(deadline, callback_method)
@@ -237,7 +237,7 @@ class IOLoopAdapter:
     def remove_timeout(self, call):
         """Remove a call
 
-        :param twisted.internet.interfaces.IDelayedCall call: The call to cancel
+        :param asyncio.Handle call: The call to cancel
 
         """
         call.cancel()
@@ -259,7 +259,7 @@ class IOLoopAdapter:
     def remove_handler(self, _):
         # The fileno is irrelevant, as it's the connection's job to provide it
         # to the loop when asked to do so. Removing the handler from the
-        # ioloop is removing it from the loop in Twisted's parlance.
+        # ioloop is removing it from the loop in acyncio's parlance.
         self.loop.remove_reader(self.connection.fileno())
         self.loop.remove_writer(self.connection.fileno())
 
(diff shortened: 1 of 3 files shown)


pika/adapters/asyncio_connection.py
```diff
@@ -0,0 +1,476 @@
...
+    def close(self, reason):
+        self.closed = reason
+        while self._getters:
+            waiter = self._getters.popleft()
+            if not waiter.done():
+                waiter.set_exception(reason)
+        self._queue.clear()
+
+
+class AsyncioChannel:
```
vitaly-krugl: This should probably use modern-style class declaration and be derived from `object`.

pika/adapters/asyncio_connection.py
```diff
@@ -0,0 +1,476 @@
...
+        if event_state & self.connection.WRITE:
+            self.loop.add_writer(self.connection.fileno(), self._handle_write)
+
+
+class AsyncioProtocolConnection(base_connection.BaseConnection):
+    """A hybrid between a Pika Connection and a acyncio Protocol. Allows using
+    asyncio's non-blocking loop.create_connection method for connecting to the
+    server.
+
+    It has one caveat: AcyncioProtocolConnection objects have a ready
```
vitaly-krugl: If this class could be made compatible with the other async adapters, such as SelectConnection, TornadoConnection, and LibevConnection, then it could easily integrate with and take advantage of the existing acceptance testing logic in tests/acceptance/async_adapter_tests.py

pika/adapters/asyncio_connection.py
appetito: it is for python3 only - all classes are modern-style)

pika/adapters/asyncio_connection.py
appetito: This adapter use different approach then  SelectConnection or TornadoConnection, so i suppose existing acceptance testing logic may not fit to him

sashas-lb:
what is the status of this change ? is it going to enter the master branch soon ?

vitaly-krugl:
@sashas-lb: it really cannot enter the master branch without a comprehensive test suite. It's impossible to maintain, otherwise.

pika/adapters/asyncio_connection.py
vitaly-krugl: Is it practical to create an adapter that wraps around the new Asyncio adapter and presents an API that is compatible with `SelectConnection`, which could then be plugged into the existing async acceptance test suite? If not, then a dedicated comprehensive test suite is needed for this adapter, just as there is one for `BlockingConnection`. The comprehensive test suite needs to include both acceptance and unit tests. This is necessary for maintainability of the Asyncio adapter.

carlosefr:
We're using this adapter out-of-tree and haven't had any issues up to now. I've been looking around a bit, mostly to get a feeling of how much work would be involved in properly testing this to maybe help it move forward.

I've reached two conclusions:

1. The amount of work involved would be enormous. Even if there is a way to wrap this adapter in something that can be integrated into the existing test suite, it's very hard to even begin because what needs to be implemented for this to work is far from clear. Existing adapters don't seem to help much here.
2. Being based on the Twisted adapter, it would seem that the better course of action would be to try using the same tests as it does. However, it looks like the Twisted adapter also has little to no tests.

This is a catch-22 situation: The adapter doesn't get included because it would require "a comprehensive test suite" but that's so much work that the chances of it happening are close to none.

On the other hand, including it would be the only way of giving it a fighting chance of getting a minimum amount of tests. At least the same chance the Twisted adapter currently has.

Why not include it as experimental with no maintenance guarantees? Right now it pays off to fix it out-of-tree if it breaks. At least this would give some incentive to fix it in a centralized location.

vitaly-krugl:
I added a comprehensive test suite single-handedly when I rewrote the BlockingConnection adapter, so it can be done if someone is sufficiently motivated.

The twisted adapter and others without sufficient test coverage have a history of breaking and being unstable. It was a conscious decision taken by @gmr and myself to not allow this to happen again in this project.

vitaly-krugl:
@carlosefr, my recommendation is to start writing tests and see how far you you will get. I think you might be surprised by how much progress you will make, as was I when working on BlockingConnection rewrite.

Comprehensive tests have another benefit: they server as code examples and lend stability to the API, helping catch accidental changes of the API.

carlosefr:
I'm not advocating against tests, far from it. Nor am I advocating that it's somehow unreasonable to request that contributions be accompanied with proper tests. But, as with everything, context matters and Pika doesn't provide the proper context for this to be a feasible requirement.

Allow me to elaborate on what I mean by this and I apologise beforehand if it happens to read a bit too harsh or if I repeat bit of my previous comment.

Pika is sufficiently documented if you're consuming its user-facing API, but is seriously lacking if you're looking to extend it (i.e. there is no documentation at all). I've spent a number of hours browsing adapter code and tests and got nowhere. I can't tell what's the interface adapters are supposed to be implementing, nor what a "complete" adapter is supposed to look like. The actual interface is whatever the adapters want it to be, it seems.

Proper testing in this scenario means either a **full test suite** for each adapter or understanding the core bits that are already being tested otherwise. The first option requires a deep level of knowledge in AMQP features and behavior while both options require a **core-developer knowledge** of Pika. There's no way to justify the time investment, mostly because there's no way to estimate an upper bound for that investment and trying to estimate it is already a significant investment by itself.

Other projects enforcing proper test requirements don't have such a wide gap between the effort of contributing something and building the tests to go along with it. This makes it an impossible proposition for outside contributions as most of them will be of the "scratch one's own itch" kind.

Per my experience, most contributors to open-source projects are "drive-by shooters" that won't even reply after the initial patch or pull-request. A small percentage will be willing to add the extra effort to clean the patch up and write tests but, if that requires more than twice the effort, you'll find that **no one** is willing to do it. And I'm being very optimistic, I think.

TL;DR: Poorly tested code clearly marked as such is better than no code at all. It doesn't even have to work in the future. The simple fact that it's there will attract random fixes from whoever is using it despite the warnings (like myself, which looked into this because of a bug we were having that turned out to be our own code's fault). This alone is worth it, IMHO.

vitaly-krugl:
The core if fairly well tested - either directly or indirectly (via existing BlockingConnection and  async adapter tests).

This leaves:
1. Asyncio API review/acceptance by the project's owner (@gmr); I am not at all versed in AsyncIO yet.
2. Unit tests of the Asyncio adapter code (which is a pretty small adapter as far as the amount of code in it is concerned, so should not be too much effort)
3. Asyncio adapter integration tests (since these are user-level, there should be sufficient documentation and existing integration tests can serve as template).

vitaly-krugl:
P.S., I agree that the core could be better documented. While rewriting the BlockingConnection adapter, I made an effort to improve the quality of docstrings in the core and elsewhere where I had trouble understanding during my development effort.

pika/adapters/asyncio_connection.py
dmacnet: This code fails to pass the passed-in asyncio loop to the Future() calls, so it hangs when yielding from .ready if you're not using the default loop.

janLo:
What is the status of this PR? Is it going to get merged anytime soon? Or is there another option available to use pika with asyncio?

michaelklishin:
Note that asyncio adapter was also contributed in #768 and later #805, the latter of which was merged.

PR #696, opened 2016-01-19
Client properties override

gmr:
Address #571 by creating an alternate way to override client properties. I don't think this belongs as a connection parameter, but rather at the Connection object level.

### DIFF
--- a/tests/unit/blocking_connection_tests.py
+++ b/tests/unit/blocking_connection_tests.py
@@ -4,6 +4,7 @@ Tests for pika.adapters.blocking_connection.BlockingConnection
 
 """
 import platform
+import socket
 
 from pika.exceptions import AMQPConnectionError
 
--- a/tests/unit/blocking_connection_tests.py
+++ b/tests/unit/blocking_connection_tests.py
@@ -3,7 +3,7 @@
 Tests for pika.adapters.blocking_connection.BlockingConnection
 
 """
-import socket
+import platform
 
 try:
     from unittest import mock
@@ -17,11 +17,14 @@ except ImportError:
     import unittest
 
 import pika
+from pika import connection
+from pika import exceptions
 from pika.adapters import blocking_connection
 
 
 class BlockingConnectionMockTemplate(blocking_connection.BlockingConnection):
     pass
+
 
 class SelectConnectionTemplate(blocking_connection.SelectConnection):
     is_closed = False
@@ -46,7 +49,8 @@ class BlockingConnectionTests(unittest.TestCase):
             on_open_callback=mock.ANY,
             on_open_error_callback=mock.ANY,
             on_close_callback=mock.ANY,
-            stop_ioloop_on_close=mock.ANY)
+            stop_ioloop_on_close=mock.ANY,
+            client_properties=None)
 
     @patch.object(blocking_connection, 'SelectConnection',
                   spec_set=SelectConnectionTemplate)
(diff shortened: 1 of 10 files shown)


vitaly-krugl:
The concept makes sense to me. Have not completed the review yet.

tests/unit/blocking_connection_tests.py
```diff
@@ -205,6 +210,46 @@ def test_sleep(self, select_connection_class_mock):
...
+                'connection.blocked': True,
+                'consumer_cancel_notify': True,
+                'publisher_confirms': True
+            },
+            'information': 'See http://pika.rtfd.org',
+            'version': pika.__version__
+        }
+        with mock.patch('pika.connection.Connection.connect'):
+            with self.assertRaises(exceptions.ConnectionClosed):
+                conn = blocking_connection.BlockingConnection()
```
vitaly-krugl: The `self.assertDictEqual` call below is unreachable because `blocking_connection.BlockingConnection()` raises an exception with the given setup

tests/unit/blocking_connection_tests.py
```diff
@@ -205,6 +210,46 @@ def test_sleep(self, select_connection_class_mock):
...
+                self.assertDictEqual(conn._client_properties, expectation)
```
vitaly-krugl: BlockingConnection doesn't have the `_client_properties` member, since it's not subclassed from the `Connection` hierarchy. The Connection hierarchy is reachable via `conn._impl`.

vitaly-krugl: I think that it might suffice only to test that the expected `client_properties` value was passed to the next layer (SelectConnection constructor). With unit tests, it might be better to isolate testing to the given layer/function. For this, as in `test_sleep`, simply patch `SelectConnection` and `_process_io_for_connection_setup`, then call `blocking_connection.BlockingConnection()`, and finally use `select_connection_class_mock.assert_called_once_with()` with the appropriate args to verify that the expected `client_properites` value was passed to SelectConnection (it would suffice to use `mock.ANY` for all args except `client_properties`) 

```
@patch.object(blocking_connection, 'SelectConnection',
                   spec_set=SelectConnectionTemplate)
```

tests/unit/tornado_tests.py
```diff
@@ -30,3 +35,41 @@ class TornadoConnectionTests(unittest.TestCase):
     def test_tornado_connection_call_parent(self, mock_init):
         obj = tornado_connection.TornadoConnection()
         mock_init.called_once_with(None, None, False)
+
+    def test_client_properties_default(self):
```
vitaly-krugl: Similarly to my `blocking_connection_test` feedback, this layer's unit test should avoid duplicating the logic from `connection_test` by simply testing to make sure that the expected `connection_properties` value was passed to the next layer (`BaseConnection` constructor)

vitaly-krugl:
Hi @gmr, I completed the code review. I found a couple of issues in the tests and also noted a suggestion for simplifying the test logic in the adapters.

vitaly-krugl:
@gmr: it's been bugging me a bit whether this should be part of ConnectionParameters and URLParameters. It's beginning to feel like these new client settings belong in those classes, along with the likes of `locale` and `virtual_host`, instead of additional constructor args in all of the adapters. The rationale is that all of them are used during the connection handshake, just like `locale` and `virtual_host`.

If you agree, then one challenge that might come up is how to encode the client overrides dict in the URL for URLParameters. There might be an easy solution for it. The user can serialize the overrides dict fragment via JSON and then combine it with the rest of their URL (using `urllib.quote`, if needed). URLParameters can then reverse the process for this new parameter.

vitaly-krugl:
@gmr: Hi Gavin, would you mind considering an alternative proposal for setting the client overrides via ConnectionParameters/URLParameters? Please read the rationale in my previous comment here. The alternative proposal is a functional implementation against my own fork https://github.com/vitaly-krugl/pika/pull/2 and is on top of PR #714. It includes support in all of the connection Parameters classes as well as tests. Please note that only three modules are affected by this proposal: connection.py and two test scripts.

gmr:
Reviewed the changes you made, they work for me.

PR #701, opened 2016-02-07
Implement support for Parameters.blocked_connection_timeout, tests, and examples.

### LINKED ISSUE #653: Connection blocked events do not get processed when calling `consume()`
If RabbitMQ is below a resource limit when I connect, the BlockingConnection can end up in a hanging state if I first publish a message and then directly go into `consume()`, even when using the `inactivity_timeout` parameter.

As far as I can see the problem seems to be that when `consume()` sends the `basic_consume` message it waits for the `OK` from Rabbit and does not call the defined `on_blocked_connection_callback` before entering the actual consumption loop. Obviously I could wait for a second after publishing my message and then call `process_data_events()` before entering `consume()`, but that's not a solution. 

How to reproduce:

- Make sure you have a RabbitMQ which is already below resource limits. For example by setting `disk_free_limit` to some really high value.

- This callback gets called as long as the `time.sleep(1)` is there and the sleep is longer than it takes for the `blocked_connection` event to arrive. If you remove the sleep, then it just hangs and the callback does not get called because it's waiting for Rabbit's `OK` message before entering the consumption loop:

        import time
        import pika

        connection_parameters = pika.ConnectionParameters(
            host='192.168.1.101',
            port=5672,
            credentials=pika.PlainCredentials('guest', 'guest'),
        )

        conn = pika.BlockingConnection(connection_parameters)

        def callback(method):
            print('connection blocked!')

        conn.add_on_connection_blocked_callback(callback)

        chan = conn.channel()

        chan.basic_publish(
            exchange='test_exchange',
            routing_key='key',
            body='test'
        )

        time.sleep(1)

        conn.process_data_events()

        for msg in chan.consume(queue='somequeue', inactivity_timeout=1):
            print(msg)

A possible solution might be to also make this `_flush_output` handle the `connection_blocked` event: https://github.com/pika/pika/blob/master/pika/adapters/blocking_connection.py#L1522

vitaly-krugl:
Fixes #653

(it's more of a practical complement to the request in issue #653)

This PR adds a new feature to counter deadlock brought on by indefinite Connection.Blocked state (RabbitMQ's AMQP extension). Users may configure `blocked_connection_timeout` via `ConnectionParameters` or `URLParameters` with the desired number of seconds to wait for connection to become unblocked (Connection.Unblock). If the timeout expires before connection becomes unblocked, the connection will be torn down, triggering the adapter-specific mechanism for informing client app about the closed connection (e.g., on_close_callback or ConnectionClosed exception) with `reason_code` of `pika.connection.InternalCloseReasons.BLOCKED_CONNECTION_TIMEOUT`.

See also http://www.rabbitmq.com/connection-blocked.html

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -78,6 +78,29 @@ class BaseConnection(connection.Connection):
                                              on_open_error_callback,
                                              on_close_callback)
 
+    def __repr__(self):
+        def get_socket_repr(sock):
+            if sock is None:
+                return None
+
+            sockname = sock.getsockname()
+
+            peername = None
+            try:
+                peername = sock.getpeername()
+            except socket.error:
+                # not connected?
+                pass
+
+            return '%s->%s' % (sockname, peername)
+
+        return (
+            '<%s state=%s socket=%s params=%s>' %
+            (self.__class__.__name__,
+             self.connection_state,
+             get_socket_repr(self.socket),
+             self.params))
+
     def add_timeout(self, deadline, callback_method):
         """Add the callback_method to the IOLoop timer to fire after deadline
         seconds. Returns a handle to the timeout
@@ -315,7 +338,8 @@ class BaseConnection(connection.Connection):
             LOGGER.error("Socket Error: %s", error_code)
 
         # Disconnect from our IOLoop and let Connection know what's up
-        self._on_terminate(-1, repr(error_value))
+        self._on_terminate(connection.InternalCloseReasons.SOCKET_ERROR,
+                           repr(error_value))
 
     def _handle_timeout(self):
         """Handle a socket timeout in read or write.
@@ -349,7 +373,8 @@ class BaseConnection(connection.Connection):
             error_msg = ('BAD libc:  Write-Only but Read+Error. '
                          'Assume socket disconnected.')
             LOGGER.error(error_msg)
-            self._on_terminate(-1, error_msg)
+            self._on_terminate(connection.InternalCloseReasons.SOCKET_ERROR,
+                               error_msg)
 
         if self.socket and (events & self.ERROR):
             LOGGER.error('Error event %r, %r', events, error)
@@ -391,7 +416,9 @@ class BaseConnection(connection.Connection):
         # Empty data, should disconnect
         if not data or data == 0:
             LOGGER.error('Read empty data, calling disconnect')
-            return self._on_terminate(-1, "EOF")
+            return self._on_terminate(
+                connection.InternalCloseReasons.SOCKET_ERROR,
+                "EOF")
 
         # Pass the data into our top level frame dispatching method
         self._on_data_available(data)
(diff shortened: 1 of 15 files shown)


vitaly-krugl:
@gmr, please take a look when you have a moment

pika/adapters/base_connection.py
```diff
@@ -78,6 +78,29 @@ def __init__(self,
                                              on_open_error_callback,
                                              on_close_callback)
 
+    def __repr__(self):
+        def get_socket_repr(sock):
```
gmr: This would probably be a pretty handy function to abstract out to the top level of a connection if it matches the format in the RabbitMQ management plugin.

pika/adapters/blocking_connection.py
```diff
@@ -1118,20 +1162,26 @@ def __init__(self, channel_impl, connection):
...
-        self._message_confirmation_result.reset()
-        self._pending_events = deque()
-        self._consumer_infos = dict()
-
     def __int__(self):
         """Return the channel object as its channel number
 
+        NOTE: inherited from legacy BlockingConnection; might be error-prone;
+        use `channel_number` property instead.
+
```
gmr: Maybe deprecate this?

pika/connection.py
```diff
@@ -320,21 +356,6 @@ class ConnectionParameters(Parameters):
     """Connection parameters object that is passed into the connection adapter
     upon construction.
 
-    :param str host: Hostname or IP Address to connect to
```
gmr: Why the removal of the params from the docstring?

gmr:
I don't have any blocking concerns. Please do review the comments for future consideration.

pika/adapters/base_connection.py
vitaly-krugl: @gmr, I am not sure whether I understand your question correctly: I considered abstracting this repr to the `Connection` class, but then realized that the stack is designed such that `Connection` doesn't know anything about sockets.

The second point regarding
> if it matches the format in the RabbitMQ management plugin.

I think you're recommending to check how the management plugin represents connections visually and match their format, right?

pika/adapters/blocking_connection.py
vitaly-krugl: I considered removing the `BlockingChannel.__int__` (and `Channel.__int__`?) methods. Should I worry about breaking user code that relied on this shorthand way of getting the channel number for `"%i"` or similar in logs, exception messages, etc.?

pika/connection.py
vitaly-krugl: @gmr: I removed the parameters from the `ConnectionParameters` class docstring, because the exact same information was in the constructor's docstring in `ConnectionParameters.__init__`. It was the usual pain point of having to change the same thing in multiple places. What is the benefit of having that same information in both the class and constructor docstrings?

vitaly-krugl:
@gmr, thank you for merging. I replied to all of your feedback in this PR, including several questions for you.

PR #702, opened 2016-02-07
Update heartbeat negotiation tests for the new heartbeat negotiation logic that favors max heartbeat value.

vitaly-krugl:


vitaly-krugl:
Ooops, went to wrong repo

PR #705, opened 2016-02-09
Permit both int and long as delivery_tag in Channel.basic_reject

vitaly-krugl:
This picks up @quantum5's commits from PR #676 + @vitaly-krugl's tests


vitaly-krugl:
@gmr, this replaces PR #676 with proper merge against Pika master. FYI

PR #681, opened 2015-12-29
basic_publisher won't stop if the connection interrupt in example

JetDrag:
Added previously tasks in basic_publisher won't stop if the connection interrupt in example. Set self._stop True when connection accidental disconnection to  abandon messages added previously.

PR #710, opened 2016-02-18
Cleaned up async publisher example in docs

vitaly-krugl:
May preempt #681

This change:

1. Simplifies the stopping logic by getting rid of the `_closing` member and relying only on the `_stopping` member. 
2. Simplifies the re-connection logic by having a single code path for both connecting and reconnecting in the `run` method's loop.
3. Get's rid of nested `ioloop.start` calls that used to nest deeper with each reconnect.

### DIFF
--- a/docs/examples/asynchronous_publisher_example.rst
+++ b/docs/examples/asynchronous_publisher_example.rst
@@ -11,7 +11,7 @@ publisher.py::
     import json
 
     LOG_FORMAT = ('%(levelname) -10s %(asctime)s %(name) -30s %(funcName) '
-                  '-35s %(lineno) -5d: %(message)s')
+                    '-35s %(lineno) -5d: %(message)s')
     LOGGER = logging.getLogger(__name__)
 
 
@@ -43,13 +43,14 @@ publisher.py::
             """
             self._connection = None
             self._channel = None
-            self._deliveries = []
-            self._acked = 0
-            self._nacked = 0
-            self._message_number = 0
+
+            self._deliveries = None
+            self._acked = None
+            self._nacked = None
+            self._message_number = None
+
             self._stopping = False
             self._url = amqp_url
-            self._closing = False
 
         def connect(self):
             """This method connects to RabbitMQ, returning the connection handle.
@@ -63,7 +64,8 @@ publisher.py::
             """
             LOGGER.info('Connecting to %s', self._url)
             return pika.SelectConnection(pika.URLParameters(self._url),
-                                         self.on_connection_open,
+                                         on_open_callback=self.on_connection_open,
+                                         on_close_callback=self.on_connection_closed,
                                          stop_ioloop_on_close=False)
 
         def on_connection_open(self, unused_connection):
@@ -75,16 +77,7 @@ publisher.py::
 
             """
             LOGGER.info('Connection opened')
-            self.add_on_connection_close_callback()
             self.open_channel()
-
-        def add_on_connection_close_callback(self):
-            """This method adds an on close callback that will be invoked by pika
-            when RabbitMQ closes the connection to the publisher unexpectedly.
-
-            """
-            LOGGER.info('Adding connection close callback')
-            self._connection.add_on_close_callback(self.on_connection_closed)
 
         def on_connection_closed(self, connection, reply_code, reply_text):
             """This method is invoked by pika when the connection to RabbitMQ is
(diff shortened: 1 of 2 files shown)


JetDrag:
Sorry @vitaly-krugl , Outlook mobile's relay gateways are blocked in China(the fu*king GFW), so I cant receive any notification from Github till now. I've try the new example and it works fine. It seems that Pika has the ability to identify the unexpect connection error so that you remove the 'reconnect' mothed? BTW, i found something wrong maybe. If the server has the same existing exchange,queue,bind relationship Pika said connection is closed during those things declaring. But the new user guide says that it will skip over.

vitaly-krugl:
@JetDrag, thanks for teaching me about the GFW - I didn't know about this term before :)

Sorry, I am not sure what you mean by "set the connection error event priority highest?". How might we go about doing that?

Regarding
> If the server has the same existing exchange,queue,bind relationship Pika said connection is closed during those things declaring.

The message in the exception or in the pika DEBUG logs should give you more information about why it closed. Perhaps you were trying to declare something with exclusive, but another connection was already using that object. Perhaps your app used an invalid exchange type. It could be a number of things. As usual, a simple script that reproduces the problem would be a good way to proceed after looking at DEBUG logs.

JetDrag:
It seems to be my fault that I forgot to set 'durable' to the exchange which i did before.

About the event priority, My thought is that adding a arg to the timeout callback to indicate whether the timer callback should delay when some reconnect event happen. If the arg is True, the timeout callback will invoked after the  reconnect event. It's very useful when I do some periodic but not time-accurate tasks.

And I found a situation that when the eventloop is blocked e.g blocked on comsume, the heartbeat won't send normal even using select adapter. Is that a bug？

vitaly-krugl:
@JetDrag, thanks for the update. 

Regarding timeout and event priority, it sounds like you're talking about  the `SelectConnection.add_timeout` API. It wold probably add some unnecessary complexity in the timer logic to accomplish this, and it should be easy to manage it in the app's own logic (just cancel the timer and start a new one when needed).

When you say "when the eventloop is blocked e.g blocked on comsume", I think you mean the SelectConnection's consume callback is called and your own code is blocking it, right? The async connection types (SelectConnection, TornadoConnection, and TwistedConnection) are intended for I/O-bound applications and all execute entirely in the user thread. Therefore, if the user thread is blocked inside a callback, then the pika async adapter can't execute any other code until control returns from the callback to the async adapter's event loop.

vitaly-krugl:
@JetDrag, I recently prototyped "ThreadedConnection" in https://github.com/vitaly-krugl/pika/tree/threaded-connection, which is a superset of of "BlockingConnection", and has a background thread that handles connection-level I/O with the AMQP broker. This prototype ThreadedConnection will send and receive heartbeats and other data while the user thread is blocked. It also lets you share a single physical connection from multiple threads via connection "cloning".

ThreadedConnection is just a prototype, so it's not good for production. In particular, if you're consuming, you want to set a reasonable Basic.QoS and only use `noack=False` consumers. Otherwise, it will end up loading the contents of the entire queue into the app's memory, if the app can't keep up with the incoming messages; this could be particularly bad for the `noack=True` consumers, since the broker will discard all sent messages, even if the app runs out of memory and crashes.

JetDrag:
@vitaly-krugl Many Thanks to your contribution to Pika. I'm now using the similar solutions to slove the heartbeat problem that use a thread to run the tasks reveived in basic_consume and ack or nack the message according to the result the thread return. It have the same restrictions.

It seems that the async adapter don't async the code even just IO block in the callback?

vitaly-krugl:
Right, since everything executes in the same thread, when the thread is blocked in the callback, IO is impossible in pika.

When you use multiple threads, be sure to avoid calling into the same connection (or any of its channels) from multiple threads. Pika connections are not thread-safe.

FYI,
Vitaly

JetDrag:
@vitaly-krugl Thanks，I'm take care of it.

JetDrag:
@vitaly-krugl Hi, vitaly, I still have some questions here. Will the selection adapter support double consumer runing on separated channel? And how will it behave if events arrive at the same time?

I'm a new coder ,so I consult you about that.

Many Thanks

vitaly-krugl:
@JetDrag, there are a couple of good groups for questions like this:

1. Concerning pika: https://groups.google.com/forum/#!forum/pika-python
2. Concerning RabbitMQ and miscellaneous AMQP clients: https://groups.google.com/forum/#!forum/rabbitmq-users

Please re-post your question on one of those groups and I will reply there. Perhaps the question and answer will be useful for someone else, too. Those groups are more natural places for people to look for and get answers on this topic.

Best,
Vitaly

PR #712, opened 2016-03-01
Fixed Channel-related bugs in Channel and Connection classes

vitaly-krugl:
@gmr Please review and consider merging.

Fixed a number of bugs discovered through code review in Channel-related logic in `Channel`and `Connection` classes. See commit message for details.

Sorry about combining these changes with pylint cleanup; it was difficult to use pylint productively with so many pylint messages.

I created `pylintrc `at root level with defaults that seem to work well for pika. This should be much easier going forward.

To make it easy to find changes that merit review, I labeled corresponding sections of changed code with **REVIEWME** in comments. Search for this label in channel.py and conneciton.py in the "Files changed" view. Here is a high-level summary of changed files:

Changes that merit review:
* pika/channel.py: merits review
* pika/connection.py: merits review
* tests/unit/channel_tests.py: added test logic
* tests/unit/connection_tests.py: added test logic
* pika/compat.py: minor change
* pika/exceptions.py: minor change

Minor changes and primarily pylint-related fixups:
* pika/adapters/blocking_connection.py: remove the deprecated type arg; everything else is pylint fixup
* pika/adapters/base_connection.py: minor change, primarily in `__repr__`.
* tests/unit/base_connection_tests.py: minor; added test for `__repr__`.
* tests/unit/{content_frame_dispatcher_tests.py → content_frame_assembler_tests.py}
* tests/unit/heartbeat_tests.py: minor fix
*  tests/unit/tornado_tests.py: minor fix
* docs/version_history.rst: minor change
* pylintrc: new file; pylint defaults

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -86,7 +86,7 @@ class _CallbackResult(object):
         """True if the object is in a signaled state"""
         return self._ready
 
-    def signal_once(self, *_args, **_kwargs): # pylint: disable=W0613
+    def signal_once(self, *_args, **_kwargs):
         """ Set as ready
 
         :raises AssertionError: if result was already signalled
@@ -162,7 +162,7 @@ class _CallbackResult(object):
         return self._values
 
 
-class _IoloopTimerContext(object):  # pylint: disable=R0903
+class _IoloopTimerContext(object):
     """Context manager for registering and safely unregistering a
     SelectConnection ioloop-based timer
     """
@@ -197,7 +197,7 @@ class _IoloopTimerContext(object):  # pylint: disable=R0903
         return self._callback_result.is_ready()
 
 
-class _TimerEvt(object):  # pylint: disable=R0903
+class _TimerEvt(object):
     """Represents a timer created via `BlockingConnection.add_timeout`"""
     __slots__ = ('timer_id', '_callback')
 
@@ -220,7 +220,7 @@ class _TimerEvt(object):  # pylint: disable=R0903
         self._callback()
 
 
-class _ConnectionBlockedUnblockedEvtBase(object):  # pylint: disable=R0903
+class _ConnectionBlockedUnblockedEvtBase(object):
     """Base class for `_ConnectionBlockedEvt` and `_ConnectionUnblockedEvt`"""
     __slots__ = ('_callback', '_method_frame')
 
@@ -245,19 +245,17 @@ class _ConnectionBlockedUnblockedEvtBase(object):  # pylint: disable=R0903
         self._callback(self._method_frame)
 
 
-class _ConnectionBlockedEvt(  # pylint: disable=R0903
-        _ConnectionBlockedUnblockedEvtBase):
+class _ConnectionBlockedEvt(_ConnectionBlockedUnblockedEvtBase):
     """Represents a Connection.Blocked notification from RabbitMQ broker`"""
     pass
 
 
-class _ConnectionUnblockedEvt(  # pylint: disable=R0903
-        _ConnectionBlockedUnblockedEvtBase):
+class _ConnectionUnblockedEvt(_ConnectionBlockedUnblockedEvtBase):
     """Represents a Connection.Unblocked notification from RabbitMQ broker`"""
     pass
 
 
-class BlockingConnection(object):  # pylint: disable=R0902
+class BlockingConnection(object):
     """The BlockingConnection creates a layer on top of Pika's asynchronous core
     providing methods that will block until their expected response has
     returned. Due to the asynchronous nature of the `Basic.Deliver` and
@@ -404,7 +402,7 @@ class BlockingConnection(object):  # pylint: disable=R0902
             # __exit__ part
             self._event_dispatch_suspend_depth -= 1
 
-    def _process_io_for_connection_setup(self):  # pylint: disable=C0103
+    def _process_io_for_connection_setup(self):
         """ Perform follow-up processing for connection setup request: flush
         connection output and process input while waiting for connection-open
         or connection-error.
@@ -570,8 +568,7 @@ class BlockingConnection(object):  # pylint: disable=R0902
 
                 evt.dispatch()
 
-    def add_on_connection_blocked_callback(self,  # pylint: disable=C0103
-                                           callback_method):
+    def add_on_connection_blocked_callback(self, callback_method):
         """Add a callback to be notified when RabbitMQ has sent a
         `Connection.Blocked` frame indicating that RabbitMQ is low on
         resources. Publishers can use this to voluntarily suspend publishing,
@@ -589,8 +586,7 @@ class BlockingConnection(object):  # pylint: disable=R0902
         self._impl.add_on_connection_blocked_callback(
             functools.partial(self._on_connection_blocked, callback_method))
 
-    def add_on_connection_unblocked_callback(self,  # pylint: disable=C0103
-                                             callback_method):
+    def add_on_connection_unblocked_callback(self, callback_method):
         """Add a callback to be notified when RabbitMQ has sent a
         `Connection.Unblocked` frame letting publishers know it's ok
         to start publishing again. The callback will be passed the
@@ -767,7 +763,7 @@ class BlockingConnection(object):  # pylint: disable=R0902
         # Prepare `with` context
         return self
 
-    def __exit__(self, tp, value, traceback):
+    def __exit__(self, exc_type, value, traceback):
         # Close connection after `with` context
         self.close()
 
@@ -811,7 +807,7 @@ class BlockingConnection(object):  # pylint: disable=R0902
         return self._impl.basic_nack
 
     @property
-    def consumer_cancel_notify_supported(self):  # pylint: disable=C0103
+    def consumer_cancel_notify_supported(self):
         """Specifies if the server supports consumer cancel notification on the
         active connection.
 
@@ -821,7 +817,7 @@ class BlockingConnection(object):  # pylint: disable=R0902
         return self._impl.consumer_cancel_notify
 
     @property
-    def exchange_exchange_bindings_supported(self):  # pylint: disable=C0103
+    def exchange_exchange_bindings_supported(self):
         """Specifies if the active connection supports exchange to exchange
         bindings.
 
@@ -846,12 +842,12 @@ class BlockingConnection(object):  # pylint: disable=R0902
     publisher_confirms = publisher_confirms_supported
 
 
-class _ChannelPendingEvt(object):  # pylint: disable=R0903
+class _ChannelPendingEvt(object):
     """Base class for BlockingChannel pending events"""
     pass
 
 
-class _ConsumerDeliveryEvt(_ChannelPendingEvt):  # pylint: disable=R0903
+class _ConsumerDeliveryEvt(_ChannelPendingEvt):
     """This event represents consumer message delivery `Basic.Deliver`; it
     contains method, properties, and body of the delivered message.
     """
@@ -871,7 +867,7 @@ class _ConsumerDeliveryEvt(_ChannelPendingEvt):  # pylint: disable=R0903
         self.body = body
 
 
-class _ConsumerCancellationEvt(_ChannelPendingEvt):  # pylint: disable=R0903
+class _ConsumerCancellationEvt(_ChannelPendingEvt):
     """This event represents server-initiated consumer cancellation delivered to
     client via Basic.Cancel. After receiving Basic.Cancel, there will be no
     further deliveries for the consumer identified by `consumer_tag` in
@@ -897,12 +893,12 @@ class _ConsumerCancellationEvt(_ChannelPendingEvt):  # pylint: disable=R0903
         return self.method_frame.method
 
 
-class _ReturnedMessageEvt(_ChannelPendingEvt):  # pylint: disable=R0903
+class _ReturnedMessageEvt(_ChannelPendingEvt):
     """This event represents a message returned by broker via `Basic.Return`"""
 
     __slots__ = ('callback', 'channel', 'method', 'properties', 'body')
 
-    def __init__(self, callback, channel, method, properties, body):  # pylint: disable=R0913
+    def __init__(self, callback, channel, method, properties, body):
         """
         :param callable callback: user's callback, having the signature
             callback(channel, method, properties, body), where
@@ -933,7 +929,7 @@ class _ReturnedMessageEvt(_ChannelPendingEvt):  # pylint: disable=R0903
         self.callback(self.channel, self.method, self.properties, self.body)
 
 
-class ReturnedMessage(object):  # pylint: disable=R0903
+class ReturnedMessage(object):
     """Represents a message returned via Basic.Return in publish-acknowledgments
     mode
     """
@@ -1014,7 +1010,7 @@ class _ConsumerInfo(object):
         return self.state == self.CANCELLED_BY_BROKER
 
 
-class _QueueConsumerGeneratorInfo(object):  # pylint: disable=R0903
+class _QueueConsumerGeneratorInfo(object):
     """Container for information about the active queue consumer generator """
     __slots__ = ('params', 'consumer_tag', 'pending_events')
 
@@ -1037,7 +1033,7 @@ class _QueueConsumerGeneratorInfo(object):  # pylint: disable=R0903
             self.__class__.__name__, self.params, self.consumer_tag)
 
 
-class BlockingChannel(object):  # pylint: disable=R0904,R0902
+class BlockingChannel(object):
     """The BlockingChannel implements blocking semantics for most things that
     one would use callback-passing-style for with the
     :py:class:`~pika.channel.Channel` class. In addition,
@@ -1076,7 +1072,7 @@ class BlockingChannel(object):  # pylint: disable=R0904,R0902
 
     # Broker's basic-ack/basic-nack args when delivery confirmation is enabled;
     # may concern a single or multiple messages
-    _OnMessageConfirmationReportArgs = namedtuple(  # pylint: disable=C0103
+    _OnMessageConfirmationReportArgs = namedtuple(
         'BlockingChannel__OnMessageConfirmationReportArgs',
         'method_frame')
 
@@ -1135,7 +1131,7 @@ class BlockingChannel(object):  # pylint: disable=R0904,R0902
         self._basic_consume_ok_result = _CallbackResult()
 
         # Receives the broker-inititated Channel.Close parameters
-        self._channel_closed_by_broker_result = _CallbackResult(  # pylint: disable=C0103
+        self._channel_closed_by_broker_result = _CallbackResult(
             self._OnChannelClosedByBrokerArgs)
 
         # Receives args from Basic.GetEmpty response
@@ -1291,8 +1287,7 @@ class BlockingChannel(object):  # pylint: disable=R0904,R0902
         self.connection._request_channel_dispatch(self.channel_number)
 
 
-    def _on_consumer_cancelled_by_broker(self,  # pylint: disable=C0103
-                                         method_frame):
+    def _on_consumer_cancelled_by_broker(self, method_frame):
         """Called by impl when broker cancels consumer via Basic.Cancel.
 
         This is a RabbitMQ-specific feature. The circumstances include deletion
@@ -1316,8 +1311,7 @@ class BlockingChannel(object):  # pylint: disable=R0904,R0902
         else:
             self._add_pending_event(evt)
 
-    def _on_consumer_message_delivery(self, channel,  # pylint: disable=W0613
-                                      method, properties, body):
+    def _on_consumer_message_delivery(self, _channel, method, properties, body):
         """Called by impl when a message is delivered for a consumer
 
         :param Channel channel: The implementation channel object
@@ -1474,7 +1468,7 @@ class BlockingChannel(object):  # pylint: disable=R0904,R0902
                     _ReturnedMessageEvt(
                         callback, self, method, properties, body))))
 
-    def basic_consume(self,  # pylint: disable=R0913
+    def basic_consume(self,
                       consumer_callback,
                       queue,
                       no_ack=False,
@@ -1529,7 +1523,7 @@ class BlockingChannel(object):  # pylint: disable=R0904,R0902
             arguments=arguments,
             consumer_callback=consumer_callback)
 
-    def _basic_consume_impl(self,  # pylint: disable=R0913
+    def _basic_consume_impl(self,
                             queue,
                             no_ack,
                             exclusive,
@@ -1645,7 +1639,7 @@ class BlockingChannel(object):  # pylint: disable=R0904,R0902
                 consumer_info.state)
 
             # Assertion failure here signals disconnect between consumer state
-            # in BlockingConnection and Connection
+            # in BlockingChannel and Channel
             assert (consumer_info.cancelled_by_broker or
                     consumer_tag in self._impl._consumers), consumer_tag
 
@@ -1764,7 +1758,7 @@ class BlockingChannel(object):  # pylint: disable=R0904,R0902
         else:
             self._cancel_all_consumers()
 
-    def consume(self, queue, no_ack=False,  # pylint: disable=R0913
+    def consume(self, queue, no_ack=False,
                 exclusive=False, arguments=None,
                 inactivity_timeout=None):
         """Blocking consumption of a queue instead of via a callback. This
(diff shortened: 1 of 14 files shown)


pika/adapters/blocking_connection.py
```diff
@@ -2229,10 +2223,10 @@ def confirm_delivery(self):
         # of publisher acknowledgments
         self._impl.add_on_return_callback(self._on_puback_message_returned)
 
-    def exchange_declare(self, exchange=None,  # pylint: disable=R0913
+    def exchange_declare(self, exchange=None,
                          exchange_type='direct', passive=False, durable=False,
                          auto_delete=False, internal=False,
-                         arguments=None, **kwargs):
```
vitaly-krugl: REVIEWME - remove the deprecated `type` arg

pika/channel.py
```diff
@@ -107,10 +128,23 @@ def add_on_cancel_callback(self, callback):
...
-        :param method callback: The method to call on callback
+        If channel closing is initiated by user (either directly of indirectly
+        by closing a connection with open or opening channels) and closing
+        concludes gracefully without Channel.Close from the broker and without
+        loss of connection, the callback will receive reply_code and reply_text
+        that were passed to `Channel.close` or `Connection.close`.
+
+        If channel was closed due to loss of connection, the callback will
+        receive reply_code and reply_text representing the loss of connection.
+
```
vitaly-krugl: REVIEWME above clarification of callback behavior.

pika/channel.py
```diff
@@ -175,23 +209,59 @@ def basic_cancel(self, callback=None, consumer_tag='', nowait=False):
...
         :param str consumer_tag: Identifier for the consumer
         :param bool nowait: Do not expect a Basic.CancelOk response
-        :raises: ValueError
+
+        :raises ValueError:
 
         """
         self._validate_channel_and_callback(callback)
-        if consumer_tag not in self.consumer_tags:
+
```
vitaly-krugl: REVIEWME `basic_cancel` changes below

pika/channel.py
```diff
@@ -417,20 +489,46 @@ def basic_recover(self, callback=None, requeue=False):
...
 
+        :raises ChannelClosed: if channel is already closed
+        :raises ChannelAlreadyClosing: if channel is already closing
         """
-        if not self.is_open:
-            raise exceptions.ChannelClosed()
-        LOGGER.info('Channel.close(%s, %s)', reply_code, reply_text)
-        if self._consumers:
-            LOGGER.debug('Cancelling %i consumers', len(self._consumers))
-            for consumer_tag in dictkeys(self._consumers):
```
vitaly-krugl: REVIEWME `Channel.close` changes below

pika/channel.py
```diff
@@ -442,16 +540,20 @@ def confirm_delivery(self, callback=None, nowait=False):
...
+            `spec.Basic.Nack`.
         :param bool nowait: Do not send a reply frame (Confirm.SelectOk)
 
         """
         self._validate_channel_and_callback(callback)
-        if (self.connection.publisher_confirms is False or
-            self.connection.basic_nack is False):
+
+        # TODO confirm_deliver should require a callback; it's meaningless
+        # without a user callback to receieve Basic.Ack/Basic.Nack notifications
```
vitaly-krugl: REVIEWME Do you agree with this TODO?

pika/channel.py
```diff
@@ -944,32 +1017,81 @@ def _on_cancelok(self, method_frame):
...
                        method_frame.method.reply_text,
                        self)
-        if self.connection.is_open:
-            self._send_method(spec.Channel.CloseOk())
-        self._set_state(self.CLOSED)
-        self.callbacks.process(self.channel_number, '_on_channel_close', self,
-                               self, method_frame.method.reply_code,
-                               method_frame.method.reply_text)
-        self._cleanup()
+
```
vitaly-krugl: REVIEWME `_on_close` changes below

pika/channel.py
```diff
@@ -944,32 +1017,81 @@ def _on_cancelok(self, method_frame):
...
+            self._set_state(self.CLOSED)
+            try:
+                self.callbacks.process(self.channel_number, '_on_channel_close',
+                                       self, self,
+                                       method_frame.method.reply_code,
+                                       method_frame.method.reply_text)
+            finally:
+                self._cleanup()
+
+    def _on_close_meta(self, reply_code, reply_text):
```
vitaly-krugl: REVIEWME new dedicated function for use by `Connection` class when connection is lost abruptly and channels remain that need to be cleaned up.

pika/channel.py
```diff
@@ -944,32 +1017,81 @@ def _on_cancelok(self, method_frame):
...
         self._set_state(self.CLOSED)
-        self.callbacks.process(self.channel_number, '_on_channel_close', self,
-                               self, 0, '')
-        self._cleanup()
+
+        try:
+            self.callbacks.process(self.channel_number, '_on_channel_close',
+                                   self, self,
+                                   self._closing_code_and_text[0],
+                                   self._closing_code_and_text[1])
```
vitaly-krugl: REVIEWME passes cached reply_code/reply_text to callback.

pika/channel.py
```diff
@@ -983,15 +1105,16 @@ def _on_deliver(self, method_frame, header_frame, body):
...
+
         if consumer_tag in self._cancelled:
             if self.is_open and consumer_tag not in self._consumers_with_noack:
                 self.basic_reject(method_frame.method.delivery_tag)
             return
+
         if consumer_tag not in self._consumers:
-            return self._add_pending_msg(consumer_tag, method_frame,
-                                         header_frame, body)
-        while self._pending[consumer_tag]:
```
vitaly-krugl: REVIEWME removed "pending deliveries" support that was in support of legacy `BlockingConnection` and is no longer needed.

pika/channel.py
```diff
@@ -1051,19 +1174,27 @@ def _on_getok(self, method_frame, header_frame, body):
...
+    def _on_openok(self, method_frame):
         """Called by our callback handler when we receive a Channel.OpenOk and
         subsequently calls our _on_openok_callback which was passed into the
         Channel constructor. The reason we do this is because we want to make
         sure that the on_open_callback parameter passed into the Channel
         constructor is not the first callback we make.
 
-        :param pika.frame.Method frame_unused: Unused Channel.OpenOk frame
+        Suppress the state transition and callback if channel is already in
+        CLOSING state.
```
vitaly-krugl: REVIEWME change in `_on_openok` corresponding to this comment about suppressing.

pika/channel.py
```diff
@@ -283,7 +352,9 @@ def basic_get(self, callback=None, queue='', no_ack=False):
 
         """
         self._validate_channel_and_callback(callback)
+        # TODO Is basic_get meaningful when callback is None?
```
vitaly-krugl: REVIEWME Do you agree with this TODO?

pika/channel.py
```diff
@@ -283,7 +352,9 @@ def basic_get(self, callback=None, queue='', no_ack=False):
...
         self._on_getok_callback = callback
+        # TODO Strangely, not using _rpc for the synchronous Basic.Get
```
vitaly-krugl: REVIEWME Any comment on this TODO?

pika/connection.py
```diff
@@ -634,6 +638,16 @@ class should not be invoked directly but rather through the use of an
     CONNECTION_OPEN = 5
     CONNECTION_CLOSING = 6  # client-initiated close in progress
 
+    _STATE_NAMES = {
```
vitaly-krugl: REVIEWME Added state names for use in `__repr__`

pika/connection.py
```diff
@@ -672,9 +684,21 @@ def __init__(self,
         # Define our callback dictionary
         self.callbacks = callback.CallbackManager()
 
+        # Attributes that will be properly initialized by _init_connection_state
+        # and/or during connection handshake.
```
vitaly-krugl: REVIEWME Initialize attributes in constructor that were previously initialized only elsewhere.

pika/connection.py
```diff
@@ -786,6 +810,10 @@ def channel(self, on_open_callback, channel_number=None):
         :rtype: pika.channel.Channel
 
         """
+        if not self.is_open:
+            raise exceptions.ConnectionClosed(
```
vitaly-krugl: REVIEWME do not allow channel to be created on connection that hasn't completed transition to OPEN state to avoid race condition with connection setup.

pika/connection.py
```diff
@@ -1025,35 +1077,30 @@ def _client_properties(self):
...
-                    # Force any lingering callbacks to be removed
-                    # moved inside else block since channel's _cleanup removes
-                    # callbacks
-                    self.callbacks.cleanup(channel_number)
-        else:
-            self._channels = dict()
+        assert self.is_open, str(self)
 
-    def _combine(self, a, b):
+        for channel_number in dictkeys(self._channels):
```
vitaly-krugl: REVIEWME Now also initiates closing of channels that are in OPENING state (in addition to those in OPEN state)

pika/connection.py
```diff
@@ -1102,14 +1149,15 @@ def _deliver_frame_to_channel(self, value):
...
-                self._reject_out_of_band_delivery(value.channel_number,
-                                                  value.method.delivery_tag)
-            else:
-                LOGGER.warning("Received %r for non-existing channel %i", value,
-                               value.channel_number)
+            # This should never happen and would constitute breach of the
+            # protocol
+            LOGGER.critical(
+                'Received %s frame for unregistered channel %i on %s',
+                value.NAME, value.channel_number, self)
```
vitaly-krugl: REVIEWME There should be no such thing as "out-of-band delivery". It it happens, something is seriously wrong with either the broker's or pika's implementation.

pika/connection.py
```diff
@@ -1102,14 +1149,15 @@ def _deliver_frame_to_channel(self, value):
...
             return
-        return self._channels[value.channel_number]._handle_content_frame(value)
+
+        # pylint: disable=W0212
+        self._channels[value.channel_number]._handle_content_frame(value)
```
vitaly-krugl: REVIEWME `_handle_content_frame` doesn't return anything explicit, and neither does `_deliver_frame_to_channel`

pika/connection.py
```diff
@@ -1296,8 +1327,21 @@ def _on_channel_cleanup(self, channel):
         except KeyError:
             LOGGER.error('Channel %r not in channels',
                          channel.channel_number)
-        if self.is_closing and not self._has_open_channels:
-            self._on_close_ready()
+        if self.is_closing:
```
vitaly-krugl: REVIEWME logic in this block makes sure that all channels completed closing before continuing with closing of the connection.

pika/connection.py
```diff
@@ -1593,10 +1642,8 @@ def _on_terminate(self, reason_code, reason_text):
         for channel in dictkeys(self._channels):
             if channel not in self._channels:
                 continue
-            method_frame = frame.Method(channel, spec.Channel.Close(
-                reason_code,
-                reason_text))
-            self._channels[channel]._on_close(method_frame)
+            # pylint: disable=W0212
+            self._channels[channel]._on_close_meta(reason_code, reason_text)
```
vitaly-krugl: REVIEWME now calls the dedicated channel method `_on_close_meta` instead of overloading the logic of `_on_close`, which should be dedicated to handling `Channel.Close` frames from broker.

pika/connection.py
```diff
@@ -1783,52 +1820,50 @@ def _send_frame(self, frame_value):
...
+        write_buffer = [frame.Method(channel_number, method).marshal(),
                         frame.Header(channel_number, length,
                                      content[0]).marshal()]
         if content[1]:
             chunks = int(math.ceil(float(length) / self._body_max_length))
-            for chunk in range(0, chunks):
-                s = chunk * self._body_max_length
-                e = s + self._body_max_length
-                if e > length:
-                    e = length
```
vitaly-krugl: REVIEWME use more meaningful variable names; also makes pylint happy

pika/connection.py
```diff
@@ -1783,52 +1820,50 @@ def _send_frame(self, frame_value):
...
+            for chunk in xrange(0, chunks):
+                start = chunk * self._body_max_length
+                end = start + self._body_max_length
+                if end > length:
+                    end = length
                 write_buffer.append(frame.Body(channel_number,
-                                               content[1][s:e]).marshal())
-
-        with self._write_lock:
```
vitaly-krugl: REVIEWME removed write-locking, since Connection is not thread-safe by design, and a number of other accesses to the output buffer weren't guarded anyway.

tests/unit/channel_tests.py
```diff
@@ -283,25 +307,9 @@ def test_basic_consume_consumers_callback_value(self):
                                consumer_tag=consumer_tag)
         self.assertEqual(self.obj._consumers[consumer_tag], mock_callback)
 
-    def test_basic_consume_has_pending_list(self):
```
vitaly-krugl: "pending" functionality was removed by this PR

tests/unit/channel_tests.py
```diff
@@ -577,32 +640,6 @@ def test_exchange_declare_raises_channel_closed(self):
         self.assertRaises(exceptions.ChannelClosed, self.obj.exchange_declare,
                           exchange='foo')
 
-    @mock.patch('pika.channel.Channel._rpc')
-    def test_exchange_declare_with_type_arg_raises_deprecation_warning(self,
```
vitaly-krugl: the deprecated `type` arg was removed by this PR

tests/unit/channel_tests.py
```diff
@@ -921,19 +958,6 @@ def test_cleanup(self):
         self.obj.callbacks.cleanup.assert_called_once_with(
             str(self.obj.channel_number))
 
-    def test_get_pending_message(self):
```
vitaly-krugl: corresponding functionality was removed by this PR

tests/unit/channel_tests.py
```diff
@@ -1009,31 +1033,110 @@ def test_on_cancelok_removed_consumer(self):
         self.obj._on_cancelok(frame_value)
         self.assertNotIn(consumer_tag, self.obj._consumers)
 
-    def test_on_cancelok_removed_pending(self):
```
vitaly-krugl: corresponding functionality was removed by this PR

tests/unit/channel_tests.py
```diff
@@ -1282,13 +1429,3 @@ def test_validate_channel_and_callback_raises_value_error_not_callable(self
         self.assertRaises(ValueError, self.obj._validate_channel_and_callback,
                           'foo')
 
-    @mock.patch('logging.Logger.warning')
-    def test_on_close_warning(self, warning):
```
vitaly-krugl: Moved to a sorted position in script, similar to most other methods here.

pika/channel.py
gmr: Yeah, that seems reasonable.

pika/channel.py
gmr: IIRC _rpc doesn't handle the full Basic.GetOk method, header, and body frames.

pika/channel.py
gmr: Why the tuple and trailing comma? Creating an extra data structure for the formatting seems unnecessary.

pika/channel.py
```diff
@@ -944,32 +1015,80 @@ def _on_cancelok(self, method_frame):
...
-            self._send_method(spec.Channel.CloseOk())
-        self._set_state(self.CLOSED)
-        self.callbacks.process(self.channel_number, '_on_channel_close', self,
-                               self, method_frame.method.reply_code,
-                               method_frame.method.reply_text)
-        self._cleanup()
+
+        # AMQP 0.9.1 requires CloseOk response to Channel.Close; Note, we should
+        # not be called when connection is closed
+        self._send_method(spec.Channel.CloseOk())
```
gmr: Shouldn't this come after  the check on 1031?

pika/channel.py
```diff
@@ -944,32 +1015,80 @@ def _on_cancelok(self, method_frame):
...
+        callbacks.
+
+        :param int reply_code: The reply code to pass to on-close callback
+        :param str reply_text: The reply text to pass to on-close callback
+
+        """
+        LOGGER.debug('Handling meta-close on %s', self)
+
+        if not self.is_closed:
+            self._closing_code_and_text = (reply_code, reply_text)
```
gmr: Style nit: redundant parentheses.

pika/connection.py
gmr: Hmm I seem to remember something about consuming with ``noack=True`` with messages in the socket recv buffer prior to the Channel.Close being issued. Even with a high QoS prefetch, could you not have frames that have yet to be decoded when you go to close a channel?

pika/connection.py
```diff
@@ -1545,7 +1594,7 @@ def _on_terminate(self, reason_code, reason_text):
         self._remove_heartbeat()
 
         # Remove connection management callbacks
-        # TODO: This call was moved here verbatim from legacy code and the
+        # TODO This call was moved here verbatim from legacy code and the
         # following doesn't seem to be right: `Connection.Open` here is
         # unexpected, we don't appear to ever register it, and the broker
         # shouldn't be sending `Connection.Open` to us, anyway.
```
gmr: IIRC the original implementation was fairly agnostic to which side of the conversation it was on.

pika/connection.py
```diff
@@ -1667,40 +1714,30 @@ def _read_frame(self):
         """
         return frame.decode_frame(self._frame_buffer)
 
-    def _reject_out_of_band_delivery(self, channel_number, delivery_tag):
```
gmr: Just to reiterate, I think there is a case where you could have a pile of frames on the stack that are unprocessed when a Channel.Close is emitted. My thinking is that issuing the close should be similar to Basic.Cancel. I guess one could make the argument that both should not be respected until their respective Ok frames are received though.

pika/connection.py
```diff
@@ -1783,52 +1820,50 @@ def _send_frame(self, frame_value):
...
                         frame.Header(channel_number, length,
                                      content[0]).marshal()]
         if content[1]:
             chunks = int(math.ceil(float(length) / self._body_max_length))
-            for chunk in range(0, chunks):
-                s = chunk * self._body_max_length
-                e = s + self._body_max_length
-                if e > length:
-                    e = length
+            for chunk in xrange(0, chunks):
```
gmr: IIRC xrange has some portability issues -- pypy? I don't recall exactly what the issue was that had me go back to it.

gmr:
Other than the things I pointed out that deserve conversation or consideration, this looks good. Do you mind rebasing once we've come to a decision about the few points of concern?

pika/channel.py
vitaly-krugl: @gmr, regarding

> IIRC _rpc doesn't handle the full Basic.GetOk method, header, and body frames.

Good point about how pika's `_rpc` implementation works at present. However, since `Basic.Get` is defined by AMQP as synchronous, it should be treated like all the other synchronous requests and not get ahead of the proverbial queue. I had a long back-and-forth about the semantics of `Basic.Get` with @michaelklishin (pivotal/RabbitMQ), and he beat into me that we should treat `Basic.Get` like all the other synchronous requests, and made a believer out of me.

An additional important benefit of treating `Basic.Get` synchronously is that pika would then be able to allow multiple outstanding `Basic.Get` requests on different or same queue on the same channel and be able to do the right thing. Pika's present non-synchronous implementation of `Basic.Get` breaks when you attempt to make an additional `basic_get` call before the prior one completes (even if the second request is against a different channel). This is because there is only one `Channel._on_getok_callback` instance member, and `Channel` can't properly handle multiple outstanding `Basic.Get` requests this way.

pika/channel.py
vitaly-krugl: @gmr, regarding
> Shouldn't this come after the check on 1031?

No, it should not. Per AMQP 0.9.1, each peer must respond to the other peer's `Channel.Close`. This is what this specific compound scenario looks like:

pika: sends `Channel.Close` to broker

broker: sends `Channel.Close` to pika client (e.g., in response to error from prior client request on this channel)

pika gets `Channel.Close` from broker and is required to reply to broker  with `Channel.CloseOk` per AMQP 0.9.1.

broker gets `Channel.Close` from pika client and is required to reply with `Channel.Close.Ok` (even if the broker already sent `Channel.Close` for some reason) 

Since pika client sent `Channel.Close` earlier, it should expect to get for `Channel.CloseOk` from the broker per AMQP 0.9.1.

That's why `self._send_method(spec.Channel.CloseOk())` needs to come before the check `if self.is_closing` on 1031

pika/channel.py
vitaly-krugl: > Style nit: redundant parentheses

Will be addressed in subsequent commit into this PR

pika/connection.py
vitaly-krugl: @gmr, regarding
> Hmm I seem to remember something about consuming with noack=True with messages in the socket recv buffer prior to the Channel.Close being issued. Even with a high QoS prefetch, could you not have frames that have yet to be decoded when you go to close a channel?

This would indicate protocol error either in client or in broker. Here is why: `value.channel_number in self._channels` should continue to evaluate to TRUE until the channel is fully closed: `Channel.Close` to broker and then matching `Chanel.CloseOk` from broker or an unsolicited `Channel.Close` from broker. In either case, there should be no more messages destined to the channel after the channel receives either  `Channel.CloseOk` or `Channel.Close` from the broker, because the `Channel.CloseOk` or `Channel.Close` from the broker effectively flush the channel.

It turns out that there used to be a bug in pika that caused pika to remove a Channel prematurely in one condition, allowing such a race condition as you described to occur. I fixed that bug in this PR. The bug was in `Connection._close_channels`. There is a REVIEWME-tagged comment in this PR that describes the problem and the fix. Duplicating it here for completeness:

Regarding fix in `Connection._close_channels`:
> Now also initiates closing of channels that are in OPENING state (in addition to those in OPEN state). Also doesn't force-close channels that are in CLOSING state, allowing them to finish closing.
The legacy behavior lead to a race condition whereby frames could arrive on a channel that was prematurely removed by this function.

pika/connection.py
vitaly-krugl: @gmr, regarding
> IIRC the original implementation was fairly agnostic to which side of the conversation it was on.

Do you mean that the original implementation implemented both Client and Broker?

pika/connection.py
vitaly-krugl: @gmr, regarding
> Just to reiterate, I think there is a case where you could have a pile of frames on the stack that are unprocessed when a Channel.Close is emitted. My thinking is that issuing the close should be similar to Basic.Cancel. I guess one could make the argument that both should not be respected until their respective Ok frames are received though.

In response to your other comment, I provided a detailed explanation of why that should not be possible, and also described a race condition bug that I fixed in this PR that used to allow for that protocol breach in pika. As a quick summary for completeness here, pika will not remove the channel number from `Connection._channels` until it receives the proper acknowledgement from the broker. Therefore, it would be impossible to receive frames for that channel after it's removed from `Connection._channels` - the broker will not send any more messages on the channel after sending `Basic.CloseOk` or `Basic.Close` to the client on that channel. It would be a violation of the AMQP protocol if it sent more frames after sending `Basic.CloseOk` or `Basic.Close` on the same channel.

Regarding your comparison with `Basic.Cancel`, there is one important difference between it and `Channel.Close`: `Channel.Close` is only supported in synchronous mode - i.e., the broker must respond with `Channel.CloseOk` according to AMQ 0.9.1. On the other hand: although `Basic.Cancel` is defined a synchronous, it may be used asynchronously by setting `nowait=True` (unfortunate in my opinion), therefore eliminating `Basic.CancelOk`, so there is no simple way to tell when the Consumer is fully closed. As the result, `pika.Channel` ends up leaking some data structures (in `self._cancelled`, I think). In any case, `Channel.Close` flow doesn't have the problem that `Basic.Cancel` has. (with Consumers, we could probably safely clean up the leaked consumer-related object(s) after the next syncronous `_rpc` request completes, but that's another story for another day :) )

pika/connection.py
vitaly-krugl: @gmr, regarding
> IIRC xrange has some portability issues -- pypy? I don't recall exactly what the issue was that had me go back to it.

`xrange` is used in a number of places in pika via `pika.compat.xrange`. See `from pika.compat import (xrange, ...` at the top of the file.

vitaly-krugl:
@gmr, I believe that I addressed your core concerns with robust rationale for why the implemented behavior is correct in those cases. Let's discuss further, if needed. Either email or Google Hangouts would work, or we can continue the discussion in this PR.

I am going to add a commit with corrections for the several style-related issues that you pointed out and rebase. I can add more changes after that if we conclude that they are needed.

vitaly-krugl:
@gmr, I rebased against master and made several small changes - see my previous comment about addressing your concerns, etc.

pika/channel.py
gmr: I prefer ``'{}'.format()`` as well. Perhaps post merge cleanup at some point.

pika/connection.py
gmr: It implemented the semantics where it could be used as either.

PR #714, opened 2016-03-06
Refactored connection Parameters, ConnectionParameters, and URLParameters classes to simplify logic and make them more robust

vitaly-krugl:
* Partition the validation and parsing logic into separate parameter-specific methods, instead of long single function.
* Perform validation on assignment via property-setters to avoid accepting invalid values.
* Block accidental assignment of invalid members.
* Implemented unit tests


vitaly-krugl:
@gmr, please review and consider for merging

pika/channel.py
```diff
@@ -15,7 +15,8 @@
 
 
 LOGGER = logging.getLogger(__name__)
-MAX_CHANNELS = 32768
+
+MAX_CHANNELS = 65535  # per AMQP 0.9.1 spec.
```
vitaly-krugl: Prior to this PR, there was a discrepancy between the old value (32768) of this constant and the hard-coded value (65535) used by the old "channels_max" validation logic in `connection.Parameters`.

pika/connection.py
```diff
@@ -53,7 +56,7 @@ class Parameters(object):
     :param str DEFAULT_VIRTUAL_HOST: '/'
     :param str DEFAULT_USERNAME: 'guest'
     :param str DEFAULT_PASSWORD: 'guest'
-    :param int DEFAULT_HEARTBEAT_INTERVAL: None
+    :param int DEFAULT_HEARTBEAT_TIMEOUT: None
```
vitaly-krugl: RabbitMQ actually interprets the heartbeat value that we pass it in Connection.Tune as a timeout rather than an interval between heartbeats. RabbitMQ will send several Heartbeat frames during each heartbeat-timeout period.

pika/connection.py
```diff
@@ -26,6 +27,7 @@
 from pika import spec
 
 from pika.compat import xrange, basestring, url_unquote, dictkeys
+from pika.compat import dict_iteritems
```
gmr: At this point, why not from ``pika import compat``? I don't consider this a blocker, just more of a thought about the coding style of importing a bunch of methods like that.

pika/connection.py
```diff
@@ -48,66 +52,119 @@ class InternalCloseReasons(object):
...
-    :param str DEFAULT_LOCALE: 'en_US'
-    :param int DEFAULT_CONNECTION_ATTEMPTS: 1
-    :param int|float DEFAULT_RETRY_DELAY: 2.0
-    :param int|float DEFAULT_SOCKET_TIMEOUT: 0.25
-    :param bool DEFAULT_SSL: False
-    :param dict DEFAULT_SSL_OPTIONS: {}
-    :param int DEFAULT_SSL_PORT: 5671
-    :param bool DEFAULT_BACKPRESSURE_DETECTION: False
-    :param number DEFAULT_BLOCKED_CONNECTION_TIMEOUT: None
+    :param bool backpressure_detection: `DEFAULT_BACKPRESSURE_DETECTION`
```
gmr: I wonder if we can/should remove this soon given ``connection.blocked`` has been around for some time at this point.

gmr:
Looks good :+1:

pika/connection.py
vitaly-krugl: @gmr, regarding 
> At this point, why not from pika import compat? I don't consider this a blocker, just more of a thought about the coding style of importing a bunch of methods like that.

I agree that this would be better (using the compat namespace), but I stuck to the established precedent in this file. See on previous line: `from pika.compat import xrange, basestring, url_unquote, dictkeys`. I felt that I should either change all the existing use here or stick to the established precedent for consistency. I chose the latter (perhaps unwisely?)

pika/connection.py
vitaly-krugl: @gmr, regarding
> I wonder if we can/should remove this soon given connection.blocked has been around for some time at this point.

I agree. I am adding DEPRECATED comments on it in `ConnectionParameters` and `URLParameters` and also in docs/version_history.rst. I am tagging on this minor change to PR #712 that we're still discussing.

PR #716, opened 2016-03-09
Alternative proposal for supplying client properties overrides via Connection Parameters

vitaly-krugl:


vitaly-krugl:
Was meant for my own fork, first.

Issue #719, opened 2016-03-10
Disconnect raises warning


Pankrat:
Pika raises a warning for every connection that is closed. This is annoying if one logs warnings to Sentry. It also indicates that the disconnect is not done in a clean way (which doesn't seem to be the case).

Here is a minimal example:

```python
import pika
import logging
import sys

logging.basicConfig(stream=sys.stdout, level=logging.WARNING)
pika.BlockingConnection().close()
```

The output for both 0.10.0 and master is:

```log
WARNING:pika.connection:Disconnected from RabbitMQ at localhost:5672 (200): Normal shutdown
```

The RabbitMQ logs don't indicate any problem:

```log
=INFO REPORT==== 10-Mar-2016::17:31:09 ===
accepting AMQP connection <0.5957.2> (127.0.0.1:41447 -> 127.0.0.1:5672)

=INFO REPORT==== 10-Mar-2016::17:31:09 ===
closing AMQP connection <0.5957.2> (127.0.0.1:41447 -> 127.0.0.1:5672)
```

IMHO this shouldn't be raised as a warning - either info or debug would be fine.

vitaly-krugl:
That's a good point. In master, this is now done in `Connection._on_terminate`. Please submit a pull request that changes that `LOGGER.warning` call to `LOGGER.info`. Thanks.

tripleee:
In case you still need to use pika 0.10.0 (Debian Squeeze!) here's a workaround.  It took me a while to figure out where exactly to add the filter.

    import logger
    from pika.connection import LOGGER as pika_logger

    class LoggerFilterNormalCloseIsFine (logger..Filter):
        def filter (self, record):
            return not record.getMessage().endswith('(200): Normal shutdown')
    pika_logger.addFilter(LoggerFilterNormalCloseIsFine())

vitaly-krugl:
Closing via merged PR #720.

Thanks @tripleee for the workaround tip against the released 0.10.0.

PR #720, opened 2016-03-11
Log disconnect as an info - not a warning

Pankrat:
This fixes issue #719. 

I wasn't sure whether to continue to log certain termination reasons as a warning (based on `reason_code`) but it looks to me like this is already taken care of by the callers of `_on_terminate`.


vitaly-krugl:
@Pankrat: thanks, your assessment makes sense. That this doesn't cause trouble is covered by existing tests, so no new tests are needed with this PR. Merging...

PR #722, opened 2016-03-19
Implements callback mechanism for data sending process

dukhlov:
Some requests (like  Basic.Ack) don't require reply.
But It is desired to have information at least was it sent or not.

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -430,10 +430,10 @@ class BaseConnection(connection.Connection):
         total_bytes_sent = 0
         try:
             while self.outbound_buffer:
-                frame = self.outbound_buffer.popleft()
+                frame_buffer = self.outbound_buffer.popleft()
                 while True:
                     try:
-                        num_bytes_sent = self.socket.send(frame)
+                        num_bytes_sent = self.socket.send(frame_buffer.data)
                         break
                     except _SOCKET_ERROR as error:
                         if error.errno == errno.EINTR:
@@ -442,21 +442,23 @@ class BaseConnection(connection.Connection):
                             raise
 
                 total_bytes_sent += num_bytes_sent
-                if num_bytes_sent < len(frame):
+                if num_bytes_sent < len(frame_buffer.data):
                     LOGGER.debug("Partial write, requeing remaining data")
-                    self.outbound_buffer.appendleft(frame[num_bytes_sent:])
+                    frame_buffer.data = frame_buffer.data[num_bytes_sent:]
+                    self.outbound_buffer.appendleft(frame_buffer)
                     break
-
+                else:
+                    frame_buffer.on_frame_sent()
         except socket.timeout:
             # Will only come here if the socket is blocking
             LOGGER.debug("socket timeout, requeuing frame")
-            self.outbound_buffer.appendleft(frame)
+            self.outbound_buffer.appendleft(frame_buffer)
             self._handle_timeout()
 
         except _SOCKET_ERROR as error:
             if error.errno in (errno.EAGAIN, errno.EWOULDBLOCK):
                 LOGGER.debug("Would block, requeuing frame")
-                self.outbound_buffer.appendleft(frame)
+                self.outbound_buffer.appendleft(frame_buffer)
             else:
                 return self._handle_error(error)
 
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -449,8 +449,7 @@ class BlockingConnection(object):  # pylint: disable=R0902
         #   empty outbound buffer and any waiter is ready
         is_done = (lambda:
                    self._closed_result.ready or
-                   (not self._impl.outbound_buffer and
-                    (not waiters or any(ready() for ready in  waiters))))
+                   any(ready() for ready in waiters))
 
         # Process I/O until our completion condition is satisified
         while not is_done():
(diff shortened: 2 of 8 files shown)


vitaly-krugl:
@dukhlov, what's the use case for this pull request? Which problem(s) are you trying to solve?

dukhlov:
@vitaly-krugl, Hi. I little bit of context. I try to use single connection for whole application. For this case I created my own implementation of blocking connection which works over SelectConnection. Then I can create a lot of channels per thread and perform request via single connection. Processing ioloop I do in dedicated thread. It works but my blocking connection's basic_ack returns when data actually is not sent to the socket and it makes my connection not fully blocked :). Original Blocking connection wait until whole outbound buffer become empty, but in my case a couple of threads may write simultaneously and append buffer. And I can wait much longer than it is needed. 

BTW: I see that a lot of bugs fixed in master. When do you plan new release?

vitaly-krugl:
Hi @dukhlov, thanks for the background concerning your PR:

1. Regarding bug-fix releases, that's managed by @gmr. (+1)
2. Regarding sharing a single physical AMQP connection in a multi-threaded environment, I recently created a fairly well-functioning prototype "ThreadedConnection" that shares almost all of the current Blocking Connection's implementation and the entire API (the current BlockingConnection is also based on SelectConnection, as you are most likely aware). You can "clone" a ThreadedConnection instance via the `ThreadedConnection.clone_connection` method, which creates a new instance of ThreadedConnection that shares the same physical AMQP connection with the broker. Each cloned instance of ThreadedConnection may be used in its own thread and it gets the complete BlockingConnection API, including creating many channels, consumers, etc. See https://github.com/vitaly-krugl/pika/blob/threaded-connection/pika/adapters/threaded_connection.py. It presently passes all of the generic BlockingConnection tests (fails a few tests that rely on implementation-dependent logic that differs between BlockingConnection and ThreadedConnection).
 * If you would like to try it, set a reasonable QoS value for use with `noack=False` consumers to avoid overwhelming your app's RAM (in case your consumer logic can't keep up with incoming messages). `noack=True` consumers are a bit of a challenge for blocking-style interfaces in general for RabbitMQ, since RabbitMQ doesn't fully support Basic.Flow, so the blocking interface might end up consuming too much of the queue into RAM before the user of the API can process those messages.

dukhlov:
@vitaly-krugl, Thank you for information! I reviewed your implementation and see that you also implemented callbacks for message sending via socket. So I think implementing this by this patch is possibly small part to make you thread branch closer to upstream master. You ThreadingConnection implementation looks like a bit complicated to me. Because it looks like that we could use SelectConnection as is for sending messages from different threads (it have self.write_lock to handle correct frame ordering and appending outbound buffer) send and read operations are performed by ioloop only and should be executed in single thread. So it is not clear to me why we need a lot of proxies and gateways. It is probably because 10.0 release and current master code are different and it didn't work in described way previously. Anyway for now it is necessary for me to understand may this patch get merged soon? How long should I wait for the next pika release? because I need to plan my futher activities and work.
Probably @gmr could provide more info.

vitaly-krugl:
@dukhlov, is your implementation of blocking connection open-source? If so, please point me to it - I would like to learn how it works. Thx.

vitaly-krugl:
@dukhlov, regarding your questions, I had several goals in my prototype, including:

1. Each thread has access to the full pika BlockingConnection and BlockingChannel API
2. Connection Heartbeats are sent/processed independently of what the client threads are doing to avoid lost connections due to heartbeat timeouts and enable more reasonable heartbeat timeout values (such as 60 seconds default heartbeat timeout in recent RabbitMQ releases)
3. Generally-applicable, application-independent solution

The Gateway Connection represents the shared AMQP connection running in its own thread.

I would like to explore simpler implementations, if practical.

> Everything should be made as simple as possible, but not simpler. (A. Einstein)

dukhlov:
@vitaly-krugl, Yes it is open source https://review.openstack.org/#/c/285239/44/oslo_messaging/_drivers/pika_driver/pika_connection.py

I know that Select connection has problems with synchronisation, only writing is synchronised but there a lot of problems with connection state handling, here is a lot of logic like, if channel.is_open do something with channel. it is not synchronised and if ioloop thread detect connection problem and close it after if channel.is_open but before actual working with channel, errors like NoneType don't have some method could appears. In my patch ThreadSafeSelect connection is needed just for fixing such problems, BTW, a lot of them are already fixed in master. and Blocking connection is just tiny wrapper with waiting executing select connection method and waiting for callback.
Probably this code is less mature then your but I think that it should be so simple. And select connection should be adapted to be able to work in multithreaded environment correctly.
I am newcomer to pika, so if I am missing something please guide me:)
Thank you!

vitaly-krugl:
Hi @dukhlov, 

Sorry, but pika's core adapters do not support threading: see "Is Pika thread safe?" in  http://pika.readthedocs.org/en/0.10.0/faq.html. There is some locking support in the Connection class (`_write_lock`) in master that is incomplete and is in the process of being removed. Threading support is a non-goal and a counter-pattern of pika core.

dukhlov:
@vitaly-krugl,
Yes I saw it but thought that threading support will be completed soon. It is interesting why do you want to remove it? Is overhead for locking in so significant?

vitaly-krugl:
@dukhlov regarding
> Yes I saw it but thought that threading support will be completed soon. It is interesting why do you want to remove it? Is overhead for locking in so significant?
 Show all checks

It's an additional burden on maintenance without a broadly-applicable benefit. Each of the async adapters is intended for single-threaded use in an I/O-bound environment.

gmr:
I'm not keen on the amount of changes here that seem to be for a less than ideal use case.

I see threading as a bad thing in general. It's rarely properly implemented, hard to do properly, and Python's support for threading is terrible.

I could see putting some primitives in place to help with it, but I do not see changing the core asynchronous nature of Pika to support BlockingConnection, when I've wanted to remove it for some time.

gmr:
Reading back to the original intent of the patch, I think it adds needless complexity. I could see adding a Connection level primitive for checking to see that the outbound write buffer is empty. I do appreciate the time and consideration you made in the patch, but don't feel it's right at this time.

PR #724, opened 2016-03-31
Don't allow use of subsequent basic_get calls

### LINKED ISSUE #709: There is only one basic_get callback allowed.
The implementation of `basic_get()` doesn't work if called more than once (before the callback is executed). It seems to only have 1 slot for callbacks [here](https://github.com/pika/pika/blob/0114fc45e7173a9520a4cb0f51ea96b4e9eeafa2/pika/channel.py#L281) as opposed to a list (or otherwise linking it to the specific request). This is, IMO, unexpected behavior. At the very least, it should be clearly documented. Here's a simple test case. Note, the failure scenario has to be quit by hand.

```
$ python2 testpika.py 1
message body
$ python2 testpika.py 2
message body
Basic.GetOk received with no active callback
```

Here's the source for `testpika.py`:
```
#!/usr/bin/env python2

import logging
import sys

import pika


def on_open(connection):
    connection.channel(on_channel_open)


def on_queue_setup(method_frame):
    global channel, num_messages
    props = pika.BasicProperties(content_type='text/plain', delivery_mode=1)
    for i in range(num_messages):
        channel.basic_publish('', 'test', 'message body', props)

    for i in range(num_messages):
        channel.basic_get(on_basic_get, 'test', no_ack=True)

msg_count = 0
def on_basic_get(self, method, properties, body):
    global num_messages, msg_count

    print body

    msg_count += 1
    if msg_count == num_messages:
        connection.close()


def on_channel_open(new_channel):
    global channel
    channel = new_channel

    channel.queue_declare(on_queue_setup, 'test')

num_messages = int(sys.argv[1])

logging.getLogger('pika').addHandler(logging.StreamHandler())
parameters = pika.URLParameters('amqp://guest:guest@localhost:5672/%2F')
channel = None
connection = pika.SelectConnection(parameters=parameters,
                                   on_open_callback=on_open)

try:
    connection.ioloop.start()

except KeyboardInterrupt:
    connection.close()
```

dschep:
This ensures users of future versions of Pika won't get bitten by  #709

Todo:
- [x] Fix `BlockingConnection.basic_get` test.
- [x] Add acceptance test

### DIFF
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -336,8 +336,9 @@ class Channel(object):
         """Get a single message from the AMQP broker. If you want to
         be notified of Basic.GetEmpty, use the Channel.add_callback method
         adding your Basic.GetEmpty callback which should expect only one
-        parameter, frame. For more information on basic_get and its
-        parameters, see:
+        parameter, frame. Due to implementation details, this cannot be called
+        a second time until the callback is executed.  For more information on
+        basic_get and its parameters, see:
 
         http://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.get
 
@@ -354,6 +355,8 @@ class Channel(object):
         """
         self._validate_channel_and_callback(callback)
         # TODO Is basic_get meaningful when callback is None?
+        if self._on_getok_callback is not None:
+            raise exceptions.DuplicateGetOkCallback()
         self._on_getok_callback = callback
         # TODO Strangely, not using _rpc for the synchronous Basic.Get. Would
         # need to extend _rpc to handle Basic.GetOk method, header, and body
@@ -1156,6 +1159,8 @@ class Channel(object):
 
         """
         LOGGER.debug('Received Basic.GetEmpty: %r', method_frame)
+        if self._on_getok_callback is not None:
+            self._on_getok_callback = None
 
     def _on_getok(self, method_frame, header_frame, body):
         """Called in reply to a Basic.Get when there is a message.
--- a/tests/acceptance/enforce_one_basicget_test.py
+++ b/tests/acceptance/enforce_one_basicget_test.py
@@ -1,0 +1,32 @@
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
+
+from mock import MagicMock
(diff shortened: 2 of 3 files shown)


vitaly-krugl:
Hi @dschep, it looks the `Channel._on_getok_callback` is not getting reset when Basic.GetEmpty is received (instead of Basic.GetOk). This is what's causing the failure of the new logic. This is easy to fix in the Channel's `_on_getempty` callback. Please include unit/acceptance tests in your PR that correlate to all your changes.

Thank you,
Vitaly

dschep:
Thanks @vitaly-krugl! I'll make those changes :smile:

dschep:
@vitaly-krugl Basic.GetEmpty case fixed and added an acceptance test.

tests/acceptance/enforce_one_basicget_test.py
```diff
@@ -0,0 +1,32 @@
...
+    import unittest
+
+from mock import MagicMock
+from pika.frame import Method, Header
+from pika.exceptions import DuplicateGetOkCallback
+from pika.channel import Channel
+from pika.connection import Connection
+
+
+class OnlyOneBasicGetTestCase(unittest.TestCase):
```
vitaly-krugl: @dschep This is a unit test that uses mock, not an acceptance test. Therefore it belongs under tests/unit/, but not under tests/acceptance/, which establish live connections with the broker. Also, there is an existing unit test script for Channel unit tests, where these things belong: https://github.com/pika/pika/blob/master/tests/unit/channel_tests.py

pika/channel.py
```diff
@@ -1156,6 +1159,8 @@ def _on_getempty(self, method_frame):
 
         """
         LOGGER.debug('Received Basic.GetEmpty: %r', method_frame)
+        if self._on_getok_callback is not None:
```
vitaly-krugl: @dschep: there is no need to check `if self._on_getok_callback is not None` before setting `self._on_getok_callback = None`. The check is redundant. Just setting `self._on_getok_callback = None` would suffice.

PR #735, opened 2016-04-17
Added appveyor.yml for windows build/test and fixed several core and test bugs on Windows.

vitaly-krugl:
Added appveyor.yml for windows build/test (not deploying yet).

Fixed bugs in code and tests that were causing test failures on Windows. More details below...

Skip SelectConnectionIoloop tests that are not supported on the target platform.

Support errno.EWOULDBLOCK return value from socket.connect_ex on Windows.

Use socket send/recv instead of os.write/read on SelectConnection poller's interrup sockets for compatibility with Windows.

Use socket.send/recv instead of os.write/read in select_connection_ioloop_tests.py for Windows compatibility

Use time.sleep instead of select.select in SelectPoller when there are no sockets for Windows compatibility.

Specify errno and message when raising socket.timeout in mock_timeout so that test_tornado_connection_timeout won't choke on Windows.

Implemented SelectPollerTestPollWithoutSockets unit test

Make BaseConnectin's socket easy to patch without having to patch the builtin that used to lead to hang when testing on windows.

Fix ConnectionTests.test_tornado_connection_timeout failure on windows by patching TornadoConnection._create_tcp_connection_socket

Fix hang when testing test_connection_attempts_with_timeout on Windows by removing socket.getaddrinfo patch.

Increase retry_delay in test_connection_attempts_with_timeout to avoid timer_id collision on Windows, which was causing test to hang

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,86 @@
+# Windows build and test of Pika
+
+environment:
+  erlang_download_url: "http://erlang.org/download/otp_win64_18.3.exe"
+  erlang_exe_path: "C:\\Users\\appveyor\\erlang.exe"
+  erlang_home_dir: "C:\\Users\\appveyor\\erlang"
+
+  rabbitmq_installer_download_url: "https://www.rabbitmq.com/releases/rabbitmq-server/v3.6.1/rabbitmq-server-3.6.1.exe"
+  rabbitmq_installer_path: "C:\\Users\\appveyor\\rabbitmq-server-3.6.1.exe"
+
+  matrix:
+    - PYTHON_ARCH: "32"
+      PYTHONHOME: "C:\\Python27"
+
+
+cache:
+  # RabbitMQ is a pretty big package, so caching it in hopes of expediting the
+  # runtime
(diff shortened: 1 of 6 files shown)


vitaly-krugl:
Hi @gmr,

As discussed, I created an appveyor.yml configuration file for pika in this PR. I also had to fix a number of pika and test bugs that showed up on windows.

The full test suite ran successfully on appveyor  from my fork (https://ci.appveyor.com/project/vitaly-krugl/pika/build/1.0.25).

After merging this change, you will need to go to appveyor.com and sign up for a free open-source account for pika and point it to the pika repo. It's really quick to do this: see https://www.appveyor.com/docs#step-2-add-your-project. The configuration should be very simple, since we're using appveyor.yml (in pika's root).

So far, I configured installation of pika and test requirements, download and starting of the necessary support tools (erlang, rabbitmq, etc.), and running of the test suite. Next steps (not included in this PR) might include generating and uploading Windows pika wheels and/or other artifacts.

vitaly-krugl:
Now generates and saves pika wheel as appveyor artifact.

Issue #736, opened 2016-04-19
__enter__ / __exit__ methods on Connection/Channel


pfhayes:
Would these be beneficial, or are they omitted intentionally? Happy to make a PR myself that introduces this functionality

vitaly-krugl:
Connection and Channel classes are at the asynchronous (non-blocking) core of pika. The context manager methods don't make sense in async API. They are more appropriate for the blocking API. See `__enter__/__exit__` in the `BlockingConnection` class.

pfhayes:
great that's what I was looking for - thanks!

pfhayes:
I notice that the BlockingChannel does not have an __enter__/__exit__ - would that be an appropriate addition vs. closing the BlockingChannel manually?

vitaly-krugl:
@pfhayes, PEP-343 on BlockingChannel sounds reasonable to me. @gmr?

We're hoping to cut a new release in the near future, so this change might not get merged until after the release.

PR #629 added PEP-343 in BlockingConnection. See that PR for test examples. Pika's channels have an additional caveat: if attempting to close a channel that's already closed, the channel raises `ChannelClosed` exception. So, you will need to take this into account in both implementation and tests.

PR #737, opened 2016-04-19
__enter__ / __exit__ on BlockingChannel

pfhayes:
As in issue https://github.com/pika/pika/issues/736

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -1172,6 +1172,15 @@ class BlockingChannel(object):
     def __repr__(self):
         return '<%s impl=%r>' % (self.__class__.__name__, self._impl)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, value, traceback):
+        try:
+            self.close()
+        except exceptions.ChannelClosed:
+            pass
+
     def _cleanup(self):
         """Clean up members that might inhibit garbage collection"""
         self._message_confirmation_result.reset()
--- a/tests/unit/blocking_channel_tests.py
+++ b/tests/unit/blocking_channel_tests.py
@@ -19,6 +19,7 @@ except ImportError:
 from pika.adapters import blocking_connection
 from pika import callback
 from pika import channel
+from pika import exceptions
 from pika import frame
 from pika import spec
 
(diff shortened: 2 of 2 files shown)


pika/adapters/blocking_connection.py
```diff
@@ -1172,6 +1172,15 @@ def __int__(self):
     def __repr__(self):
         return '<%s impl=%r>' % (self.__class__.__name__, self._impl)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, value, traceback):
+        try:
+            self.close()
+        except (exceptions.ChannelClosed, exceptions.ChannelAlreadyClosing):
```
vitaly-krugl: BlockingConnection is synchronous, so I wouldn't expect that it should ever see `exceptions.ChannelAlreadyClosing` percolating to its level. How are you able to reproduce `exceptions.ChannelAlreadyClosing` here?

pfhayes: I'm not able to reproduce it but it seemed like this would be safest - I'm happy to remove this though

vitaly-krugl: It may be confusing/misleading to have something here that's not supposed to happen. Error-handling code, like other code, often gets copied, and unnecessary stuff (if indeed unnecessary) tends to spread around and add to confusion.

pfhayes: okay, I've removed this check

tests/unit/blocking_channel_tests.py
```diff
@@ -65,4 +66,25 @@ def test_basic_consume(self):
             self.obj.basic_consume(mock.Mock(), "queue")
 
             self.assertEqual(self.obj._consumer_infos['ctag0'].state,
-                             blocking_connection._ConsumerInfo.ACTIVE)
\ No newline at end of file
+                             blocking_connection._ConsumerInfo.ACTIVE)
+
+    def test_with(self):
```
vitaly-krugl: The combination of the word "with" with the other words in these new test names results in somewhat confusing names. For example, "test_with_close" raises the question "test what with close?" 😃 

It's desirable to use descriptive names, since that's what will show up in the test logs. So, the more information in the test name, the better. In tests, it's acceptable to use very long names, when needed. For example, `test_with_close`, might be more clear if named something like `test_context_manager_exit_with_closed_channel`. `test_with_raise` could be `test_context_manager_does_not_suppress_exception`, and `test_with` could be `test_context_manager`

vitaly-krugl:
Please addressing test name feedback and squash into a single commit.

pfhayes:
Methods have been renamed and commit squashed

vitaly-krugl:
Thank you @pfhayes, LGTM

vitaly-krugl:
@gmr, you mentioned that you wanted to hold off on non-critical merges until after the upcoming release, so I am not merging this PR right now, unless you decide to.

gmr:
This is trivial enough, I think it's fine. Thanks for the PR.

Issue #728, opened 2016-04-06
Use backpressure_detection mechanism to deal with fast publishing messages. It might be a bug of pika


yangyuliang1981:
Use backpressure_detection mechanism to deal with fast publishing messages.
I guess it is a bug of pika
1. I used add_backpressure_callback,but it raised an exception.
2. I added callback functions **add_on_connection_blocked_callback *_and *_add_on_connection_unblocked_callback**.
   But they are not triggered.
   
   In pika docs, it said "Publishers can use this to voluntarily suspend publishing, instead of relying on back pressure throttling. The callback will be passed the Connection.Blocked method frame."
   
   I think I used them in wrong ways. Any suggestions??

```
import logging
import pika
import time

logging.basicConfig(level=logging.DEBUG)


def on_open(new_connection):
    global connection
    connection = new_connection
    connection.channel(on_channel_open)
    connection.add_backpressure_callback(sleep_seconds)
    connection.add_on_connection_blocked_callback(conn_lock)
    connection.add_on_connection_unblocked_callback(conn_unlock)
    connection.add_timeout(1, publish)


def on_channel_open(new_channel):
    global channel
    channel = new_channel
    channel.exchange_declare(exchange='exchange_name', exchange_type='topic')


def sleep_seconds():
    time.sleep(1)


def conn_lock():
    print 'lock'


def conn_unlock():
    print 'unlock'


def publish():
    global connection
    global channel

    channel.basic_publish(exchange='exchange_name',
                          routing_key='routing_key',
                          body='Test Message' * 100)
    print 'send'
    connection.add_timeout(0.001, publish)


channel = None
parameters = pika.ConnectionParameters(host='192.168.91.71', backpressure_detection=True)
connection = pika.SelectConnection(parameters=parameters, on_open_callback=on_open)
connection.ioloop.start()

```

**Exceptions:**
`
File "C:\Python27\lib\site-packages\pika-0.10.0-py2.7.egg\pika\connection.py", line 1062, in _detect_backpressure
    int(buffer_size / avg_frame_size))
ZeroDivisionError: integer division or modulo by zero
`

yangyuliang1981:
Using the latest code from github, it works ok, while the code from pypi is not work.
The former fix some bugs with connection close.

Not test backpressure yet, another issue.

vitaly-krugl:
@yangyuliang1981, backpressure will likely get deprecated. The blocked/unblocked notifications are the preferred mechanism.

[...]

PR #738, opened 2016-04-20
Fix divide-by-zero error during backpressure calculation.

vitaly-krugl:
Fixes issue #728
Fixed update of bytes_sent in Connection._send_message (issue #728)

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -2165,6 +2165,7 @@ class Connection(object):
 
         self.outbound_buffer += write_buffer
         self.frames_sent += len(write_buffer)
+        self.bytes_sent += sum(len(frame) for frame in write_buffer)
         self._flush_outbound()
         if self.params.backpressure_detection:
             self._detect_backpressure()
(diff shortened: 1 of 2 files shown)


vitaly-krugl:
CC @gmr: Ready for review

gmr:
Are you sure this is the right fix? What about
https://github.com/vitaly-krugl/pika/blob/issue728-backpressure-divide-by-zero/pika/connection.py#L2120

Won't this PR count the # of bytes sent to be 2x what they are?

vitaly-krugl:
@grm: No, `_send_frame` and `_send_message` are two distinct, non-overlapping code paths presently in pika; reference https://github.com/vitaly-krugl/pika/blob/issue728-backpressure-divide-by-zero/pika/connection.py#L2136-L2139

```
        if content:
            self._send_message(channel_number, method, content)
        else:
            self._send_frame(frame.Method(channel_number, method))
```

> Won't this PR count the # of bytes sent to be 2x what they are?

vitaly-krugl:
Rebased against 51fd186f1fabe0da8582609d266df1e28e57e3f4

mattcaldwell:
Any updates on this? Fix works for me.

PR #739, opened 2016-04-21
Avoid error callback from scope of Connection constructor.

vitaly-krugl:
1. Perform actual socket connection from the context of a timer to avoid error callback in the scope of Connection constructor or user code.
2. Remove connection-attempt timer, if any, when cleaning up.
3. Secondary:
   - Eliminate unclosed socket resource leaks from `test_eintr` in `select_connection_ioloop_tests` to make sure that the resource leaks reported by python 3.x test runs weren't from code under test.

**Rationale**: I thought it might be beneficial to fix the connection error callback taking place in the context of the Connection constructor, when socket connection fails with no additional retries (no additional retries is default). This is an error-prone situation that I stumbled on several times already. The problem is that the user might get a callback before the connection constructor returns, so an attempt to access a member variable that would normally hold a reference to the connection in user’s code results in attribute error instead.

**Implementation**: `Connection.connect` schedules a 0-sec timer to perform the rest of the operation in a private method, which would be processed from the context of the event loop after the constructor returns.

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -970,6 +970,9 @@ class Connection(object):
         """
         self.connection_state = self.CONNECTION_CLOSED
 
+        # Holds timer when the initial connect or reconnect is scheduled
+        self._connection_attempt_timer = None
+
         # Used to hold timer if configured for Connection.Blocked timeout
         self._blocked_conn_timer = None
 
@@ -992,6 +995,7 @@ class Connection(object):
         self._frame_buffer = None
         self._channels = None
         self._backpressure_multiplier = None
+        self.remaining_connection_attempts = None
 
         self._init_connection_state()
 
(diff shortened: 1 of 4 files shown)


tests/unit/connection_tests.py
```diff
@@ -504,6 +483,47 @@ def test_close_channels(self):
         with self.assertRaises(AssertionError):
             self.connection._close_channels(200, 'reply text')
 
+    @mock.patch('pika.frame.ProtocolHeader')
```
vitaly-krugl: These tests were renamed and fixed up to call `Connection._on_connect_timer` instead of `Connection.connect`.

vitaly-krugl:
CC @gmr This is ready for review

vitaly-krugl:
Rebased against 51fd186f1fabe0da8582609d266df1e28e57e3f4

pika/connection.py
```diff
@@ -1159,35 +1163,22 @@ def connect(self):
...
-            # or raise an exception. To be forward-compatible with failure
-            # reporting from fully non-blocking connection establishment,
-            # connect() should set INIT state and schedule a 0-second timer to
-            # continue the rest of the logic in a private method. The private
-            # method should use itself instead of connect() as the callback for
-            # scheduling retries.
+        # Schedule a timer callback to start the actual connection logic from
+        # event loop's context, thus avoiding error callbacks in the context of
+        # the caller, which could be the constructor.
+        self._connection_attempt_timer = self.add_timeout(
```
vitaly-krugl: This change also makes it so that there is a single code path for connection-establishment logic and the corresponding success/error-reporting logic, regardless of whether it's the initial connection attempt or a reconnection attempt.

vitaly-krugl:
👍

PR #740, opened 2016-04-21
Add OS X to travis matrix, but only with python 3.5 (to avoid builds taking too long)

vitaly-krugl:
@gmr: this adds OS X build/test to pika. Had to overcome lack of support in Travis for Python and RabbitMQ on OS X builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,106 @@
 language: python
+
+os:
+  - linux
+
 python:
   - 2.6
   - 2.7
   - 3.3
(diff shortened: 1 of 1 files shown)


vitaly-krugl:
According to @solarce's comment in https://github.com/travis-ci/travis-ci/issues/5941, RabbitMQ is not a supported service on OSX out-of-the-box in Travis-CI, yet. Also, see https://docs.travis-ci.com/user/osx-ci-environment/.

Issue #464, opened 2014-04-22
TLS/SSL support is vulnerable to MITM attacks


tiran:
Pika's TLS/SSL code accepts all X.509 certificate because it doesn't verify the peer's cert at all. As consequence of this pika is vulnerable to all active attacks like man-in-the-middle attacks. An adversary can use self-signed certificate to listen to or modify all AMQP traffic. TLS without proper cert validation cancels all security of TLS.

In order to verify the peer you have to do two things:

1) Check that the leaf certificate is signed by a trusted root certification authority. "cert_reqs=ssl.CERT_REQUIRED" enables the check. Trusted root CA certs must be loaded first.

2) Verify that the common name (CN) or subject alternative name (SAN) DNS names matches the peer's FQDN. Python 3.x has ssl.match_hostname().

Both checks MUST be performed in order to validate the certificate correctly.

kherrala:
I would argue that optional support for 2) is required in core, because it is too cumbersome to override this logic. Would it be to opinionated to have hostname check in place if "cert_reqs" is set as in part 1) ? Python 2.x can be supported using https://pypi.python.org/pypi/backports.ssl_match_hostname

I can make a pull request for this change if needed.

kherrala:
@vitaly-krugl Any comment on this?

vitaly-krugl:
@kherrala, I will need to wrap my brain around SSL once again, before I can answer your question. Another area that pika probably doesn't support is ssl renegotiation, as pointed out in #703.

https://github.com/openwebos/libpalmsocket may be a good algorithmic reference for re-handshakes, etc. I rewrote libpalmsocket a while back, adding support for SSL renegotiation, certificates, and workarounds for a number of OpenSSL bugs.

kherrala:
@vitaly-krugl I have released some of the SSL patches we are currently running in production on hundreds of clients with some times poor connectivity. They can be found from my fork in master branch. These are essentially the same changes as in my previous pull request #501. Unfortunately I don't have the time to verify them on current master but they are working fine on my original fork based on Pika 0.9.14. Maybe this can be useful for you if you are intending to fix this?

From my second patch it should be obvious that MITM protection cannot be implemented using ssl_options or the current API. 

I haven't looked at SSL renegotiations, but they don't seem to be a problem for us. I believe they are handled using Python socket wrapper and standard event processing automatically?

vitaly-krugl:
[...]

PR #742, opened 2016-04-24
Added TLS examples with mutual and server-only authentication.

vitaly-krugl:


hasB4K:
Hi @vitaly-krugl,

For now, the hostname of the server certificate is not checked in both cases (like described by you in #464). Is there a way to do this easily with the current implementation of pika? Or should I create a class that inherit of BlockingConnection and do a workaround in a overrided method?
If a workaround is needed, I will be glad to suggest a pull request to a review when I will need to add SSL in pika (which is hopefully soon enough).

I think that a even simpler workaround is to provide only your CA in `ca_certs`...

Maybe you should add a notification regarding this issue in your examples :)

Regards,

vitaly-krugl:
Hi @hasB4K, sorry for not replying earlier. I am and will continue to be overwhelmed with other issues for the foreseeable future, so deferring to @gmr.

Best regards,
Vitaly

lukebakken:
Thanks @vitaly-krugl - I updated your changes and included in #840

PR #751, opened 2016-05-17
Added param time_limit to BlockingConnection.start_consuming

Zhomart:
Otherwise `channel.stop_consuming` doesn't stop loop in start_consuming, because `connection.process_data_events` waits data from server forever.

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -1731,7 +1731,7 @@ class BlockingChannel(object):
 
         return unprocessed_messages
 
-    def start_consuming(self):
+    def start_consuming(self, time_limit=None):
         """Processes I/O events and dispatches timers and `basic_consume`
         callbacks until all consumers are cancelled.
 
@@ -1739,6 +1739,7 @@ class BlockingChannel(object):
         pika callback, because dispatching `basic_consume` callbacks from this
         context would constitute recursion.
 
+        :param float time_limit: look at BlockingConnection.process_data_events
         :raises pika.exceptions.RecursionError: if called from the scope of a
             `BlockingConnection` or `BlockingChannel` callback
 
@@ -1752,7 +1753,7 @@ class BlockingChannel(object):
 
         # Process events as long as consumers exist on this channel
         while self._consumer_infos:
-            self.connection.process_data_events(time_limit=None)
+            self.connection.process_data_events(time_limit=time_limit)
 
     def stop_consuming(self, consumer_tag=None):
         """ Cancels all consumers, signalling the `start_consuming` loop to


pika/adapters/blocking_connection.py
```diff
@@ -1731,14 +1731,15 @@ def _remove_pending_deliveries(self, consumer_tag):
...
-    def start_consuming(self):
+    def start_consuming(self, time_limit=None):
         """Processes I/O events and dispatches timers and `basic_consume`
         callbacks until all consumers are cancelled.
 
         NOTE: this blocking function may not be called from the scope of a
         pika callback, because dispatching `basic_consume` callbacks from this
         context would constitute recursion.
 
+        :param float time_limit: look at BlockingConnection.process_data_events
```
vitaly-krugl: What are you trying to accomplish? This is not an adequate explanation.

pika/adapters/blocking_connection.py
```diff
@@ -1752,7 +1753,7 @@ def start_consuming(self):
 
         # Process events as long as consumers exist on this channel
         while self._consumer_infos:
-            self.connection.process_data_events(time_limit=None)
+            self.connection.process_data_events(time_limit=time_limit)
```
vitaly-krugl: How does this help it not wait forever? This won't help the while loop exit, will it?

vitaly-krugl:
@Zhomart, thank you for your submission, but it's not clear exactly which problem you're trying to solve. Please describe the specific scenario in detail.

Zhomart:
I've created a class `RpcServer`

```
class RpcServer:
  def __init__(self):
    self.connection = pika.BlockingConnection(...)
    self.channel = self.connection.channel()
  
  def add(self, name, callable):
    pass
  
  def run(self):
    "This method blocks thread"
    self.channel.start_consuming()
  
  def stop(self):
    self.channel.stop_consuming()
    self.channel.close()
```

Now I want to test it:

```
class TestRpcServer:
  def setUp(self):
    self.rpc_server = RpcServer()
    self.thread = Thread(target=self.rpc_server.run)
  
  def tearDown(self):
    self.rpc_server.stop()
    self.thread.join() # make sure that thread is stopped

  def test_something(self):
    assert True
```

When I run test suite, it goes first `setUp`, then `test_something` and then finally `tearDown`.
But it hangs there.

Then I opened local source code of pika from my `~/.virtualenv` and changed following line:
```
# to 
self.connection.process_data_events(time_limit=0.5) # worked well
```

Second line worked and my test suite finished.

I really don't know how amqp works, but it looks like if `time_limit` is `None`, then `process_data_events` waits data from rabbitmq server.
But when I'm cancelling all subscribers, it doesn't get any data from server and
that is why `process_data_events` waits for data forever.

vitaly-krugl:
@Zhomart, thanks for following up. You example doesn't show how `TestRpcServer.tearDown` gets called. If you're calling `TestRpcServer.tearDown` from another thread, then you're causing a race condition, because pika connections are not thread-safe.

vitaly-krugl:
@Zhomart, perhaps you can try the following:

1. From the thread that runs your pika connection, use your connection's `add_timeout` to schedule a timer callback;
2. In the callback, check a flag in a member variable or global variable that lets it know if it should stop; if the flag is set, then the timer callback would call `stop_consuming`; if the flag is not set, it re-schedules the timer for a new interval to check the flag.
3. Your other thread simply sets the above-mentioned flag and joins the connection thread.

Zhomart:
Thank you @vitaly-krugl . Your suggestion works perfectly. I understood that problems is I'm running `start_consuming` from one thread and `stop_consuming` from another.

I'll close this issue.

vitaly-krugl:
@Zhomart - I'm glad the suggestion worked for you.

PR #778, opened 2016-10-26
use the socket type and proto field from addr info

lexdene:


vitaly-krugl:
@lexdene, which specific problem is this addressing?

lexdene:
@vitaly-krugl, I created this pr because I was confused with the hard code `sock_type` and `sock_proto` param inside `_create_tcp_connection_socket` function.

lexdene:
@vitaly-krugl, I pushed new commits to get rid of the default arg values and add docstring.

vitaly-krugl:
@lexdene - thank you!

PR #780, opened 2016-11-01
Update TwistedProtocolConnection to work with Twisted endpoints

jeremycline:
No other connection adapter has ``parameters`` as a required argument,
and Twisted's ``protocol.Factory.forProtocol`` expects a callable that
requires no arguments to work. This sets the default value for
``parameters`` to be ``None`` to match other adapters

Signed-off-by: Jeremy Cline <jeremy@jcline.org>


lukebakken:
Superceded by #847

PR #781, opened 2016-11-15
Fixing ioloop command call to stopping consumer

priyaaank:
Command that should be called to stop a consumer on ioloop should be `stop`. Fixing the sample consumer in the documentation.

### DIFF
--- a/docs/examples/asynchronous_consumer_example.rst
+++ b/docs/examples/asynchronous_consumer_example.rst
@@ -334,7 +334,7 @@ consumer.py::
             LOGGER.info('Stopping')
             self._closing = True
             self.stop_consuming()
-            self._connection.ioloop.start()
+            self._connection.ioloop.stop()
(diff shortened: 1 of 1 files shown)


docs/examples/asynchronous_consumer_example.rst
```diff
@@ -334,7 +334,7 @@ consumer.py::
             LOGGER.info('Stopping')
             self._closing = True
             self.stop_consuming()
-            self._connection.ioloop.start()
+            self._connection.ioloop.stop()
```
vitaly-krugl: I think there is more to it than this. `stop_consuming` sends a message to the broker and expects a response callback. If you stop the ioloop now, the response won't arrive and the response callback won't be handled. More needs to be done to make these states work correctly. See https://github.com/pika/pika/pull/710/files for an example of cleanup in the Async Publisher code, which is quite similar to the consumer code.

PR #788, opened 2016-12-28
Do not use exec statement.

proofit404:
This PR replace exec statement with direct metaclass call.


lukebakken:
Superceded by #845. Thanks!

PR #789, opened 2017-01-03
Properly decode user name and password passed in as a URL parameter

blkmajik:
When using complex passwords, URLParameters was not decoding the user name nor password parameters passed in the URL.  This patch addresses this issue.


lukebakken:
Thank you! I have included your changes in #842 and added a test.

PR #802, opened 2017-03-13
.Stop() method does not close the IOLoop

cessor:
In Line 334, we call 

Given that this call resides in the stop method, the intention is most likely to stop consuming, close the client, and thus also closing the ioloop. 

According to the documentation at https://github.com/pika/pika/blob/master/pika/adapters/tornado_connection.py#L23

this likely should read 

    self._connection.ioloop.stop()

Likely a minor slip of the pen (or the CTRL+C, V, for that matter). :)

### DIFF
--- a/docs/examples/tornado_consumer.rst
+++ b/docs/examples/tornado_consumer.rst
@@ -331,7 +331,7 @@ consumer.py::
             LOGGER.info('Stopping')
             self._closing = True
             self.stop_consuming()
-            self._connection.ioloop.start()
+            self._connection.ioloop.stop()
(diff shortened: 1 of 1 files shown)


cessor:
I just realised that this is mentioned in the comment, and apoligize for causing work! 

 > The IOLoop is started again because this method is invoked when CTRL-C is pressed raising a KeyboardInterrupt exception. This exception stops the IOLoop which needs to be running for pika to communicate with RabbitMQ. All of the commands issued prior to starting the IOLoop will be buffered but not processed.

Issue #753, opened 2016-05-27
Socket Error: 104 consuming messages with task that take a long time


pat1:
As in #418 I have problem consuming messages with task that take a long time.
I am using git master setting heartbeat_interval=0.

My code is very similar to 
http://pika.readthedocs.io/en/0.10.0/examples/asynchronous_consumer_example.html
but my on_message method take some minutes to "consume" the message.

after some time consuming message I get:

```
INFO 2016-05-27 16:28:51,083 __main__ connect  101 : Connecting to localhost
INFO 2016-05-27 16:28:51,084 pika.adapters.base_connection _create_and_connect_to_socket  216 : Connecting to ::1:5672
INFO 2016-05-27 16:28:51,089 __main__ on_connection_open  117 : Connection opened
INFO 2016-05-27 16:28:51,089 __main__ add_on_connection_close_callback  126 : Adding connection close callback
INFO 2016-05-27 16:28:51,089 __main__ open_channel  169 : Creating a new channel
INFO 2016-05-27 16:28:51,092 __main__ on_channel_open  181 : Channel opened
INFO 2016-05-27 16:28:51,092 __main__ add_on_channel_close_callback  193 : Adding channel close callback
INFO 2016-05-27 16:28:51,092 __main__ start_consuming  223 : Issuing consumer related RPC commands
INFO 2016-05-27 16:28:51,092 __main__ add_on_cancel_callback  233 : Adding consumer cancellation callback
INFO 2016-05-27 16:28:51,102 __main__ on_message  263 : Received message # 1 from None
INFO 2016-05-27 16:43:45,104 __main__ run  72  : Done
INFO 2016-05-27 16:43:45,883 __main__ acknowledge_message  312 : Acknowledging message 1
ERROR 2016-05-27 16:43:45,884 pika.adapters.base_connection _handle_error  362 : Socket Error: 104
INFO 2016-05-27 16:43:45,884 pika.connection _on_terminate  1891: Disconnected from RabbitMQ at localhost:5672 (-1): error(104, 'Connection reset by peer')
WARNING 2016-05-27 16:43:45,885 __main__ on_channel_closed  209 : Channel 1 was closed: (-1) error(104, 'Connection reset by peer')
WARNING 2016-05-27 16:43:45,886 pika.connection close  1135: Suppressing close request on <SelectConnection CLOSED socket=None params=<ConnectionParameters host=localhost port=5672 virtual_host=/ ssl=False>>
[...]

PR #803, opened 2017-03-16
blocking_adapter: only consider processable events

westphahl:
In `process_data_events()` the`common_terminator` that is passed to
`_flush_output()` was also set to true for events that could not be
processed because of the call context.

Considering those events anyway will lead to `_flush_output()` returning
immediately.

This can lead to a dropped connection e.g. when there is a timeout
event ready but `process_data_events()` is called from a callback
(e.g. `on_message`). The timeout event is considered as a terminator in
this case but can not be processed.

Fixes: #753

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -699,8 +699,10 @@ class BlockingConnection(object):
             until I/O produces actionalable events. Defaults to 0 for backward
             compatibility. This parameter is NEW in pika 0.10.0.
         """
-        common_terminator = lambda: bool(
-            self._channels_pending_dispatch or self._ready_events)
+        with self._acquire_event_dispatch() as dispatch_acquired:
+            # Check if we can actually process pending events
+            common_terminator = lambda: bool(dispatch_acquired and
+                (self._channels_pending_dispatch or self._ready_events))
 
         if time_limit is None:
             self._flush_output(common_terminator)


westphahl:
The problem can be reproduce with the following setup:

```python
import logging
import pika


class Consumer:

    def run(self, queue):
        self.connection = connection = pika.BlockingConnection()

        channel.basic_qos(prefetch_count=1, all_channels=True)
        channel.queue_declare(queue=queue, durable=False,
                              exclusive=False, auto_delete=False)
        channel.basic_consume(self.on_message, queue)
        connection.add_timeout(1, self.on_timeout)
        try:
            channel.start_consuming()
        except KeyboardInterrupt:
            channel.stop_consuming()

    def on_timeout(self):
        self.connection.add_timeout(1, self.on_timeout)

    def on_message(self, channel, method, properties, body):
        self.connection.sleep(200)
        channel.basic_ack(method.delivery_tag)


if __name__ == "__main__":
    logging.basicConfig(level=logging.INFO)
    consumer = Consumer()
    consumer.run("foobar")
```

Which will give you the following traceback when the on_message() tries to ack the message.

```
INFO:pika.adapters.base_connection:Connecting to ::1:5672
INFO:pika.adapters.blocking_connection:Created channel=1
ERROR:pika.adapters.base_connection:Socket Error: 104
INFO:pika.connection:Disconnected from RabbitMQ at localhost:5672 (-1): ConnectionResetError(104, 'Connection reset by peer')
ERROR:pika.adapters.blocking_connection:Connection close detected; result=BlockingConnection__OnClosedArgs(connection=<SelectConnection CLOSED socket=None params=<ConnectionParameters host=localhost port=5672 virtual_host=/ ssl=False>>, reason_code=-1, reason_text="ConnectionResetError(104, 'Connection reset by peer')")
  File "/opt/src/bug_demo.py", line 32, in <module>
    consumer.run("foobar")
  File "/opt/src/bug_demo.py", line 17, in run
    channel.start_consuming()
  File "/opt/src/pika/pika/adapters/blocking_connection.py", line 1759, in start_consuming
  File "/opt/src/pika/pika/adapters/blocking_connection.py", line 719, in process_data_events
    self._dispatch_channel_events()
  File "/opt/src/pika/pika/adapters/blocking_connection.py", line 518, in _dispatch_channel_events
    impl_channel._get_cookie()._dispatch_events()
  File "/opt/src/pika/pika/adapters/blocking_connection.py", line 1388, in _dispatch_events
    evt.body)
  File "/opt/src/bug_demo.py", line 26, in on_message
    channel.basic_ack(method.delivery_tag)
  File "/opt/src/pika/pika/adapters/blocking_connection.py", line 1967, in basic_ack
    self._flush_output()
  File "/opt/src/pika/pika/adapters/blocking_connection.py", line 1254, in _flush_output
    *waiters)
  File "/opt/src/pika/pika/adapters/blocking_connection.py", line 474, in _flush_output
    result.reason_text)
pika.exceptions.ConnectionClosed: (-1, "ConnectionResetError(104, 'Connection reset by peer')")
```

lukebakken:
@westphahl I can't reproduce the error you show while testing against RabbitMQ `3.6.10` `stable` branch. I've attached the script as well as output from runs using python 2 and 3.
[repro-py3.txt](https://github.com/pika/pika/files/1171477/repro-py3.txt)
[repro.py.txt](https://github.com/pika/pika/files/1171479/repro.py.txt)
[repro.txt](https://github.com/pika/pika/files/1171478/repro.txt)

pika/adapters/blocking_connection.py
```diff
@@ -699,8 +699,10 @@ def process_data_events(self, time_limit=0):
...
             compatibility. This parameter is NEW in pika 0.10.0.
         """
-        common_terminator = lambda: bool(
-            self._channels_pending_dispatch or self._ready_events)
+        with self._acquire_event_dispatch() as dispatch_acquired:
+            # Check if we can actually process pending events
+            common_terminator = lambda: bool(dispatch_acquired and
+                (self._channels_pending_dispatch or self._ready_events))
 
         if time_limit is None:
```
lukebakken: Once we get to this line, the `yield` in `_acquire_event_dispatch()` has been exited and we no longer hold the resource. The `if` / `else` statements *must* be indented to be part of the `with` block.

westphahl: The context manager `_aquire_event_dispatch()` will yield an integer that is indicating if the resource was acquired (in case > 0).

westphahl: https://github.com/pika/pika/blob/master/pika/adapters/blocking_connection.py#L388

westphahl:
Unfortunately I also can't reproduce the problem anymore. Only difference I saw is that you probably tested it on OS X (KQueuePoller) where I used Linux. Maybe it was the RabbitMQ version.

Thanks anyways for checking the PR!

pika/adapters/blocking_connection.py
lukebakken: Right, but then the `yield` exits, hitting the `finally` block, which decrements `_event_dispatch_suspend_depth`. I realize that this is all happening via a single event loop with callbacks. I assume part of the reason for this code is so that later calls don't try to acquire the same resource. In addition, nowhere else in the code is `_acquire_event_dispatch` called and the value used outside of the `yield`.

lukebakken:
@westphahl - I am re-trying on Arch Linux today.

lukebakken:
Couldn't reproduce on Arch Linux. I opened #843 with a change that ensures the value returned by `_acquire_event_dispatch()` is valid throughout its use. See [this discussion](https://github.com/pika/pika/pull/803#pullrequestreview-51924699)

PR #807, opened 2017-04-02
add delivery mode constants

eshizhan:
```
prop = pika.BasicProperties(
        delivery_mode=pika.spec.DELIVERY_PERSISTENT # use constant instead of number 2
)
```

Signed-off-by: eshizhan <eshizhan@126.com>


pika/spec.py
```diff
@@ -26,6 +27,8 @@
 COMMAND_INVALID = 503
 CONNECTION_FORCED = 320
 CONTENT_TOO_LARGE = 311
+DELIVERY_NONPERSISTENT = 1
```
michaelklishin: Let's rename this to `TRANSIENT_DELIVERY_MODE`.

pika/spec.py
```diff
@@ -26,6 +27,8 @@
...
+DELIVERY_PERSISTENT = 2
```
michaelklishin: Let's rename this to `PERSISTENT_DELIVERY_MODE`

michaelklishin:
We will apply this change with slightly different constant names manually. Thank you, @eshizhan!

michaelklishin:
@lukebakken should it go into `pika.channel` then?

michaelklishin:
@lukebakken and I agreed to keep the constants in `pika.spec` as they are, well, defined in the AMQP 0-9-1 spec :)

michaelklishin:
See #839.

Issue #817, opened 2017-05-16
'Must have completion callback with nowait=False' if closing channel with existing consumers


desertkun:
If you have existing consumers and decide to close a channel, you will end up with an error:
```python
ValueError: Must have completion callback with nowait=False
```
That's because [this line](https://github.com/pika/pika/blob/master/pika/channel.py#L529) definitely doing something wrong (it passes NO callback and sets nowait=False)

lukebakken:
Resolved by #818, thanks!

behconsci:
I am still getting this error with version 0.11.2

```File "/usr/local/lib/python3.6/site-packages/pika/channel.py", line 1125, in _on_deliver
    header_frame.properties, body)
  File "/usr/local/lib/python3.6/site-packages/asgi_rabbitmq/core.py", line 195, in consume_message
    amqp_channel.basic_cancel(consumer_tag=tag)
  File "/usr/local/lib/python3.6/site-packages/pika/channel.py", line 231, in basic_cancel
    'Must have completion callback with nowait=False')
ValueError: Must have completion callback with nowait=False

    amqp_channel.basic_cancel(consumer_tag=tag)
  File "/usr/local/lib/python3.6/site-packages/pika/channel.py", line 231, in basic_cancel
    'Must have completion callback with nowait=False')
ValueError: Must have completion callback with nowait=False```

lukebakken:
@behconsci please provide code to reproduce. Pika's maintainers do so on what amounts to their free time, so adding time to reproduce an issue when only a stack trace is provided hinders fixing it.

lukebakken:
@behconsci - here's the problem. The [default parameters](https://github.com/pika/pika/blob/master/pika/channel.py#L202) for `basic_cancel` *will* cause that exception to happen every time, because `nowait` defaults to `False` which requires a callback. `asgi_rabbitmq` has already addressed this [here](https://github.com/proofit404/asgi_rabbitmq/blob/master/asgi_rabbitmq/core.py#L201-L202), so you need to upgrade `asgi_rabbitmq`.

We may consider changing the method signature to prevent this from happening in the future.

behconsci:
thanks @lukebakken! I upgraded `asgi_rabbitmq` and it worked again.

lukebakken:
@behconsci thank you for letting me know that fixed your issue.

vitaly-krugl:
@lukebakken, this should be fixed now because we removed the `nowait` parameter, right?

[...]

PR #818, opened 2017-05-16
Fixed basic_cancel when closing the channel

desertkun:
See #817


michaelklishin:
It's not actually necessary to cancel consumers when closing a channel (the server will discard them all anyway) but it's a decent start.

PR #840, opened 2017-07-24
Added TLS examples with mutual and server-only authentication - Updated

lukebakken:
Includes changes from #742


vitaly-krugl:
Thanks @lukebakken

PR #844, opened 2017-07-25
Update mock expectation post #818

michaelklishin:


tests/unit/channel_tests.py
```diff
@@ -1,4 +1,4 @@
-"""
+gi"""
```
lukebakken: What does this change do?

michaelklishin: Err, looks like an Emacs artefact. Will fix.

michaelklishin:
@lukebakken thanks for spotting the typo, ready for another round.

tests/unit/channel_tests.py
```diff
@@ -522,7 +522,9 @@ def test_close_basic_cancel_called(self):
         self.obj._consumers['abc'] = None
         with mock.patch.object(self.obj, 'basic_cancel') as basic_cancel:
             self.obj.close()
-            basic_cancel.assert_called_once_with(consumer_tag='abc', nowait=True)
+            # this is actually not necessary but Pika currently cancels
+            # every consumer before closing the channel
+            basic_cancel.assert_called_once_with(consumer_tag='abc', nowait = True)
```
lukebakken: This is super nit-picky, but typical Python style for parameter default assignment is to not use spaces.

PR #860, opened 2017-08-14
add gevent connection adapter

cainbit:
add gevent connection adapter

The adapter has beed used for months in on production projects,  and I'll keep a close eye on issues if the pull request is accepted.

### DIFF
--- a/pika/adapters/gevent_connection.py
+++ b/pika/adapters/gevent_connection.py
@@ -6,19 +6,28 @@ from pika.adapters import BaseConnection
 from pika.heartbeat import HeartbeatChecker
 
 
-class PIKAGeventHeartbeatChecker(HeartbeatChecker):
+class PikaGeventHeartbeatChecker(HeartbeatChecker):
     def send_and_check(self):
-        gevent.spawn(super(PIKAGeventHeartbeatChecker, self).send_and_check)
+        """Send and check heartbeat in a Greenlet object"""
+        gevent.spawn(super(PikaGeventHeartbeatChecker, self).send_and_check)
 
 
-class PIKAGeventConnection(BaseConnection):
+class PikaGeventConnection(BaseConnection):
+    """ A standard Pika connection adapter for gevent.
+
+        :param pika.connection.Parameters parameters: Connection parameters
+        :param on_open_callback: The method to call when the connection is open
+        :type on_open_callback: method
+        :param on_open_error_callback: Method to call if the connection cant be opened
+        :type on_open_error_callback: method
+    """
     def __init__(self,
                  parameters=None,
                  on_open_callback=None,
                  on_open_error_callback=None,
                  on_close_callback=None):
         self.connected = False
-        super(PIKAGeventConnection, self).__init__(
+        super(PikaGeventConnection, self).__init__(
             parameters,
             on_open_callback,
             on_open_error_callback,
@@ -26,16 +35,33 @@ class PIKAGeventConnection(BaseConnection):
         )
 
     def add_timeout(self, deadline, callback_method):
+        """Add the callback_method to gevent hub to fire after deadline
+        seconds.
+
+        :param int deadline: The number of seconds to wait to call callback
+        :param method callback_method: The callback method
+        :rtype: timer instance
+
+        """
         timer = gevent.get_hub().loop.timer(deadline)
         timer.start(callback_method)
         return timer
 
     def remove_timeout(self, timeout_id):
+        """Remove the timer from gevent hub using the instance returned from
+        add_timeout.
+
+        param: timer instance handle
+
+        """
         timer = timeout_id
         timer.stop()
 
     def _adapter_connect(self):
-        error = super(PIKAGeventConnection, self)._adapter_connect()
+        """Connect and read data in a Greenlet object
+
+        """
+        error = super(PikaGeventConnection, self)._adapter_connect()
         if not error:
             self.socket.setblocking(1)
             self.connected = True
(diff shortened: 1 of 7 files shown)


pika/adapters/gevent_connection.py
```diff
@@ -0,0 +1,64 @@
+"""Use pika with the GEVENT"""
+
+import gevent
+
+from pika.adapters import BaseConnection
+from pika.heartbeat import HeartbeatChecker
+
+
+class PIKAGeventHeartbeatChecker(HeartbeatChecker):
```
lukebakken: Please use `Pika` instead of `PIKA`

cainbit: OK, I will fix this and add test in next commit.

michaelklishin:
@cainbit sorry, any updates on this?

michaelklishin:
@cainbit if you have no plans to finish this PR that's fine but please let us know so that another contributor or one of RabbitMQ engineers can step up. Thank you.

kmonson:
How is this coming along? 

The [VOLTTRON](https://github.com/VOLTTRON/volttron) project is exploring RabbitMQ as an alternative to ZMQ going forward. gevent support is required for this to happen. 

We really like what rabbit offers. It would allow us to move away from much (all?) of our custom routing and authorization solution. 

I cannot commit to anything, but we may be able to help out with this if needed.

lukebakken:
@kmonson you're more than welcome to give the `cainbit:gevent_connection_adapter` branch a try. If you have success it will give us more confidence in the changes. I haven't had time to review them or add this adapter to the test suite.

cainbit:
There is still 4 cases fail, I can't point out the reasons, I had wrote some code, not using unittest, to test the same cases on my local env, it runs well.

kmonson:
Could the monkey patching you are doing in the test be causing problems?

cainbit:
@kmonson I also think this is the most probable cause, I'm trying to find a way to fix the test to work with gevent.

tests/acceptance/async_test_base.py
```diff
@@ -205,4 +205,11 @@ def libev_test(self):
         """LibevConnection"""
         self.start(adapters.LibevConnection)
 
+    @unittest.skipIf(adapters.GeventConnection is None, 'gevent is not installed')
+    def libev_test(self):
```
vitaly-krugl: @cainbit -- The name `libev_test` is already used by the test function above

kmonson:
How is this adapter cooperative? I don't see anywhere where you are releasing control to other greenlets in your code. You are still just using the normal stdlib socket after setting it to block.

pika/adapters/gevent_connection.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def _on_connect_timer(self):
+        gevent.spawn(super(GeventConnection, self)._on_connect_timer)
+
+    def _adapter_connect(self):
+        """Connect and read data in a Greenlet object
+
+        """
+        error = super(GeventConnection, self)._adapter_connect()
+        if not error:
```
vitaly-krugl: Since `BaseConnection. _adapter_connect` is documented to return `None` on success, it might be prudent to check for `error is None` instead.

pika/adapters/gevent_connection.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def _adapter_disconnect(self):
+        """Disconnect the connection
+
+        """
+        super(GeventConnection, self)._adapter_disconnect()
+
+    def _handle_error(self, error_value):
+        if self.socket:
+            super(GeventConnection, self)._handle_error(error_value)
```
vitaly-krugl: Interesting, if there is an error reported to `_handle_error`, but there is no socket presently, the error is just suppressed? This doesn't sound right. If there is a legitimate reason for it, then a code comment is in order.

And again, should be checking `self.socket is not None` as objects may override `__bool__` and give it some unpredictable semantic.

vitaly-krugl:
> How is this adapter cooperative? I don't see anywhere where you are releasing control to other greenlets in your code. You are still just using the normal stdlib socket after setting it to block. 

It looks like the implementation is relying on the application to call gevent's [monkey.patch_all()](https://github.com/pika/pika/pull/860/files#diff-290659f337a39e1fce9b0aa969984c4eR212). However, this doesn't seem to be communicated via user documentation anywhere in the pull request.

kmonson:
@vitaly-krugl That was my thought as well. In this case I think you would only need to patch socket. patch_all() is kind of a shotgun approach that is usually not advised.

It seems like this could be more elegantly fixed by overriding _create_tcp_connection_socket to return a gevent socket. Then no patching is required.

vitaly-krugl:
> It seems like this could be more elegantly fixed by overriding _create_tcp_connection_socket to return a gevent socket. Then no patching is required

Yeah, it seems that creating the appropriate gevent socket would be preferable to monkey-patching, which could break some auxiliary things, such as the test framework and other adapters.

13steinj:
any updates on this?

lukebakken:
@13steinj as you can see, there are no updates. We would appreciate tests being added for this code, so feel free to add them. Thanks.

13steinj:
@lukebakken more than just that, I'd like to change the implementation slightly. 

I haven't used timers and as such would need to look into whether or not it's okay for them to just be noop, but I don't like the fact that the ioloop is a mocked object, and the connection internally starts reading the loop in a separate greenlet.

IMO the ioloop start/stop should be used, keeping a greenlet in memory and then killing it when necessary on stop (documenting that since this is using greenlets and communicating across separate threads that the IOLoop stop does not guarantee that `stop` will stop the io loop in its tracks but rather send a signal notifying the greenlet to die, and in theory in that small window another cycle of the loop may be run-- it's a race condition that can't really be avoided so long as reading the loop is on a separate green thread)

lukebakken:
@13steinj - feel free to fork this repository and merge in `cainbit:gevent_connection_adapter` or start from scratch. Open a PR from your fork & branch and we'll review it. Thanks!

vitaly-krugl:
@cainbit, @13steinj, @lukebakken: please hold off on the gevent effort. I am refactoring core/adapter/transport interfaces in pika via pull request #1002. #1002 facilitates a cleaner integration and more generic testing of adapter integrations.

vitaly-krugl:
@lukebakken, given the following:

1. Incompatibility of the callback-based API implementation in this PR with the spirit of the gevent programming model.
2. Lack of complete test suite
3. Lack of support from contributor

I am inclined to close this pull request.

michaelklishin:
Should this PR be resurrected, we would consider it as long as @vitaly-krugl's comments are addressed.

Kampii:
@13steinj I know that you are busy, but can you take a look at this? We left this issue hanging in our app and still hoping it can be fixed.

13steinj:
@Kampii I myself don't have a vested interest as I don't personally use pika right now in a non company setting, and the issue at work has been put on the backburner (because it was surprisingly easier to just use gevent + a blocking connection + reset the connection upon failure until the rewrite of the internals were done). 

I'll take a look at said rewrite when home and if it wasn't too major as to also require major changes in a custom adapter I wrote I'll open a PR, otherwise you'll have to wait until the weekend.

PR #864, opened 2017-08-24
Allow setting TCP options, such as keepalive and user-timeout

katajakasa:
This is a proposal for adding tcp keepalive and tcp-user-timeout support to pika. The idea was to make the tcp option setting generic so that it can support possibly more options in the future.

As far as possible problems go, I'm not entirely sure  how acceptable it is to set SO_KEEPALIVE for every connection since it does change behaviour. Also, compatibility on different platforms still needs work.

Even so, here's a PR. I'd like to hear thoughts on it, and whether something like this would be accepted and what needs to be done to that end ? :)

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -16,6 +16,29 @@ try:
 except AttributeError:
     SOL_TCP = 6
 
+try:
+    SOL_SOCKET = socket.SOL_SOCKET
+except:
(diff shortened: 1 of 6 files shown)


katajakasa:
Squashed, documentation updated. Ready.

lukebakken:
Thank you for the PR and tests! I have opened #880 which includes your commits. I'm going to add a couple other changes there.

PR #870, opened 2017-08-28
Cast port to int

nvllsvm:
Check if the ``port`` is an integer by casting it to one.

My motivation/desire for this change is to negate the need to cast the port value from an environment variable. The is the behavior I've grown to expect when configuring connections in Python.

### DIFF
--- a/tests/unit/connection_parameters_tests.py
+++ b/tests/unit/connection_parameters_tests.py
@@ -276,11 +276,17 @@ class ParametersTests(_ParametersTestsBase):
         params.port = 5672
         self.assertEqual(params.port, 5672)
 
-        with self.assertRaises(TypeError):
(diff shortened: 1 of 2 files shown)


tests/unit/connection_parameters_tests.py
```diff
@@ -276,11 +276,17 @@ def test_port(self):
         params.port = 5672
         self.assertEqual(params.port, 5672)
 
+        params.port = '5672'
+        self.assertEqual(params.port, 5672)
+
+        params.port = 1.5
+        self.assertEqual(params.port, 1)
```
lukebakken: This is a change from current behavior, though I'm not all that concerned about it. @michaelklishin - thoughts?

lukebakken:
Thank you!

PR #873, opened 2017-09-06
update heartbeat determin policy when tune

zjj:
if client_value > 0 then this value should be used as heartbeat.
because sometimes, the max value of client_value and the server_value
may too big for firmwall, the firmwall may have broken off the
connection.


zjj:
@michaelklishin  this pull request seems a bit ... well, but the current heartbeat policy doesn't seem good. please, check this simple patch. thx

PR #883, opened 2017-10-10
Add test showing that None is a valid socket timeout value

### LINKED ISSUE #863: Pika Socket Timeout should support >= 0 for connection
Pika enforces that socket_timeout > 0 in pika version 0.11.0 in connection.py line 479. Although this is probably a good idea, I have a use-case where I would like the socket to never timeout. I can put a large number in there, but that is unnecessary as this spec is not enforced by other AMPQ clients nor anywhere in the AMPQ spec. tocket_timeout >= 0 should work. tx

lukebakken:
Fixes #863


michaelklishin:
Wow, I stopped receiving notifications about this repo in my general GitHub feed :/

PR #886, opened 2017-10-27
Ensure that Channel.Close stops event processing and raises exception

### LINKED ISSUE #841: start_consuming() keeps running even after channel closed by exception
I'm encountering this problem when a channel exception caused the channel to close in the middle of a `start_consuming` loop, there is no way to know it as neither did `start_consuming` return, nor was an exception thrown. It basically kept looping as if it was normal, despite the entire channel is gone.

_My use case necessitates BlockingConnection; hence I can't find an alternative._

Here's the code to reproduce that. Trigger the problem by randomly writing something with a `logs.something` key to `egress` exchange. The callback will cause the broker to shutdown my channel, as the exchange I publish to is invalid.

```python
from pika import BlockingConnection, ConnectionParameters


def callback(channel, methods, props, body):
    channel.basic_publish(
        'exchange_not_exist',
        routing_key='123',
        body='Nope this is wrong'
    )


def main():
    conn = BlockingConnection(
        ConnectionParameters('localhost')
    )
    ch = conn.channel()
    ch.exchange_declare('egress', exchange_type='topic', durable=True)
    q = ch.queue_declare(exclusive=True).method.queue
    ch.queue_bind(q, 'egress', 'logs.#')
    ch.basic_consume(callback, q, True)
    print('Starts consuming.')
    ch.start_consuming()
    print('I\'m done.')


if __name__ == '__main__':
    main()
```

lukebakken:
Fixes #841 

These changes basically emulate what is done [in this code](https://github.com/pika/pika/blob/master/pika/adapters/blocking_connection.py#L1253-L1258) but also uses `_request_channel_dispatch` to ensure `start_consuming()` or other blocking channel methods resume.

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -448,7 +448,7 @@ class BlockingConnection(object):
         is_done = (lambda:
                    self._closed_result.ready or
                    (not self._impl.outbound_buffer and
-                    (not waiters or any(ready() for ready in  waiters))))
+                    (not waiters or any(ready() for ready in waiters))))
 
         # Process I/O until our completion condition is satisified
         while not is_done():
@@ -756,7 +756,6 @@ class BlockingConnection(object):
 
             # Drive I/O until Channel.Open-ok
             channel._flush_output(opened_args.is_ready)
-
 
         return channel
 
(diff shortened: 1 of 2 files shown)


lukebakken:
@streamliner18 - please give these changes a try. Thanks!

PR #887, opened 2017-10-27
Ensure process_data_events returns

### LINKED ISSUE #885: basic_cancel no longer stops basic_consume
This was fine under pika-0.10.0 but no longer working with pika-0.11.0

My message loop looks something like:
```
            self._channel.basic_qos(prefetch_count=1)
            self._consumer_tag = self._channel.basic_consume(callback,
                                                             queue=self._queue,
                                                             consumer_tag=self._consumer_tag)
            self._channel.start_consuming()
```

Stop message loop runs this:

```
self._channel.basic_cancel(consumer_tag=self._consumer_tag)
```

I use a signal to run the stop message loop
```
    signal.signal(signal.SIGTERM, handle_shutdown)  # sig num 15, terminated
    signal.signal(signal.SIGQUIT, handle_shutdown)  # sig num 3, terminal quit
    signal.signal(signal.SIGINT, handle_shutdown)  # sig num 2, terminal interrupt
```

I know it runs the cancel even with pika 11 but the cancel doesn't seem to interrupt the message loop code anymore.

UPDATE:
This seems to only affect blocking the connection.

For now I am going back to pika 0.10 which resolves the issue

lukebakken:
In certain scenarios process_data_events does not return to allow start_consuming() to return

Fixes #885 

Code to reproduce: [repro.py.txt](https://github.com/pika/pika/files/1423704/repro.py.txt)

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -1599,6 +1599,9 @@ class BlockingChannel(object):
             # If channel was closed, self._consumer_infos will be empty
             if consumer_tag in self._consumer_infos:
                 del self._consumer_infos[consumer_tag]
+                # Schedule termination of connection.process_data_events using a
+                # negative channel number
+                self.connection._request_channel_dispatch(-self.channel_number)
             raise
 
         # NOTE: Consumer could get cancelled by broker immediately after opening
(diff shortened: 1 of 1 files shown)


lukebakken:
@radzhome - if you have a chance to check out this PR, I would appreciate it. Thanks!

PR #888, opened 2017-10-31
Ensure large numbers are encoded correctly

### LINKED ISSUE #861: Python int type does not fit into struct 'i' format
https://github.com/pika/pika/blob/master/pika/data.py#L118-L123

I'm getting a "struct.error: 'i' format requires -2147483648 <= number <= 2147483647" where my value is 2592000000. That value fits into a Python int, which is `long` and on our platform is 64-bit wide, while struct 'i' format is defined as strictly 32-bit. I can work around that by explicitly casting to `long`, but it is a bug in pika nonetheless.

lukebakken:
Fixes #861

The struct module throws `struct.error` when attempting to encode a greater-than-4-byte number using the "i" (4 byte) format. This will catch the error and promote the encoding to "q" (8 byte) format.

### DIFF
--- a/pika/data.py
+++ b/pika/data.py
@@ -119,8 +119,12 @@ def encode_value(pieces, value):
         pieces.append(struct.pack('>cq', b'l', value))
         return 9
     elif isinstance(value, int):
-        pieces.append(struct.pack('>ci', b'I', value))
-        return 5
+        try:
+            pieces.append(struct.pack('>ci', b'I', value))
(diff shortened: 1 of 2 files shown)


lukebakken:
I'm not thrilled with catching `struct.error` but I expect this to be a rare occasion and it seems preferable to making this comparison every time - `-2147483648 <= value <= 2147483647`

Issue #892, opened 2017-11-11
Question re. interrupting `BlockingChannel.consume` from another thread


Bogdanp:
In [Dramatiq][dramatiq] I've got an architecture where consumers live on their own thread and call `BlockingChannel`'s `consume` method with an inactivity timeout of `1s` by default. Whenever a consumer gets a message from Rabbit, it puts it on a shared in-memory work queue where it is grabbed and processed by any of `n` worker threads. Once the worker thread is done processing the message, it gets put on an "acks" queue that belongs to the consumer. When the consumer hits the inactivity timeout it acks any pending messages off its queue and then waits for new messages from RMQ.

Acking from the worker threads is not an option since `BlockingChannel`s are not thread safe.

When you set a QoS value the fact that `consume` can't be interrupted from another thread really starts to hurt throughput, at which point you are forced to either:

1. use a tiny `inactivity_timeout` (<100ms), which "fixes" the throughput issue but is essentially a spinlock so it's inefficient,
2. switch to `SelectConnection` and implement your own interrupt, this is probably what any sane person would do, but the problem is you'd end up implementing a lot of stuff that `BlockingChannel` already implements, or
3. hack your own interrupt by hooking into implementation details of `BlockingChannel`.

The latter is what I ended up doing ([here][c1], [here][c2] and [here][c3]) and my interrupt implementation is working well, but I don't feel too good about depending on `BlockingConnection`'s implementation details so my "issue" is more of a question than anything: is there a better way I could be doing this and, if not, would this be a good addition to `BlockingConnection` (I'd be happy to contribute)? 

[dramatiq]: http://dramatiq.io/
[c1]: https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/brokers/rabbitmq.py#L359-L373
[c2]: https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/brokers/rabbitmq.py#L422-L435
[c3]: https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/brokers/rabbitmq.py#L439-L458

lukebakken:
Hi Bogdan -

Sorry it has taken a while to review how you've modified Pika to suit your needs. If you would like to submit a pull request with proposed changes (like [this suggestion](https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/brokers/rabbitmq.py#L362-L363)), I would be happy to review them.

Thanks! 🎄

zjj:
[...]

PR #902, opened 2017-12-20
add thread lock for blocking connection, thread safe

zjj:
This is an attempt to address #892.

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -19,6 +19,7 @@ import contextlib
 import functools
 import logging
 import time
+import threading
 
 import pika.channel
 from pika import compat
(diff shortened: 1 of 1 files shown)


zjj:
is the lock for _flush_output enough ???

michaelklishin:
Thread safety is a property of an algorithm (a sequence of steps). You haven't mentioned what specific scenario you are trying to make safer, so we can't know whether locking in that one function is enough.

Locking around I/O reads should be done with care. The same can be said for writes but in practice most if not all clients do it and combined with heartbeats it works OK. It can have devastating effect on throughput, as demonstrated by recent .NET client changes (there were multiple ones, both for publishers and consumers, so I'm not linking to a specific one). Were any benchmarks conducted to compare the difference for the "no concurrent access" (only one thread) scenario, which is likely to be the most common one?

zjj:
@michaelklishin  see issue #892

vitaly-krugl:
@michaelklishin  - there are far too many other code paths in pika blocking connection that are not thread safe, so a one-off solution is lackluster. 

The inactivity_timeout feature in BlockingChannel.consume can be used to easily solve the problem in application code without adding a hack in pika.

lukebakken:
@vitaly-krugl thank you for adding your input.

PR #904, opened 2017-12-21
exception shall be socket.error in _PollerBase

zjj:


zjj:
@michaelklishin  some tests are not covered.

lukebakken:
Hello!

Starting in Python 3.3, the exception raised is `OSError`. I will incorporate this fix into a new PR today. Thanks!

michaelklishin:
It would help a lot if you explained what problem we are trying to solve here. How can it be reproduced?

lukebakken:
`socket.error` was replaced by `OSError` in Python 3.3, but some instances of `socket.error` remain.

zjj:
python2 socket.err
python3 either OSError or IOError works.

PR #905, opened 2017-12-21
use STREAM sock for _get_interrupt_pair instead of DGRAM

zjj:


zjj:
this patch shall be merge after #904,  We found STREAM is much stabler than DGRAM  on Windows while large amount of msgs.

lukebakken:
Thank you! I have opened #908 which contains a similar fix.

PR #907, opened 2017-12-26
Use Python 2 / 3 compatible socket exception object

lukebakken:
Supercedes #904


lukebakken:
@michaelklishin [this commit](https://github.com/pika/pika/pull/907/commits/aff208b39a7c8126945fbb6286a2d0c464c52fc7) has the changes, and the next has changes due to YAPF formatting as required.

lukebakken:
@zjj - if you could review this, that would be great. Thanks!

PR #908, opened 2017-12-26
Replace home-grown socketpair

lukebakken:
Use the version of socketpair that comes in Python 3.5, which is also in this project:

https://github.com/mhils/backports.socketpair

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -632,22 +632,10 @@ class _PollerBase(_AbstractBase):  # pylint: disable=R0902
     def _get_interrupt_pair():
         """ Use a socketpair to be able to interrupt the ioloop if called
         from another thread. Socketpair() is not supported on some OS (Win)
-        so use a pair of simple UDP sockets instead. The sockets will be
+        so use a pair of simple TCP sockets instead. The sockets will be
         closed and garbage collected by python when the ioloop itself is.
         """
-        try:
-            read_sock, write_sock = socket.socketpair()
-
-        except AttributeError:
-            LOGGER.debug("Using custom socketpair for interrupt")
-            read_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-            read_sock.bind(('localhost', 0))
-            write_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-            write_sock.connect(read_sock.getsockname())
-
-        read_sock.setblocking(0)
-        write_sock.setblocking(0)
-        return read_sock, write_sock
+        return socket.socketpair()
 
     def _read_interrupt(self, interrupt_fd, events):  # pylint: disable=W0613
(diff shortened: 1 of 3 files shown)


lukebakken:
Tested on Windows 8.1 VM using Python 2 and RabbitMQ 3.7.2

pika/compat/__init__.py
```diff
@@ -162,3 +163,55 @@ def get_linux_version(release_str):
...
+                csock.setblocking(True)
+                ssock, _ = lsock.accept()
+            except Exception:
+                csock.close()
+                raise
+        finally:
+            lsock.close()
+        return (ssock, csock)
+
+    socket.socketpair = socketpair
```
vitaly-krugl: @lukebakken - WARNING this assignment will break application code that has its own check and workaround for missing socket.socketpair, particularly when the application needs to create a socket pair with parameters not supported by this limited implementation. 

This is why we originally contained that logic inside the internal method `_get_interrupt_pair` and didn't set attributes in the `socket` module.

gmr:
Monkey-patching socket.socketpair should be avoided. I'd just keep all socketpair definition and use for pika within the pika package.

pika/adapters/select_connection.py
```diff
@@ -632,22 +632,10 @@ def _dispatch_fd_events(self, fd_event_map):
...
-
-        except AttributeError:
-            LOGGER.debug("Using custom socketpair for interrupt")
-            read_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-            read_sock.bind(('localhost', 0))
-            write_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-            write_sock.connect(read_sock.getsockname())
-
-        read_sock.setblocking(0)
-        write_sock.setblocking(0)
```
vitaly-krugl: @lukebakken - WARNING The original code returned a non-blocking socket pair, but the pair is blocking after this PR. The change from non-blocking to blocking can lead to a deadlock. The implementation was designed for non-blocking model - see https://github.com/pika/pika/blob/2b145f4e40e4fa2dda3b5e94dba8a86fbaf83319/pika/adapters/select_connection.py#L539 and https://github.com/pika/pika/blob/2b145f4e40e4fa2dda3b5e94dba8a86fbaf83319/pika/adapters/select_connection.py#L651.

A simple/pathological case that would reproduce the deadlock is calling `stop` enough times from code executing within the event loop itself to fill the socket's transmit buffer.

A more complex - albeit still possibly pathological - case is where a background thread or many background threads call `stop`, while the event loop code is blocked waiting for something from the thread that happens to make the `stop` call that would exceed the `_w_interrupt` socket's transmit buffer.

Is there any reason for removing the logic that made the sockets non-blocking?

lukebakken: Nope, I just missed it in the change. Addressing it now.

PR #915, opened 2018-01-10
Refactor Connection._send_message in terms of _send_frame to avoid  code duplication

vitaly-krugl:
Refactor Connection._send_message in terms of Connection._send_frame to avoid  code duplication.

Also, remove default content arg default value of None since implementation assumes content is a sequence.

Renamed `method` arg name to `method_frame` to reflect the expected object type and for consistency with other methods in same module.

Fixed method comment not to refer to lock, since locking was unnecessary and was removed some time ago.

Covered by existing tests

cc @lukebakken

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -2255,37 +2255,27 @@ class Connection(object):
         else:
             self._send_frame(frame.Method(channel_number, method))
 
-    def _send_message(self, channel_number, method, content=None):
-        """Send the message directly, bypassing the single _send_frame
-        invocation by directly appending to the output buffer and flushing
-        within a lock.
+    def _send_message(self, channel_number, method_frame, content):
+        """Publish a message.
 
         :param int channel_number: The channel number for the frame
-        :param pika.amqp_object.Method method: The method frame to send
-        :param tuple content: If set, is a content frame, is tuple of
-                              properties and body.
+        :param pika.object.Method method_frame: The method frame to send
+        :param tuple content: A content frame, which is tuple of properties and
+                              body.
 
         """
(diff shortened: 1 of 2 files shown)


pika/connection.py
```diff
@@ -2199,37 +2199,27 @@ def _send_method(self, channel_number, method, content=None):
...
+        :param pika.object.Method method_frame: The method frame to send
+        :param tuple content: A content frame, which is tuple of properties and
+                              body.
 
         """
         length = len(content[1])
-        write_buffer = [frame.Method(channel_number, method).marshal(),
-                        frame.Header(channel_number, length,
-                                     content[0]).marshal()]
+        self._send_frame(frame.Method(channel_number, method_frame))
```
gmr: IIRC this approach caused problems in the past with interspersed frames due to IOLoop yielding or some such behavior? I don't remember the details, but the write buffer behavior was to make sure that multi-frame deliveries were atomic in being written.

vitaly-krugl: Hi @gmr, thanks for chiming in. I remember that too. That was before BlockingConnection was rebased on SelectConnection. That legacy blocking socket implementation used to periodically try to read incoming data in the midsts of queuing of frames which could inject frames (e.g., heartbeat ack), thus corrupting the message being sent. I fixed that as part of the big refactoring effort a few years ago.

gmr:
Looks like I created a conflict in cleaning up the unit tests. Mind rebasing the tests and resolving?

lukebakken:
Working on those conflicts now

vitaly-krugl:
Thank you.

PR #918, opened 2018-01-11
Do not monkey-patch socket, ensure socketpair returns non-blocking sockets

### LINKED ISSUE #916: Override of socket.socketpair may break some application code
cc @lukebakken 

PR #908 overrides socket.socketpair. This assignment will break application code that has its own check and workaround for missing socket.socketpair, particularly when the application needs to create a socket pair with parameters not supported by this limited implementation. See https://github.com/pika/pika/pull/908/files#r160840234

This is why we originally contained that limited socketpair logic inside the internal method `_get_interrupt_pair` and didn't set attributes in the `socket` module.

### LINKED ISSUE #917: PR #908 change in _get_interrupt_pair may cause a deadlock
The original code returned a non-blocking socket pair, but the pair is blocking after PR #908. See https://github.com/pika/pika/pull/908/files#r160844142.

The change from non-blocking to blocking can lead to a deadlock. The implementation was designed for non-blocking model - see https://github.com/pika/pika/blob/2b145f4e40e4fa2dda3b5e94dba8a86fbaf83319/pika/adapters/select_connection.py#L539 and https://github.com/pika/pika/blob/2b145f4e40e4fa2dda3b5e94dba8a86fbaf83319/pika/adapters/select_connection.py#L651.

A simple/pathological case that would reproduce the deadlock is calling `stop` enough times from code executing within the event loop itself to fill the socket's transmit buffer.

A more complex - albeit still possibly pathological - case is where a background thread or many background threads call `stop`, while the event loop code is blocked waiting for something from the thread that happens to make the `stop` call that would exceed the `_w_interrupt` socket's transmit buffer.

Is there any reason for removing the logic that made both sockets non-blocking? See https://github.com/pika/pika/pull/908/files#diff-69e7c8c211c62cdb127e8ab3324ce7c6L649

lukebakken:
Fixes #916
Fixes #917

### DIFF
--- a/pika/compat/__init__.py
+++ b/pika/compat/__init__.py
@@ -167,7 +167,12 @@ if platform.system() == 'Linux':
 _LOCALHOST = '127.0.0.1'
 _LOCALHOST_V6 = '::1'
 
-def socketpair(family=socket.AF_INET, type=socket.SOCK_STREAM, proto=0):
+def _nonblocking_socketpair(family=socket.AF_INET, type=socket.SOCK_STREAM, proto=0):
+    """
+    Returns a pair of sockets in the manner of socketpair with the additional
+    feature that they will be non-blocking. Prior to Python 3.5, socketpair
+    did not exist on Windows at all.
+    """
     if family == socket.AF_INET:
         host = _LOCALHOST
     elif family == socket.AF_INET6:
@@ -196,6 +201,10 @@ def socketpair(family=socket.AF_INET, type=socket.SOCK_STREAM, proto=0):
             raise
     finally:
         lsock.close()
+
+    # Make sockets non-blocking to prevent deadlocks
+    # See https://github.com/pika/pika/issues/917
     csock.setblocking(False)
     ssock.setblocking(False)
+
     return (ssock, csock)
--- a/pika/compat/__init__.py
+++ b/pika/compat/__init__.py
@@ -167,51 +167,35 @@ if platform.system() == 'Linux':
 _LOCALHOST = '127.0.0.1'
 _LOCALHOST_V6 = '::1'
 
-if not hasattr(socket, 'socketpair'):
-    # Origin: https://gist.github.com/4325783, by Geert Jansen.  Public domain.
-    def socketpair(family=socket.AF_INET, type=socket.SOCK_STREAM, proto=0):
-        if family == socket.AF_INET:
-            host = _LOCALHOST
-        elif family == socket.AF_INET6:
-            host = _LOCALHOST_V6
-        else:
-            raise ValueError(
-                'Only AF_INET and AF_INET6 socket address families '
-                'are supported')
-        if type != socket.SOCK_STREAM:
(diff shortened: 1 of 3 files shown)


pika/compat/__init__.py
```diff
@@ -167,51 +167,35 @@ def get_linux_version(release_str):
...
-    socket.socketpair = socketpair
+            csock.connect((addr, port))
+            ssock, _ = lsock.accept()
+        except Exception:
+            csock.close()
+            raise
+    finally:
+        lsock.close()
+    csock.setblocking(False)
+    ssock.setblocking(False)
```
vitaly-krugl: My 2 cents: this function's name is exactly like the real deal in `socket` module which returns blocking sockets; if some other part of pika or pika's user use it, they might expect that it behaves just like the one in `socket`. For this reason, I would add the logic that makes the sockets non-blocking in the specialized method `_get_interrupt_pair` augmented with a comment, such as "make sockets non-blocking to avoid potential deadlock".

lukebakken:
@vitaly-krugl thanks for reviewing our recently merged PRs 😄

pika/compat/__init__.py
```diff
@@ -167,51 +167,35 @@ def get_linux_version(release_str):
...
-                'are supported')
-        if type != socket.SOCK_STREAM:
-            raise ValueError('Only SOCK_STREAM socket type is supported')
-        if proto != 0:
-            raise ValueError('Only protocol zero is supported')
-
-        # We create a connected TCP socket. Note the trick with
-        # setblocking(False) that prevents us from having to create a thread.
-        lsock = socket.socket(family, type, proto)
+def socketpair(family=socket.AF_INET, type=socket.SOCK_STREAM, proto=0):
```
vitaly-krugl: Add a function comment that explains the reasons for having this method - Windows doesn't provide `socket.socketpair`.

pika/compat/__init__.py
```diff
@@ -167,51 +167,44 @@ def get_linux_version(release_str):
...
-                'are supported')
-        if type != socket.SOCK_STREAM:
-            raise ValueError('Only SOCK_STREAM socket type is supported')
-        if proto != 0:
-            raise ValueError('Only protocol zero is supported')
-
-        # We create a connected TCP socket. Note the trick with
-        # setblocking(False) that prevents us from having to create a thread.
-        lsock = socket.socket(family, type, proto)
+def _nonblocking_socketpair(family=socket.AF_INET, type=socket.SOCK_STREAM, proto=0):
```
vitaly-krugl: +1

vitaly-krugl:
@lukebakken - what's the story with `codecov/patch`, though?

lukebakken:
@vitaly-krugl  - I'm not sure. I think it has to do with test coverage specific to the changes in this PR. I'm not too concerned since everything else passes. What do you think?

vitaly-krugl:
@lukebakken - I am the wrong person to ask, as I am a bit pedantic about such things :). Your call.

Issue #919, opened 2018-01-12
exchange_declare() got an unexpected keyword argument 'type'


tobiasfielitz:
Tried [this example](http://pika.readthedocs.io/en/latest/examples/twisted_example.html), but getting the following error.

```
Twisted==17.9.0
pika==0.11.2
```

```
[Failure instance: Traceback: <class 'TypeError'>: exchange_declare() got an unexpected keyword argument 'type'
/usr/lib/python3.6/site-packages/twisted/internet/defer.py:653:_runCallbacks
/usr/lib/python3.6/site-packages/twisted/internet/defer.py:1442:gotResult
/usr/lib/python3.6/site-packages/twisted/internet/defer.py:1386:_inlineCallbacks
consumer.py:19:run
--- <exception caught here> ---
/usr/lib/python3.6/site-packages/pika/adapters/twisted_connection.py:169:wrapped
```

**consumer.py**

```
# -*- coding:utf-8 -*-

import pika
from pika import exceptions
from pika.adapters import twisted_connection
from twisted.internet import defer, reactor, protocol, task

connection_string = 'amqp://{}:{}@{}:5672/{}?backpressure_detection=t'.\
    format('scraper', 'password', 'rabbitmq_scraper', 'scraper')

@defer.inlineCallbacks
def run(connection):
    channel = yield connection.channel()
    exchange = yield channel.exchange_declare(exchange='topic_link',type='topic')
    queue = yield channel.queue_declare(queue='hello', auto_delete=False, exclusive=False)
    yield channel.queue_bind(exchange='topic_link', queue='hello', routing_key='hello.world')
    yield channel.basic_qos(prefetch_count=1)
    queue_object, consumer_tag = yield channel.basic_consume(queue='hello', no_ack=False)
    l = task.LoopingCall(read, queue_object)
    l.start(0.01)

@defer.inlineCallbacks
def read(queue_object):
    ch,method,properties,body = yield queue_object.get()
    if body:
        print(body)
    yield ch.basic_ack(delivery_tag=method.delivery_tag)

parameters = pika.URLParameters(connection_string)
cc = protocol.ClientCreator(reactor, twisted_connection.TwistedProtocolConnection, parameters)
d = cc.connectTCP('rabbitmq_scraper', 5672)
d.addCallback(lambda protocol: protocol.ready)
d.addCallback(run)
reactor.run()
```

tobiasfielitz:
Needs to be **exchange_type**.

lukebakken:
[...]

PR #920, opened 2018-01-12
Update twisted_example.rst

tobiasfielitz:
https://github.com/pika/pika/issues/919


lukebakken:
Thanks! I think that may be the last remaining use of `type` that doesn't have a Pika version check.

PR #923, opened 2018-01-18
issue-921 python 3.7 compatibility: async is keyword now

### LINKED ISSUE #921: Pika is not compatible with python 3.7
"async and await are not recommended to be used as variable, class, function or module names. Introduced by PEP 492 in Python 3.5, they will become proper keywords in Python 3.7." https://docs.python.org/3.7/whatsnew/3.5.html#new-keywords

So, you catch error in python 3.7:
/envs/plarin-3.7.0a4/lib/python3.7/site-packages/pika/adapters/libev_connection.py", line 106
    self.async = None
             ^
SyntaxError: invalid syntax

aolefira:
Fixes #921

### DIFF
--- a/pika/adapters/libev_connection.py
+++ b/pika/adapters/libev_connection.py
@@ -102,7 +102,7 @@ class LibevConnection(BaseConnection):
                 self.ioloop = pyev.default_loop()
                 self.ioloop.update()
 
-        self.async = None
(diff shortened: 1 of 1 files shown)


lukebakken:
Thank you! I have opened #924 which includes your commit as well as one small change.

PR #928, opened 2018-01-19
Do not disconnect on SSL_ERROR_WANT_WRITE

skizunov:
In certain cases, when using Python 3.5+, `SSLSocket.send` may
raise `ssl.SSL_ERROR_WANT_WRITE` when the socket is not currently
able to write. The previous behavior would close the connection.
The new behavior is to treat this just like an `EWOULDBLOCK`
socket error which is what this is analogous to.

Signed-off-by: Sergey Kizunov <sergey.kizunov@ni.com>

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -464,6 +464,16 @@ class BaseConnection(connection.Connection):
             self.outbound_buffer.appendleft(frame)
             self._handle_timeout()
 
+        except ssl.SSLError as error:
+            if error.args[0] == ssl.SSL_ERROR_WANT_WRITE:
+                # In Python 3.5+, SSLSocket.send raises this if the socket is
+                # not currently able to write. Handle this just like an
+                # EWOULDBLOCK socket error.
+                LOGGER.debug("Would block, requeuing frame")
+                self.outbound_buffer.appendleft(frame)
+            else:
+                return self._handle_error(error)
+
         except SOCKET_ERROR as error:
             if error.errno in (errno.EAGAIN, errno.EWOULDBLOCK):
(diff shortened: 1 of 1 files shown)


lukebakken:
Thanks. Is there any way to reproduce this?

skizunov:
@lukebakken : I saw this on Windows 7 on a lower spec'ed machine. Using Python 3.6.4, and  a`BlockingConnection` on an SSL connection, just keep writing data using `Channel.publish` in a tight loop. With the previous code, it would disconnect. You won't see the issue if you put a small `time.sleep` between calls to `Channel.publish`.

anton-ryzhov:
I've faced with this bug too. Hope this fix will be in release soon.

Reproduction is very simple — use 3.5+ and send one big message (100k in my case for real network, may be bigger for localhost). If this message will be bigger than underlying socket buffer — you'll get the error.

Also [doc](https://docs.python.org/3/library/ssl.html#notes-on-non-blocking-sockets) says that both `recv` and `send` may cause both `SSLWantReadError` or `SSLWantWriteError`, but `pika` now handles only one.

vitaly-krugl:
> Also [doc](https://docs.python.org/3/library/ssl.html#notes-on-non-blocking-sockets) says that both `recv` and `send` may cause both `SSLWantReadError` or `SSLWantWriteError`, but `pika` now handles only one.

That possibility of either `SSLWantReadError` or `SSLWantWriteError` in `recv` and `send` may be triggered by TLS renegotiation. Being addressed in the refactoring pull request #1002. I could definitely use some help writing test code to validate that it works in the renegotiation scenario.

Issue #929, opened 2018-01-24
Allow the use of TLS SNI


gbartl:
The python SSL library allows the specification of a server_hostname parameter as part of the SSLContext.wrap_socket function. Setting this parameters to a server address enables the TLS SNI extension. At the moment pika does not expose this particular SSL parameter to the user due to it using the similar ssl.wrap_socket function. 

Enabling TLS SNI is required to connect to rabbitmq using TLS when rabbit is deployed on OPENSHIFT.

lukebakken:
We would be happy to accept a pull request with this change. Thanks!

lukebakken:
Resolved by #932

PR #932, opened 2018-01-29
SSL advanced parameters

gbartl:
PR for #929 
Added class for defining advanced SSL options which can be optionally used instead of the normal SSL options dict. This allows advanced config using the ssl.SSLContext.wrap_socket() instead of the more basic but simpler ssl.wrap_socket().


gbartl:
Closed untill I can sort out python 2.6 compat issues

lukebakken:
@gbartl - don't worry so much about 2.6. We need to drop support for that ancient version anyway.

PR #933, opened 2018-01-30
Drop support for EOL Python 2.6 and 3.3

hugovk:
https://github.com/pika/pika/pull/932#issuecomment-361425622 says:

> don't worry so much about 2.6. We need to drop support for that ancient version anyway. If anything, detect 2.6 and disable the changes you're making.

Here's the pip installs for pika from PyPI for last month:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |   75.0% |        202,328 |
| 3.6            |   11.1% |         30,057 |
| 3.5            |    8.6% |         23,204 |
| 3.4            |    4.8% |         13,008 |
| 2.6            |    0.2% |            663 |
| 3.3            |    0.1% |            159 |
| 3.7            |    0.0% |            133 |
| 3.2            |    0.0% |             50 |
| None           |    0.0% |              5 |

Source: `pypinfo --start-date -60 --end-date -30 --percent --pip --markdown pika pyversion`

Python 3.3 is also EOL and even less used, so let's drop that too.

It also does some minor code cleanup.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,7 +9,7 @@ Introduction
 Pika is a pure-Python implementation of the AMQP 0-9-1 protocol including RabbitMQ's
 extensions.
 
-- Python 2.6+ and 3.3+ are supported.
+- Python 2.7 and 3.3+ are supported.
 
 - Since threads aren't appropriate to every situation, it doesn't
   require threads. It takes care not to forbid them, either. The same
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@ Pika
 ====
 Pika is a RabbitMQ (AMQP-0-9-1) client library for Python.
 
-|Version| |Status| |Coverage| |License| |Docs|
+|Version| |Python versions| |Status| |Coverage| |License| |Docs|
 
 Introduction
 -------------
@@ -83,6 +83,9 @@ with ``google`` style prior to issuing your pull request.
 
 .. |Version| image:: https://img.shields.io/pypi/v/pika.svg?
    :target: http://badge.fury.io/py/pika
+
+.. |Python versions| image:: https://img.shields.io/pypi/pyversions/pika.svg
+    :target: https://pypi.python.org/pypi/pika
 
 .. |Status| image:: https://img.shields.io/travis/pika/pika.svg?
    :target: https://travis-ci.org/pika/pika
--- a/README.rst
+++ b/README.rst
@@ -9,7 +9,7 @@ Introduction
 Pika is a pure-Python implementation of the AMQP 0-9-1 protocol including RabbitMQ's
 extensions.
 
-- Python 2.7 and 3.3+ are supported.
+- Python 2.7 and 3.4+ are supported.
 
 - Since threads aren't appropriate to every situation, it doesn't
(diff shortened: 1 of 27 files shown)


hugovk:
When merged, the 2.6 skips in https://github.com/pika/pika/pull/932 can be removed.

michaelklishin:
@lukebakken @gmr fine with you to merge?

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: python
 
 python:
-  - 2.6
   - 2.7
   - 3.3
```
lukebakken: Remove version `3.3`

examples/producer.py
```diff
@@ -5,7 +5,7 @@
 import json
 import random
 
-print(('pika version: %s') % pika.__version__)
```
lukebakken: Could these changes be part of a separate PR? They aren't directly related to dropping `2.6` and `3.3` support, right?

pika/adapters/blocking_connection.py
```diff
@@ -874,7 +874,7 @@ class _ConsumerCancellationEvt(_ChannelPendingEvt):
     `Basic.Cancel`
     """
 
-    __slots__ = ('method_frame')
```
lukebakken: See my comment about making these changes in a separate PR

utils/codegen.py
```diff
@@ -121,22 +121,22 @@ def genSingleDecode(prefix, cLvalue, unresolved_domain):
...
-                  cLvalue))
+            print(Exception(prefix + "(%s, offset) = data.decode_table(encoded, offset)" %
+                            cLvalue))
         else:
             raise Exception("Illegal domain in genSingleDecode", type)
 
     def genSingleEncode(prefix, cValue, unresolved_domain):
         type = spec.resolveDomain(unresolved_domain)
         if type == 'shortstr':
-            print(prefix + \
```
lukebakken: Please use a separate PR for formatting changes.

gmr:
There's a lot of work in this PR that I'd say misattributes the changes to removal of 2.6 support. Code formatting and style changes aren't the same as removing 2.6.

I don't disagree with the code formatting changes, but I'd rather 
separate concerns TBH.

gmr:
That being said I'm a big +1 for dropping 2.6 support as long as we're very clear in versioning that it is a breaking change for anyone who's stuck in Python pre 2013. (2.6 EOL was 2013).

.travis.yml
hugovk: Will do!

hugovk:
Some of the changes are only possible after dropping 2.6 (formatters/set/dict), but sure, I can move all those to another PR!

hugovk:
This PR updated to only include dropping the EOL versions. Will create another for the others.

The coverage build job fails because:
```
0.54s$ aws s3 cp --recursive s3://com-gavinroy-travis/pika/$TRAVIS_BUILD_NUMBER/ coverage
fatal error: Unable to locate credentials
```

This is unrelated, and should be disabled from PRs and forks.

Issue #158, opened 2012-07-18
Unable to pass user data to callbacks


tpot:
Hello.  I'm trying to declare and bind multiple queues to an exchange asynchronously, but can't figure out a non-racy way to pass a parameter to the queue_declare callback.  )-:   A better way would be to have some way to pass user data to the callback functions as part of the API.  Can pika do this or Is there something basic I'm missing about the interface?

I ended up using functools.partial to do this, which only works with the git version, not 0.9.5 due to how it checks whether callbacks are callable or not.

import functools

channel.queue_declare(
  callback = functools.partial(on_queue_declare, user_data),
  queue = "whatever",
  auto_delete = True
)

def on_queue_declare(host, frame):
  # Do my funky thing

art049:
That's still an issue

lukebakken:
Passing user data would also address #898

lukebakken:
Both `functools.partial` and a closure are good solutions to this issue.

lukebakken:
Re-opening. I noticed that [this library](https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/client.py#L398-L401) supports this feature. Pika should too.

lukebakken:
@vitaly-krugl has suggested that `functools.partial` be the accepted way to do this [here](https://github.com/pika/pika/pull/941#issuecomment-363241531) and I agree. I have updated some of the code in `examples/` to demonstrate this.

vitaly-krugl:
`lambda` is another way, in addition to `functools.partial`, to pass extra args to your own callback. Finally, in OOP, the callback will often be an instance method, with the instance already having the necessary context without additional args.

PR #941, opened 2018-02-01
Consistently validate callback and nowait parameters

### LINKED ISSUE #925: basic_cancel default method parameters will result in an exception.
See [this comment](https://github.com/pika/pika/issues/817#issuecomment-358736838) in issue #817

lukebakken:
These are pretty big pre-`1.0.0` changes to the API, with the thought of adding a feature for passing user-defined data to callbacks when they are called (#158).

Other event-loop driven frameworks / environments (I'm thinking of javascript / node, especially) always pass the callback as the last parameter to a method that takes a callback.

This PR makes the following changes:

* Moves the callback parameter for all methods that take one to the end of the method that requires a callback
* Changes the parameter name to `callback` in all cases
* Adds validation for `callback` and `nowait` if both are part of the method signature. If `nowait=False` (the default), then a callback must be passed and `ValueError` will be raised if one isn't. #925 raised the issue that the default method parameters will result in `ValueError` and that behavior will remain the same.

To address #158, we could add a `userdata` parameter after `callback`, *or*, if callback is passed as a two-tuple, the first element could be validated as the callback method, and the second the userdata.


vitaly-krugl:
@lukebakken, before I get into the review, I just wanted to note that twisted and tornado both allow args and kwargs to be passed along with callback; e.g., `call_later(self, delay, callback, *args, **kwargs`.

Also, pika's use of the arg name `deadline` in timer APIs is confusing (at least to me). Something like `time.time() + 10` (i.e., seconds from epoch) is what I think of as deadline. The values that we pass to `add_timeout` are better named as `delay` or `seconds_from_now` (IMHO :).

lukebakken:
@vitaly-krugl thanks for pointing out how those libs use `*args, **kwargs` - I assume that is what you'd like to see this project implement?

Please feel free to rename other parameters you think are confusing and push to the `pika-925` branch.

vitaly-krugl:
@lukebakken, regarding args/kwargs - I was just brainstorming. Since Python has `functools.partial`, the other school of thought might say that it's easy enough for clients to to use it and there is no need for pika to deal with callback args/kwargs.

vitaly-krugl:
`add_timeout(self, deadline, callback_method)`: `callback_method` needs to be renamed `callback`

vitaly-krugl:
@lukebakken - please refrain from pushing commits to this PR until I am done reviewing. Otherwise it's a little like trying to hit a moving target :).

pika/channel.py
```diff
@@ -656,62 +673,67 @@ def exchange_declare(self,
...
         """Turn Channel flow control off and on. Pass a callback to be notified
         of the response from the server. active is a bool. Callback should
         expect a bool in response indicating channel flow state. For more
         information, please reference:
 
         http://www.rabbitmq.com/amqp-0-9-1-reference.html#channel.flow
 
-        :param callable callback: The callback to call upon completion
         :param bool active: Turn flow on or off
+        :param callable callback: The callback to call upon completion
```
vitaly-krugl: document that callback is optional and may be None

vitaly-krugl:
`docs/examples/tornado_consumer.rst:            self._channel.queue_declare(self.on_queue_declareok, queue_name)`

tests/unit/channel_tests.py
```diff
@@ -885,25 +912,25 @@ def test_queue_purge_rpc_request(self, rpc, _unused):
...
         mock_callback = mock.Mock()
-        self.obj.queue_purge(mock_callback, 'foo', nowait=True)
+        self.obj.queue_purge('foo', nowait=True, callback=None)
         rpc.assert_called_once_with(
-            spec.Queue.Purge(0, 'foo'), mock_callback, [])
+            spec.Queue.Purge(0, 'foo'), None, [])
 
     def test_queue_unbind_raises_channel_closed(self):
         self.assertRaises(exceptions.ChannelClosed, self.obj.queue_unbind,
                           None, 'foo', 'bar', 'baz')
```
vitaly-krugl: The None maps to the old callback arg and should be moved to the end and use kwarg

tests/acceptance/async_adapter_tests.py
```diff
@@ -84,22 +87,22 @@ class TestConsumeCancel(AsyncTestCase, AsyncAdapters):
 
     def begin(self, channel):
         self.queue_name = self.__class__.__name__ + ':' + uuid.uuid1().hex
-        channel.queue_declare(self.on_queue_declared, queue=self.queue_name)
+        channel.queue_declare(callback=self.on_queue_declared, queue=self.queue_name)
```
vitaly-krugl: Move `callback=self.on_queue_declared` to the end to match the new arg order convention (even if kwarg works around the ordering issue) ? Elsewhere in this file too. Test code snippets get copied to apps and other tests.

tests/acceptance/async_adapter_tests.py
```diff
@@ -84,22 +87,22 @@ class TestConsumeCancel(AsyncTestCase, AsyncAdapters):
...
 
     def on_queue_declared(self, frame):
         for i in range(0, 100):
             msg_body = '{}:{}:{}'.format(self.__class__.__name__, i,
                                          time.time())
             self.channel.basic_publish('', self.queue_name, msg_body)
-        self.ctag = self.channel.basic_consume(self.on_message,
+        self.ctag = self.channel.basic_consume(callback=self.on_message,
```
vitaly-krugl: callback at the end?

pika/adapters/blocking_connection.py
```diff
@@ -1496,12 +1496,12 @@ def add_on_return_callback(self, callback):
...
 
     def basic_consume(self,
-                      consumer_callback,
                       queue,
                       no_ack=False,
                       exclusive=False,
                       consumer_tag=None,
-                      arguments=None):
+                      arguments=None,
+                      callback=None):
```
vitaly-krugl: `callback=None` in the function signature suggests that `callback` is optional. However, the code and semantics dictate that it's required. This is likely confusing to the API user.

pika/channel.py
```diff
@@ -261,17 +259,20 @@ def basic_cancel(self, callback=None, consumer_tag='', nowait=False):
...
 
-    def basic_consume(self, consumer_callback,
+    def basic_consume(self,
                       queue='',
                       no_ack=False,
                       exclusive=False,
                       consumer_tag=None,
-                      arguments=None):
+                      arguments=None,
+                      callback=None):
```
vitaly-krugl: `callback=None` in the function signature suggests that `callback` is optional. However, the code and semantics dictate that it's required. This is likely confusing to the API user.

pika/channel.py
```diff
@@ -332,11 +336,11 @@ def _generate_consumer_tag(self):
...
         :returns: consumer tag
         :rtype: str
+
         """
-        return 'ctag%i.%s' % (self.channel_number,
-                              uuid.uuid4().hex)
+        return 'ctag%i.%s' % (self.channel_number, uuid.uuid4().hex)
 
-    def basic_get(self, callback=None, queue='', no_ack=False):
+    def basic_get(self, queue='', no_ack=False, callback=None):
```
vitaly-krugl: `callback=None` in the function signature suggests that `callback` is optional. However, the semantics dictate that it's required as it doesn't make sense to call without one.. This is likely confusing to the API user.

vitaly-krugl:
@lukebakken, here is some food for though: in methods where the value of the `callback` arg is correlated with the value of `nowait`, such as in `Channel.basic_cancel`, the `nowait` arg is redundant. If callback is None, then you can surmise that `nowait` should be True and if callback is not None, then it implies that `nowait` is False.

So, let's get rid of the `nowait` arg in all those cases. This will also reduce confusion and validation logic, and make the code more readable.

pika/channel.py
```diff
@@ -536,44 +542,52 @@ def close(self, reply_code=0, reply_text="Normal shutdown"):
...
         self._set_state(self.CLOSING)
 
-        self._rpc(spec.Channel.Close(reply_code, reply_text, 0, 0),
-                  self._on_closeok, [spec.Channel.CloseOk])
+        self._rpc(
+            spec.Channel.Close(reply_code, reply_text, 0, 0), self._on_closeok,
+            [spec.Channel.CloseOk])
 
-    def confirm_delivery(self, callback=None, nowait=False):
+    def confirm_delivery(self, nowait=False, callback=None):
```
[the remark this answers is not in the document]
vitaly-krugl: I think a different name for the callback arg would make the API less confusing, so the user won't think the callback is for signaling acknowledgment of `Confirm.Select`. Perhaps something like `delivery_conf_cb`?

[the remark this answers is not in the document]
vitaly-krugl: The other thing that's unusual (defective) about `Channel.confirm_delivery` is that, unlike most APIs with nowait=False capability, it doesn't support the passing of a completion callback . Thus, `BlockingChannel.confirm_delivery` needs to work around this deficiency and dabble in some internal knowhow by using the lower-level API `Channel.add_callback`. What's the benefit of having `confirm_delivery` be inconsistent with all other synchronous APIs?

pika/channel.py
```diff
@@ -536,44 +542,52 @@ def close(self, reply_code=0, reply_text="Normal shutdown"):
...
         """
-        self._validate_channel_and_callback(callback)
+        if not is_callable(callback):
+            # NB: confirm_deliver requires a callback; it's meaningless
+            # without a user callback to receieve Basic.Ack/Basic.Nack notifications
+            raise ValueError('confirm_delivery requires a callback')
 
-        # TODO confirm_deliver should require a callback; it's meaningless
-        # without a user callback to receieve Basic.Ack/Basic.Nack notifications
+        self._validate_channel_and_callback(self._on_selectok)
```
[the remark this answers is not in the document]
lukebakken: Sounds good

pika/channel.py
```diff
@@ -261,17 +254,20 @@ def basic_cancel(self, callback=None, consumer_tag='', nowait=False):
...
-        self._rpc(spec.Basic.Cancel(consumer_tag=consumer_tag, nowait=nowait),
-                  self._on_cancelok if not nowait else None,
-                  [(spec.Basic.CancelOk, {'consumer_tag': consumer_tag})] if
-                  nowait is False else [])
+        self._rpc(
+            spec.Basic.Cancel(consumer_tag=consumer_tag, nowait=nowait),
+            self._on_cancelok if not nowait else None, [(spec.Basic.CancelOk, {
+                'consumer_tag':
+                consumer_tag
+            })] if nowait is False else [])
```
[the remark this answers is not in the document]
lukebakken: Ah I thought I had reverted those. I will simplify that.

pika/adapters/blocking_connection.py
lukebakken: Thanks for the feedback, I was wondering about that.

pika/channel.py
```diff
@@ -1352,17 +1397,45 @@ def _on_unexpected_frame(self, frame_value):
...
         if callback is not None and not is_callable(callback):
             raise ValueError('callback must be a function or method')
 
+    def _validate_callback_and_nowait(self, callback, nowait):
+        """Verify callback is callable if not None and that nowait is
+        consistent
+
+        :raises ValueError: if nowait is True and callback is not None,
+                            or if nowait is False and callback is not None
+                            and is not callable
```
vitaly-krugl: @lukebakken - "or if nowait is False and callback is not None and is not callable" does not describe what the code is actually doing.

The code actually does this: "or if nowait is False and callback is either None or is not callable: 
```
if callback is None or not is_callable(callback):
    raise ValueError(...
```

The current description is what the original code (before this PR) allowed, but the code in this method doesn't match that behavior. See also my [PR comment about consolidating `nowait` and `callback`](https://github.com/pika/pika/pull/941#issuecomment-363347436) if the current implementation's tight coupling is intended.

pika/channel.py
```diff
@@ -536,44 +542,52 @@ def close(self, reply_code=0, reply_text="Normal shutdown"):
...
+        :param callable callback: Required callback for delivery
+            confirmations that has the following signature:
+            callback(pika.frame.Method), where method_frame contains
+            either method `spec.Basic.Ack` or `spec.Basic.Nack`.
+        :raises ValueError:
 
         """
-        self._validate_channel_and_callback(callback)
+        if not is_callable(callback):
+            # NB: confirm_deliver requires a callback; it's meaningless
```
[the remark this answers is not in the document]
lukebakken: A habit of mine - https://en.wikipedia.org/wiki/Nota_bene

vitaly-krugl:
@lukebakken - regarding indentation and statically-detected errors: please run pylint on all of the modified code. It's really good at finding some bugs and reporting formatting convention violations. pika already provides a pylintrc at the root level that is configured per our conventions.

On my mac, I install pylint via `pip install pylint --user`. Then run it like this from within the project's directory (if you run it inside our tree, it automatically discovers our pylintrc): `pylint pika/channel.py`.

That said, in a functional PR like this one, please fix formatting errors that are introduced by your changes, not the legacy formatting.

lukebakken:
@vitaly-krugl thanks for the info. Should I modify the two documents that mention `yapf`?

vitaly-krugl:
@lukebakken: we should really stick to PEP-0008. I am surprised that yapf snuck into pika in a non-PEP8-compatible way. This also conflicts with the pylintrc settings that had been used to validate majority of the code. The vast majority of the code was PEP8 compatible. 

Regarding indentation, see https://www.python.org/dev/peps/pep-0008/#indentation.

vitaly-krugl:
Other than fixing pylint violations found in the new changes within this PR, let's schedule a separate follow-on PR to clean up pylint findings in the body of code, as presence of existing pylint warnings/etc. makes it difficult to notice newly-introduced violations in pylint output (same idea as resolving compiler warnings in existing code to make it possible to spot new ones easily).

In some cases, pylint has false positives, so we will need to suppress those in the affected code sections with pylint pragmas.

pika/channel.py
```diff
@@ -1352,16 +1385,46 @@ def _on_unexpected_frame(self, frame_value):
...
         """
         if not self.is_open:
             raise exceptions.ChannelClosed()
-        if callback is not None and not is_callable(callback):
-            raise ValueError('callback must be a function or method')
+
+    def _require_callback(self, callback):
+        """Require that callback is callable and is not None
+
+        :raises: ValueError
```
vitaly-krugl: By the way, would `TypeError` be more appropriate? I think it might be, because we're looking for callback to be of callable type?

tests/unit/channel_tests.py
```diff
@@ -687,25 +712,25 @@ def test_exchange_delete_rpc_request(self, rpc, _unused):
...
         mock_callback = mock.Mock()
-        self.obj.exchange_delete(mock_callback, 'foo', nowait=True)
+        self.obj.exchange_delete('foo', nowait=True, callback=None)
         rpc.assert_called_once_with(
-            spec.Exchange.Delete(0, 'foo'), mock_callback, [])
+            spec.Exchange.Delete(0, 'foo'), None, [])
 
     def test_exchange_unbind_raises_channel_closed(self):
         self.assertRaises(exceptions.ChannelClosed, self.obj.exchange_unbind,
                           None, 'foo', 'bar', 'baz')
```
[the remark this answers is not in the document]
lukebakken: The test is checking `_rpc`, and the last two arguments are `callback` and `acceptable_replies`, in that order. I can add the keyword args though

pika/channel.py
```diff
@@ -262,16 +260,16 @@ def basic_cancel(self, callback=None, consumer_tag='', nowait=False):
         self._cancelled.add(consumer_tag)
 
         self._rpc(spec.Basic.Cancel(consumer_tag=consumer_tag, nowait=nowait),
-                  self._on_cancelok if not nowait else None,
-                  [(spec.Basic.CancelOk, {'consumer_tag': consumer_tag})] if
-                  nowait is False else [])
+            self._on_cancelok if not nowait else None,
+            [(spec.Basic.CancelOk, { 'consumer_tag': consumer_tag })] if nowait is False else [])
```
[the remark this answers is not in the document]
lukebakken: Yeah I'm trying to revert any of the formatting changes I've done, I'll get this one too.

pika/channel.py
```diff
@@ -253,20 +248,18 @@ def basic_cancel(self, callback=None, consumer_tag='', nowait=False):
             # will persist until the channel is closed.
             del self._consumers[consumer_tag]
 
-        if callback is not None:
-            if nowait:
-                raise ValueError('Cannot pass a callback if nowait is True')
+        if callback:
```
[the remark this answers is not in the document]
lukebakken: In this case `_validate_callback` has made sure that it's callable or `None` which should evaluate correctly here. I can always change it to `if callable(callback)` too

examples/asynchronous_consumer_example.py
```diff
@@ -48,9 +49,11 @@ def connect(self):
...
         """
         LOGGER.info('Connecting to %s', self._url)
-        return pika.SelectConnection(pika.URLParameters(self._url),
-                                        self.on_connection_open,
-                                        stop_ioloop_on_close=False)
+        return pika.SelectConnection(parameters=pika.URLParameters(self._url),
+                                     on_open_callback=self.on_connection_open,
+                                     on_open_error_callback=self.on_connection_open_error,
+                                     on_close_callback=self.on_connection_closed,
+                                     stop_ioloop_on_close=False)
```
vitaly-krugl: The new call signature for `pika.connection.channel` is `def channel(self, channel_number=None, callback=None)`. However, the `open_channel` function in the asynchronous consumer example calls it using an incorrect signature: `self._connection.channel(on_open_callback=self.on_channel_open)`

vitaly-krugl: There are a number of others like it:
```
$ git grep 'on_open_callback='
docs/examples/asynchronous_consumer_example.rst:            self._connection.channel(on_open_callback=self.on_channel_open)
docs/examples/asynchronous_publisher_example.rst:            self._connection.channel(on_open_callback=self.on_channel_open)
docs/examples/asyncio_consumer.rst:            self._connection.channel(on_open_callback=self.on_channel_open)
docs/examples/tornado_consumer.rst:            self._connection.channel(on_open_callback=self.on_channel_open)
examples/asynchronous_consumer_example.py:        self._connection.channel(on_open_callback=self.on_channel_open)
examples/asynchronous_publisher_example.py:        self._connection.channel(on_open_callback=self.on_channel_open)
examples/confirmation.py:                                   on_open_callback=on_open)
```

vitaly-krugl: in docs/examples/comparing_publishing_sync_async.rst, asynchronous `connection.channel` call doesn't match the new `Connection.channel` signature. it's called simply as `connection.channel(on_channel_open)`, which ends up passing the callback as first arg, which is now `channel_number`.

pika/adapters/blocking_connection.py
```diff
@@ -1496,8 +1500,8 @@ def add_on_return_callback(self, callback):
                         callback, self, method, properties, body))))
 
     def basic_consume(self,
-                      consumer_callback,
-                      queue,
+                      callback,
+                      queue='',
```
vitaly-krugl: Why is queue param defaulted to `''` in this commit? The queue name is not optional, right?

pika/adapters/blocking_connection.py
```diff
@@ -2010,7 +2013,7 @@ def basic_nack(self, delivery_tag=None, multiple=False, requeue=True):
                               requeue=requeue)
         self._flush_output()
 
-    def basic_get(self, queue=None, no_ack=False):
+    def basic_get(self, queue='', no_ack=False):
```
vitaly-krugl: What's the point of default value on the `queue` arg? Queue name is required, isn't it?

pika/channel.py
```diff
@@ -332,11 +324,12 @@ def _generate_consumer_tag(self):
 
         :returns: consumer tag
         :rtype: str
+
         """
         return 'ctag%i.%s' % (self.channel_number,
                               uuid.uuid4().hex)
 
-    def basic_get(self, callback=None, queue='', no_ack=False):
+    def basic_get(self, callback, queue='', no_ack=False):
```
vitaly-krugl: `queue` shouldn't have a default value either, since you always need to pass a queue name, right?

pika/channel.py
```diff
@@ -253,20 +246,19 @@ def basic_cancel(self, callback=None, consumer_tag='', nowait=False):
...
                   self._on_cancelok if not nowait else None,
-                  [(spec.Basic.CancelOk, {'consumer_tag': consumer_tag})] if
-                  nowait is False else [])
+                  [(spec.Basic.CancelOk, { 'consumer_tag': consumer_tag })] if
+                  not nowait else [])
 
-    def basic_consume(self, consumer_callback,
+    def basic_consume(self,
+                      callback,
                       queue='',
```
vitaly-krugl: `queue` shouldn't have a default value either, since you always need to pass the name of the queue to be consued, right?

pika/channel.py
```diff
@@ -753,127 +756,132 @@ def open(self):
...
             routing_key = queue
         return self._rpc(spec.Queue.Bind(0, queue, exchange, routing_key,
                                          nowait, arguments or dict()),
-                         callback, replies)
+                         callback,
+                         [spec.Queue.BindOk] if not nowait else [])
 
-    def queue_declare(self, callback,
+    def queue_declare(self,
                       queue='',
```
vitaly-krugl: Just curious -- how common of usage is it to have the queue auto-named by RabbitMQ?

pika/channel.py
```diff
@@ -753,127 +756,132 @@ def open(self):
...
+        replies = [condition] if not nowait else []
+
         return self._rpc(spec.Queue.Declare(0, queue, passive, durable,
                                             exclusive, auto_delete, nowait,
                                             arguments or dict()),
                          callback, replies)
 
     def queue_delete(self,
-                     callback=None,
                      queue='',
```
vitaly-krugl: `queue` shouldn't have a default value here, since you always need to name a queue that you're deleting.

pika/channel.py
```diff
@@ -753,127 +756,132 @@ def open(self):
...
-        self._validate_channel_and_callback(callback)
+        self._validate_channel()
+        nowait = self._validate_rpc_completion_callback(callback)
+        replies = [spec.Queue.DeleteOk] if not nowait else []
         return self._rpc(spec.Queue.Delete(0, queue, if_unused, if_empty,
                                            nowait),
                          callback, replies)
 
-    def queue_purge(self, callback=None, queue='', nowait=False):
+    def queue_purge(self, queue='', callback=None):
```
vitaly-krugl: `queue` shouldn't have a default value since you always have to name the queue to be purged

pika/channel.py
```diff
@@ -753,127 +756,132 @@ def open(self):
...
-        replies = [spec.Queue.PurgeOk] if nowait is False else []
-        self._validate_channel_and_callback(callback)
+        self._validate_channel()
+        nowait = self._validate_rpc_completion_callback(callback)
+        replies = [spec.Queue.PurgeOk] if not nowait else []
         return self._rpc(spec.Queue.Purge(0, queue, nowait), callback, replies)
 
     def queue_unbind(self,
-                     callback=None,
                      queue='',
```
vitaly-krugl: `queue` shouldn't have a default value since you always have to name the queue to be unbound

pika/channel.py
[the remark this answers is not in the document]
vitaly-krugl: Even if it's callable, the callback could be some strange user-supplied callable object, where the user overrode `__bool__` to do something funky, so `if callback` could evaluate to false even when callback is not None. The only safe thing to do is `if callback is not None:`.

pika/connection.py
```diff
@@ -2176,12 +2187,12 @@ def _rpc(self, channel_number, method,
             raise TypeError('acceptable_replies should be list or None')
 
         # Validate the callback is callable
-        if callback_method:
-            if not utils.is_callable(callback_method):
+        if callback:
```
vitaly-krugl: Even if it's callable, the callback could be some strange user-supplied callable object, where the user overrode `__bool__` to do something funky, so `if callback` could evaluate to false even when callback is not None. The only safe thing to do is `if callback is not None:`.

pika/channel.py
```diff
@@ -1352,16 +1367,43 @@ def _on_unexpected_frame(self, frame_value):
...
+
+    def _require_callback(self, callback):
+        """Require that callback is callable and is not None
+
+        :raises: TypeError
+
+        """
+        if not callable(callback):
+            raise TypeError(
+                'Callback must be callable, but got %r', (callback,))
```
vitaly-krugl: Sorry, my mistake - that should be a % sign instead of comma after the string: `'Callback must be callable, but got %r' % (callback,)`

pika/channel.py
lukebakken: RabbitMQ "remembers" and uses the last auto-named queue used for a channel if you pass the empty string for an operation.

pika/channel.py
lukebakken: I don't know how common an auto-named queue is, but it is supported and I suspect is why queue defaults to the empty string in the code. I personally think that may be more confusing to a new user than having them pass the empty string explicitly if they want an auto-named queue. I'll see what the `bunny` library does.

pika/channel.py
vitaly-krugl: Relying on it seems like a bug waiting to happen :). I wonder if, from the client API design point of view, it wouldn't be cleaner to get rid of these defaults? I don't know the answer, just wondering.

pika/channel.py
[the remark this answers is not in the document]
lukebakken: Well that's interesting, thanks!

examples/asynchronous_consumer_example.py
lukebakken: I left the callback named as `on_open_callback` so its unambiguous what it does. I'll make sure the docs / examples are all in sync.

pika/adapters/blocking_connection.py
lukebakken: I wanted to make sure all `queue` args were the same and defaulted to the empty string, which is supported by RabbitMQ. Basically, the empty string means "use the last auto-named queue". I don't particularly care for this default, to be honest.

pika/channel.py
lukebakken: `bunny` does not use the empty string as the default, which I like better. @vitaly-krugl what do you think?

vitaly-krugl:
I don't like the empty string default. I think it's error-prone. I would
get rid of the default value on queue.

On Feb 7, 2018 7:29 AM, "Luke Bakken" <notifications@github.com> wrote:

*@lukebakken* commented on this pull request.
------------------------------

In pika/channel.py
<https://github.com/pika/pika/pull/941#discussion_r166653242>:

>
-    def queue_declare(self, callback,
+    def queue_declare(self,
                       queue='',

bunny does not use the empty string as the default, which I like better.
@vitaly-krugl <https://github.com/vitaly-krugl> what do you think?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/pika/pika/pull/941#discussion_r166653242>, or mute the
thread

examples/asynchronous_consumer_example.py
lukebakken: This should be all set now. I kept the name `on_open_callback` for both connections and channels to be consistent.

lukebakken:
@vitaly-krugl - This should be ready for a final round of review. I won't push anything else to this branch until I hear back. After that, I will update the changelog with details about what has changed in the API.

pika/channel.py
```diff
@@ -282,27 +274,27 @@ def basic_consume(self, consumer_callback,
         http://www.rabbitmq.com/confirms.html
         http://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.consume
 
-
-        :param callable consumer_callback: The function to call when consuming
-            with the signature consumer_callback(channel, method, properties,
-                                                 body), where
+        :param queue: The queue to consume from. Use the empty string to specify the
+                      most recent server-named queue for this channel.
```
vitaly-krugl: The last sentence about empty string is a proprietary feature of RabbitMQ, right? If so, need to make that clear or remove the sentence.

lukebakken:
> Did examples and docs not need to be changed as the result of arg reorder in basic_consume?

I will re-review both, as well as get the change log next.

pika/channel.py
```diff
@@ -753,127 +756,134 @@ def open(self):
...
             routing_key = queue
         return self._rpc(spec.Queue.Bind(0, queue, exchange, routing_key,
                                          nowait, arguments or dict()),
-                         callback, replies)
+                         callback,
+                         [spec.Queue.BindOk] if not nowait else [])
 
-    def queue_declare(self, callback,
-                      queue='',
+    def queue_declare(self,
```
vitaly-krugl: docs/examples/tornado_consumer.rst has an example with incorrectly ordered args that won't work: `self._channel.queue_declare(self.on_queue_declareok, queue_name)`

vitaly-krugl:
@lukebakken, I am planning to give it a final once-over on Thursday. Thanks!

pika/channel.py
vitaly-krugl: Are you planning to remove that sentence or add a reference to RabbitMQ? Ditto in `basic_get` and wherever else this is mentioned.

vitaly-krugl: Sorry, never mind - it appears that this is a feature of the AMQP 0.9.1 specification, so not specific to RabbitMQ

pika/channel.py
```diff
@@ -539,41 +540,45 @@ def close(self, reply_code=0, reply_text="Normal shutdown"):
...
+        self.callbacks.add(self.channel_number, spec.Basic.Ack, ack_nack_callback,
+                           False)
+        self.callbacks.add(self.channel_number, spec.Basic.Nack, ack_nack_callback,
+                           False)
 
-        # Send the RPC command
         self._rpc(spec.Confirm.Select(nowait),
-                  self._on_selectok if not nowait else None,
-                  [spec.Confirm.SelectOk] if nowait is False else [])
+                  callback if not nowait else None,
```
vitaly-krugl: Since callback being None or not None now determines the value of `nowait`, technically `callback if not nowait else None` can be reduced to just `callback`

tests/acceptance/async_adapter_tests.py
```diff
@@ -33,8 +33,11 @@ class TestConfirmSelect(AsyncTestCase, AsyncAdapters):
     DESCRIPTION = "Receive confirmation of Confirm.Select"
 
     def begin(self, channel):
-        channel._on_selectok = self.on_complete
-        channel.confirm_delivery()
+        channel.confirm_delivery(ack_nack_callback=self.ack_nack_callback,
+                                 callback=self.on_complete)
+
+    def ack_nack_callback(frame):
+        pass
```
vitaly-krugl: Did you intend this to do something inside ack_nack_callback?

pika/adapters/blocking_connection.py
```diff
@@ -1139,20 +1142,20 @@ def __init__(self, channel_impl, connection):
         self._basic_getempty_result = _CallbackResult(
             self._MethodFrameCallbackResultArgs)
 
-        self._impl.add_on_cancel_callback(self._on_consumer_cancelled_by_broker)
+        self._impl.add_on_cancel_callback(callback=self._on_consumer_cancelled_by_broker)
```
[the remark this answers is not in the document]
lukebakken: That was probably an old change I can revert.

pika/adapters/select_connection.py
```diff
@@ -285,19 +285,19 @@ def __init__(self):
         self._r_interrupt = None
         self._w_interrupt = None
 
-    def add_timeout(self, deadline, callback_method):
-        """Add the callback_method to the IOLoop timer to fire after deadline
+    def add_timeout(self, deadline, callback):
```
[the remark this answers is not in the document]
lukebakken: This is something we can consider changing.

pika/channel.py
```diff
@@ -28,9 +29,6 @@ class Channel(object):
 
     """
 
-    # Disable pyling messages concerning "method could be a function"
-    # pylint: disable=R0201
-
```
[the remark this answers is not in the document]
lukebakken: I'll put that back, I don't remember why I removed it. We can do pylint changes in another PR like you suggested.

tests/acceptance/async_adapter_tests.py
lukebakken: Nope, we're requiring a callable for the `ack_nack_callback` parameter so I did it this way.

vitaly-krugl: got it, thanks!

PR #948, opened 2018-02-05
Ensure prefetch_size and prefetch_count are 0 or greater

### LINKED ISSUE #944: BlockingChannel.basic_qos hanging on non-integer parameters
Under Python 3.6.4 and Pika 0.11.2, the `BlockingChannel.basic_qos` method hangs when setting its `prefetch_size` or `prefetch_count` parameter to a non-integer value.

Input:

```
$ python3 <<EOF
import pika
with pika.BlockingConnection() as connection:
    channel.basic_qos(prefetch_size="hello")
EOF
```

lukebakken:
Fixes #944

### DIFF
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -442,16 +442,16 @@ class Channel(object):
                                    falls into other prefetch limits). May be set
                                    to zero, meaning "no specific limit",
                                    although other prefetch limits may still
-                                   apply. The prefetch-size is ignored if the
-                                   no-ack option is set.
+                                   apply. The prefetch-size is ignored by
+                                   consumers who have enabled the no-ack option.
         :param int prefetch_count: Specifies a prefetch window in terms of whole
                                    messages. This field may be used in
                                    combination with the prefetch-size field; a
                                    message will only be sent in advance if both
                                    prefetch windows (and those at the channel
                                    and connection level) allow it. The
-                                   prefetch-count is ignored if the no-ack
-                                   option is set.
+                                   prefetch-count is ignored by consumers who
+                                   have enabled the no-ack option.
         :param bool all_channels: Should the QoS apply to all channels
         :param callable callback: The callback to call for Basic.QosOk response
         :raises ValueError:
@@ -459,6 +459,8 @@ class Channel(object):
         """
         self._validate_channel()
         self._validate_rpc_completion_callback(callback)
+        self._validate_zero_or_greater('prefetch_size', prefetch_size)
+        self._validate_zero_or_greater('prefetch_count', prefetch_count)
         return self._rpc(spec.Basic.Qos(prefetch_size, prefetch_count,
                                         all_channels),
                          callback, [spec.Basic.QosOk])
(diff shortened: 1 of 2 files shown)


tests/unit/channel_tests.py
```diff
@@ -393,6 +393,18 @@ def test_basic_qos_raises_channel_closed(self):
...
+        self.assertRaises(ValueError, self.obj.basic_qos, None,
+                          'foo', 123, False)
+        self.assertRaises(ValueError, self.obj.basic_qos, None,
+                          -1, 123, False)
+
+    def test_basic_qos_raises_prefetch_count_value_error(self):
+        self.assertRaises(ValueError, self.obj.basic_qos, None,
+                          123, 'foo', False)
+        self.assertRaises(ValueError, self.obj.basic_qos, None,
+                          123, -1, False)
```
vitaly-krugl: In both of these tests, you need to validate that the parameters of the exception are what you expected not just the exception type. See example in https://github.com/pika/pika/blob/c0d0b3a2a69bdf394cd20d367a10a375e716e7b0/tests/unit/blocking_connection_tests.py#L149-L152. And you don't have to verify the verbatim exception text, but just some substring, such as `self.assertIn('prefetch_count', cm.exception.args[0])` (`ValueError` will have just the first string arg)

pika/channel.py
```diff
@@ -457,6 +457,8 @@ def basic_qos(self,
         :param bool all_channels: Should the QoS apply to all channels
```
vitaly-krugl: Unrelated to this PR, but might be good to clarify in "The prefetch-count is ignored if the no-ack option is set." above that this applies to consumers. Perhaps word it as follows: "The prefetch-count is ignored by consumers with the no-ack option turned on."

lukebakken:
@vitaly-krugl - this should be all set. I will address #951 next.

lukebakken:
@vitaly-krugl - just FYI, this is ready for your review. Thanks!

tests/unit/channel_tests.py
```diff
@@ -392,6 +392,26 @@ def test_basic_qos_raises_channel_closed(self):
         self.assertRaises(exceptions.ChannelClosed, self.obj.basic_qos, 0,
                           False, True)
 
+    def test_basic_qos_raises_prefetch_size_raises_error(self):
```
vitaly-krugl: Something doesn't look right with the method name - the word "raises" before "_prefetch_size". Did you mean to name it `test_basic_qos_*invalid*_prefetch_size_raises_error`?

tests/unit/channel_tests.py
```diff
@@ -392,6 +392,26 @@ def test_basic_qos_raises_channel_closed(self):
...
+        self.obj._set_state(self.obj.OPEN)
+        with self.assertRaises(ValueError) as ex:
+            self.obj.basic_qos('foo', 123)
+        self.assertEqual("invalid literal for int() with base 10: 'foo'",
+                         ex.exception.args[0])
+        with self.assertRaises(ValueError) as ex:
+            self.obj.basic_qos(-1, 123)
+        self.assertIn('prefetch_size', ex.exception.args[0])
+
+    def test_basic_qos_raises_prefetch_count_raises_error(self):
```
vitaly-krugl: Something doesn't look right with the method name - the word "raises" before "_prefetch_count". Did you mean to name it `test_basic_qos_*invalid*_prefetch_count_raises_error`?

tests/unit/channel_tests.py
lukebakken: Sure did, good catch

Issue #930, opened 2018-01-24
Connection drops when handling tasks with a long duration


GWSzeto:
I'm starting to get the same error as #418 even when I adjust the heartbeat connection parameter.
My tasks take usually around 5 mins, but even then, when I set the heartbeat to both 0 and some very large number (ie: 60000)

My code Is essentially identical to the issue I mentioned above and the error I receive is this:
`Traceback (most recent call last):
  File "instagram_worker.py", line 43, in <module>
    channel.start_consuming()
  File "/usr/local/lib/python2.7/dist-packages/pika/adapters/blocking_connection.py", line 1780, in start_consuming
    self.connection.process_data_events(time_limit=None)
  File "/usr/local/lib/python2.7/dist-packages/pika/adapters/blocking_connection.py", line 716, in process_data_events
    self._dispatch_channel_events()
  File "/usr/local/lib/python2.7/dist-packages/pika/adapters/blocking_connection.py", line 518, in _dispatch_channel_events
    impl_channel._get_cookie()._dispatch_events()
  File "/usr/local/lib/python2.7/dist-packages/pika/adapters/blocking_connection.py", line 1403, in _dispatch_events
    evt.body)
  File "instagram_worker.py", line 37, in dataHandler
    channel.basic_ack(delivery_tag=method.delivery_tag)
  File "/usr/local/lib/python2.7/dist-packages/pika/adapters/blocking_connection.py", line 1988, in basic_ack
    self._flush_output()
  File "/usr/local/lib/python2.7/dist-packages/pika/adapters/blocking_connection.py", line 1250, in _flush_output
    *waiters)
  File "/usr/local/lib/python2.7/dist-packages/pika/adapters/blocking_connection.py", line 474, in _flush_output
    result.reason_text)
pika.exceptions.ConnectionClosed: (-1, "error(104, 'Connection reset by peer')")`

vitaly-krugl:
See comment https://github.com/pika/pika/issues/892#issuecomment-356783950

vitaly-krugl:
And which version of Pika are you using?

GWSzeto:
I'm using version 0.11.2

And after reading your comment from the other issue, I just wanted to clarify if it is like this example from the pika Doumentation http://pika.readthedocs.io/en/0.10.0/examples/connecting_async.html

gmr:
Since it's async, are you letting the IO Loop process events while you're doing work or are you blocking the whole time?

[...]

Issue #927, opened 2018-01-19
Blocking connection hanging on basic_consume when using a pacemaker


stuartspotlight:
I've already posted this in a closed thread but I'm not sure if that's the correct way to do these things so I'm asking again here. I've created a little library to bind some processing function to a rabbitMQ queue and then perform that processing once something is placed on the queue. The problem I had is that sometimes the execution of the code is sometimes longer than the heartbeat so I added a pacemaker as described [here](https://github.com/eandersson/python-rabbitmq-examples/blob/master/Flask-examples/pika_async_rpc_example.py).
The code for this pacemaker is below:-

```
class heart_runner():
    
    def __init__(self, connection):
        
        if "blockingconnection" not in str(connection).lower() and "mock" not in str(connection).lower():
            raise Exception("Heartbeat runner requires a connection to rabbitmq as connection, actually has %s, a %s"%(str(connection), type(connection)))
        
        if connection.is_open == False:
            raise Exception("Heart runner's connection to rabbitmq should be open, is actually closed")
        
        self.connection = connection
        self.internal_lock = threading.Lock()
    

    def _process_data_events(self):
            """Check for incoming data events.
            We do this on a thread to allow the flask instance to send
            asynchronous requests.
            It is important that we lock the thread each time we check for events.
            """
    
            while True:
                with self.internal_lock:
                    self.connection.process_data_events()
                    #This is how often to run the pacemaker
                    time.sleep(0.1)
```
and then ran it in a separate thread using 

```
    #setup the pacemaker
    pacemaker = heart_runner(connection)

    #set up this deamon thread in order to avoid everything dying due to a heartbeat timeout
    thread = threading.Thread(target=pacemaker._process_data_events)
    thread.setDaemon(True)
    thread.start()
```

but I found when digesting large number of messages the thread would often hang on basic_consume. When I used ctrl+C to stop it the resulting error read:-

```
Traceback (most recent call last):
  File "./example_binding_main.py", line 121, in <module>
    bind(main, "example-function")
  File "/nanowire_plugin.py", line 287, in bind
    input_channel.start_consuming()
  File "/usr/local/lib/python3.6/site-packages/pika/adapters/blocking_connection.py", line 1780, in start_consuming
    self.connection.process_data_events(time_limit=None)
  File "/usr/local/lib/python3.6/site-packages/pika/adapters/blocking_connection.py", line 707, in process_data_events
    self._flush_output(common_terminator)
[...]

Issue #752, opened 2016-05-22
in another thread with hearbeat?


daoiqi:
https://www.rabbitmq.com/releases/rabbitmq-dotnet-client/v1.6.0/rabbitmq-dotnet-client-1.6.0-user-guide.pdf
``9.1. Connections, Channels, Sessions and Models``
> Each AMQP connection is represented by an instance of ConnectionBase. Each connection has one
thread running in the background, which reads and dispatches frames from the connection socket, as well as acting as a serialization point for outbound AMQP commands. There may also be two additional threads responsible for Heartbeat support (In and Outbound), if it was negotiated during the
connection open process.

but in BlockConnection, this is all in one thread. It is necessary with another thread to deal hearbeat ?

vitaly-krugl:
That's a correct observation about `pika.BlockingConnection`. It's completely synchronous. If you are not doing any I/O on the given pika connection, you need to periodically call the connection's `BlockingConnection.process_data_events` or `BlockingConnection.sleep` method.

I am working on a threaded prototype.

daoiqi:
I have some  long time I/O. (more than 10 min )

When recive data, consumer do something, but it is long time task.  When  consumer has done, then ack to MQ, and publish new data.

If all in one thread, consumer doing task  with block, then the heartbeats not any thread to deal with.

Or I'm using `ack` mechanism error?   

which is  better?

```
1. revice data then ack to MQ  
2. do long time.         ----- if server is crash, heartbeats is not significance.
3. publish new data.
```

or

```
1. recive data
2. do long time and hearbeat check    --- if crash, mq will resend message
3. publish new data and ack to MQ.
```

Can I start thread to do I/O task, and another thread do `BlockingConnection.process_data_events`?

vitaly-krugl:
If you have a long computation, then you can set a large heartbeat timeout (longer than your anticipated compute time) or turn off heartbeat altogether. Note that the 0.10.0 version of pika doesn't facilitate setting a longer heartbeat timeout - pika in master branch has a fix for it.

Another option is to periodically call `BlockingConnection.process_data_events` while performing you long computation. Make sure that you call `BlockingConnection.process_data_events` frequently enough to avoid heartbeat timeout.

pika connection instances are not thread-safe, so calling `BlockingConnection.process_data_events` from another thread will cause bad things to happen.

tcwalther:
[...]

PR #956, opened 2018-02-14
Implement add_callback_threadsafe in all connection adapters

### LINKED ISSUE #892: Question re. interrupting `BlockingChannel.consume` from another thread

vitaly-krugl:
(*NOTE that the failed coverage test results from pika's encrypted credentials configuration issue*)

This pull request implements a new method `add_callback_threadsafe` on all supported connection adapters in pika. This method enables the application to delegate calls from other threads into the context of the adapter's IOLoop thread (similarly to Tornado's `add_callback`, Twisted's `callFromThread`, and asyncio's `call_soon_threadsafe`).

Lack of this feature has been a pain point for users of `BlockingConnection` and `SelectConnection` for as long as I can remember, and users have had to rely on clumsy and inefficient workarounds.

Here is a sampling of several relevant issues: #963, #927, #892, #930, #752, #418, #753, #636, #564.

For example, this feature facilitates the app to delegate long-lasting processing of incoming messages to a background thread and then use `add_callback_threadsafe` to efficiently delegate ACKing of the message in the context of the connection's thread, thus avoiding lost heartbeat/connection issues from blocking too long and also hacky polling techniques that applications have been using to work around the limitation in the past.

ADDITIONAL INFO FOR CHANGELOG: It is customary among frameworks - such as tornado, twisted, and asyncio - to designate a single threadsafe method that accepts a callback to be executed in the IOLoop's thread, and no others are thread-safe for ease of support/maintenance and for API orthogonality. Following this rationale, I removed the thread-safe designation inside the docstring of the ioloop `stop()` method in `select_connection.IOLoop`.

Fixes #892

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -303,9 +303,6 @@ class IOLoop(object):
     def close(self):
         """Release IOLoop's resources.
 
-        TODO Need to call this from IOLoop and SelectConnection tests as well as
-        TODO from BlockingConnection to avoid socket resource leak messages
-
         `IOLoop.close` is intended to be called by the application or test code
         only after `IOLoop.start()` returns. After calling `close()`, no other
         interaction with the closed instance of `IOLoop` should be performed.
@@ -405,7 +402,7 @@ class IOLoop(object):
 
     def process_timeouts(self):
         """[Extension] Process pending callbacks and timeouts, invoking those
-         whose time has come. Internal use only.
+        whose time has come. Internal use only.
 
         """
         # Avoid I/O starvation by postponing new callbacks to the next iteration
@@ -465,12 +462,8 @@ class IOLoop(object):
 
     def stop(self):
         """[API] Request exit from the ioloop. The loop is NOT guaranteed to
-        stop before this method returns. This is the only method that may be
-        called from another thread.
+        stop before this method returns.
 
-        TODO This shouldn't have been made thread-safe. Instead, `add_callback`
-        TODO should be used to safely call `stop` from the IOLoop's own thread.
-        TODO Update documentation/implementation
         """
         self._poller.stop()
 
@@ -545,9 +538,6 @@ class _PollerBase(_AbstractBase):  # pylint: disable=R0902
 
     def close(self):
         """Release poller's resources.
-
-        TODO Need to call this from Poller tests to avoid socket resource
-        TODO leak messages
 
         `close()` is intended to be called after the poller's `start()` method
         returns. After calling `close()`, no other interaction with the closed
@@ -738,8 +728,7 @@ class _PollerBase(_AbstractBase):  # pylint: disable=R0902
 
     def stop(self):
         """Request exit from the ioloop. The loop is NOT guaranteed to stop
-        before this method returns. This is the only method that may be called
-        from another thread.
+        before this method returns.
 
         """
         LOGGER.debug('Stopping IOLoop')
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -162,6 +162,12 @@ class _Timer(object):
         # collection of canceled timeouts
         self._num_cancellations = 0
 
+    def close(self):
+        """Release resources. Don't use the `_Timer` instance after closing
+        it
+        """
+        self._timeout_heap = None
+
     def call_later(self, delay, callback):
         """Schedule a one-shot timeout given delay seconds.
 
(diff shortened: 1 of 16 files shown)


vitaly-krugl:
Anyone have any idea why the Coverage job in travis-ci is always failing in my PR?

cc @michaelklishin, @lukebakken

shinji-s:
The attempt to populate coverage/ directory from s3 storage is failing because credential is not installed/found. Sorry I don't know how to fix though.

lukebakken:
@shinji-s pointed it out -

```
The command "mkdir coverage" exited with 0.
0.49s$ aws s3 cp --recursive s3://com-gavinroy-travis/pika/$TRAVIS_BUILD_NUMBER/ coverage
fatal error: Unable to locate credentials
```

@gmr - would you have time to address this? Thanks!!!

shinji-s:
Regarding the coverage test failure, my guess below may be well off the mark but I'd try.

The credential is given here in .travis.yml
>   \- secure:  {actual-encrypted-string-is-omitted}
>   \- secure:  {actual-encrypted-string-is-omitted}

However, https://docs.travis-ci.com/user/encryption-keys/ has this to say:
> Please note that encrypted environment variables are not available for pull requests from forks.

That is why the credential is not available in the above build?

vitaly-krugl:
@shinji-s, thank you for looking into this. Your assessment makes a ton of sense. If that's the case, then the selected mechanism is inappropriate for the project's needs, since most PRs would naturally come from forks and reviewers need the means to assess impact of the PR on code coverage.

lukebakken:
Reviewing now

pika/adapters/select_connection.py
```diff
@@ -164,6 +163,16 @@ def __init__(self):
         # collection of canceled timeouts
         self._num_cancellations = 0
 
+    def close(self):
```
lukebakken: Should all of the connection adapters implement `close`? It looks like the Twisted connection `IOLoopReactorAdapter` does not, and the [Tornado docs](http://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.close) state that `stop` must be called before close which isn't enforced in our code. There's also a good chance I'm mis-understanding how these adapters work

vitaly-krugl: @lukebakken - good points. I will follow up in a couple of days. Thanks!

vitaly-krugl: @lukebakken, the pull request is ready for resuming code review. Thanks!

vitaly-krugl: I also discovered the existing issue #970 in the process of implementing new tests.

vitaly-krugl:
@lukebakken - this pull request contains a large number of commits - 39 presently; would it be better if I squashed them into a single commit?

pika/adapters/select_connection.py
lukebakken: Great! I'll get on it today.

vitaly-krugl:
Thanks @lukebakken !

lukebakken:
Great work, this will be very useful for Pika users

PR #957, opened 2018-02-14
Ensure blocked Channel.Close methods are processed

### LINKED ISSUE #945: Tornado adapter: passive exchange.declare scenario where close hangs
Updated bug description:

When a channel is closed due to a passive declaration of an exchange that does not exist, processing of the resulting channel closure hangs.

----

Original description (kept so the context of earlier comments is not lost):

Not sure exactly when it started happening, but lately, we've been getting this message, a lot, when an AMQP client tries to passively declare a number of exchanges that are not present.

Correct behaviour is that the AMQP broker closes the channel with a "not found" exception, and our code retries opening the channel.  It seems `pika` trips up on the latter bit, confusing the AMQP broker which then responds by closing the entire connection.

We're using the Tornado client adapter on Python 2.7.

```
2018-02-05 01:02:34,929 pika.connection[connection.py:1382] WARNING Disconnected from RabbitMQ at rabbitmq:5672 (504): CHANNEL_ERROR - second 'channel.open' seen
2018-02-05 01:02:35,042 pika.adapters.base_connection[base_connection.py: 212] INFO Connecting to 172.18.0.4:5672
2018-02-05 01:02:35,047 pika.credentials[credentials.py:  71] INFO Erasing stored credential values
2018-02-05 01:02:35,763 pika.adapters.base_connection[base_connection.py: 180] WARNING Socket closed when connection was open
2018-02-05 01:02:35,764 pika.adapters.base_connection[base_connection.py: 308] WARNING Connection is closed but not stopping IOLoop
2018-02-05 01:02:35,764 pika.connection[connection.py:1382] WARNING Disconnected from RabbitMQ at rabbitmq:5672 (504): CHANNEL_ERROR - second 'channel.open' seen
2018-02-05 01:02:38,288 pika.adapters.base_connection[base_connection.py: 212] INFO Connecting to 172.18.0.4:5672
2018-02-05 01:02:38,295 pika.credentials[credentials.py:  71] INFO Erasing stored credential values
2018-02-05 01:02:38,995 pika.adapters.base_connection[base_connection.py: 180] WARNING Socket closed when connection was open
2018-02-05 01:02:38,996 pika.adapters.base_connection[base_connection.py: 308] WARNING Connection is closed but not stopping IOLoop
2018-02-05 01:02:38,997 pika.connection[connection.py:1382] WARNING Disconnected from RabbitMQ at rabbitmq:5672 (504): CHANNEL_ERROR - second 'channel.open' seen
```

I'll also point out that the "Connection is closed but not stopping IOLoop" is annoying; the `IOLoop` does not belong to `pika`, so it isn't `pika`'s place to say whether the `IOLoop` should be running or stopped.

lukebakken:
If Pika receives a server-initiated Channel.Close method, it will not process blocked Channel.Close methods which could hang application code.

See https://gist.github.com/lukebakken/ab49531e504dfd8f4c133fcb9994a7d7#file-testpika-py for code to reproduce.

Fixes #945

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -31,7 +31,6 @@ class BaseConnection(connection.Connection):
     ]
     ERRORS_TO_IGNORE = [errno.EWOULDBLOCK, errno.EAGAIN, errno.EINTR]
     DO_HANDSHAKE = True
-    WARN_ABOUT_IOLOOP = False
 
     def __init__(self,
                  parameters=None,
@@ -307,8 +306,6 @@ class BaseConnection(connection.Connection):
         """
         if self.stop_ioloop_on_close and self.ioloop:
             self.ioloop.stop()
-        elif self.WARN_ABOUT_IOLOOP:
-            LOGGER.warning('Connection is closed but not stopping IOLoop')
 
     def _handle_error(self, error_value):
         """Internal error handling method. Here we expect a socket error
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -1278,21 +1278,25 @@ class Channel(object):
     def _drain_blocked_methods_on_remote_close(self):
         """This is called when the broker sends a Channel.Close. It checks the
         blocked method queue for pending client-initiated Channel.Close methods
-        and ensures their callbacks are processed
+        and ensures their callbacks are processed, but does *not* execute the
+        blocked method
 
         """
         LOGGER.debug('Draining %i blocked frames due to remote Channel.Close', len(self._blocked))
         self._blocking = None
         # self._blocking must be checked here as a callback could
         # potentially change the state of that variable during an
-        # iteration of the while loop
+        # iteration of the while loop. Technically, we can only enter
+        # this method once due to the self.is_closing check, but we
+        # will iterate over all blocked methods for debug logging purposes
         while self._blocked and self._blocking is None:
             method = self._blocked.popleft()[0]
             if isinstance(method, spec.Channel.Close):
                 # in this case, the Channel.Close method has not yet been
-                # sent to the broker. At this point, the channel has been
-                # closed by the broker, so we do not want to send this method,
-                # but process it as though it had been sent
+                # sent to the broker but the channel has been
+                # closed by the broker by a remote Channel.Close. We do not
+                # want to send this method, but process it as though it had
+                # been sent
                 reply_code = self._closing_code_and_text[0]
                 reply_text = self._closing_code_and_text[1]
                 self._on_close_meta(reply_code, reply_text)
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -1269,6 +1269,9 @@ class Channel(object):
         """
         LOGGER.debug('%i blocked frames', len(self._blocked))
         self._blocking = None
+        # self._blocking must be checked here as a callback could
+        # potentially change the state of that variable during an
+        # iteration of the while loop
         while self._blocked and self._blocking is None:
             self._rpc(*self._blocked.popleft())
 
@@ -1280,9 +1283,12 @@ class Channel(object):
         """
         LOGGER.debug('Draining %i blocked frames due to remote Channel.Close', len(self._blocked))
         self._blocking = None
+        # self._blocking must be checked here as a callback could
+        # potentially change the state of that variable during an
+        # iteration of the while loop
         while self._blocked and self._blocking is None:
             method = self._blocked.popleft()[0]
-            if method.NAME == spec.Channel.Close.NAME:
+            if isinstance(method, spec.Channel.Close):
                 # in this case, the Channel.Close method has not yet been
                 # sent to the broker. At this point, the channel has been
                 # closed by the broker, so we do not want to send this method,
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -323,7 +323,7 @@ class Channel(object):
                                      exclusive=exclusive,
                                      arguments=arguments or dict()),
                   rpc_callback, [(spec.Basic.ConsumeOk,
-                                 {'consumer_tag': consumer_tag})])
+                                  {'consumer_tag': consumer_tag})])
 
         return consumer_tag
 
@@ -1066,13 +1066,14 @@ class Channel(object):
         self._send_method(spec.Channel.CloseOk())
 
         if self.is_closing:
-            # Since we already sent Channel.Close, we need to wait for CloseOk
+            # Since we may have already sent Channel.Close, we need to wait for CloseOk
             # before cleaning up to avoid a race condition whereby our channel
             # number might get reused before our CloseOk arrives
 
             # Save the details to provide to user callback when CloseOk arrives
             self._closing_code_and_text = (method_frame.method.reply_code,
                                            method_frame.method.reply_text)
+            self._drain_blocked_methods_on_remote_close()
         else:
             self._set_state(self.CLOSED)
             try:
@@ -1084,10 +1085,11 @@ class Channel(object):
                 self._cleanup()
 
     def _on_close_meta(self, reply_code, reply_text):
-        """Handle meta-close request from Connection's cleanup logic after
-        sudden connection loss. We use this opportunity to transition to
-        CLOSED state, clean up the channel, and dispatch the on-channel-closed
-        callbacks.
+        """Handle meta-close request from either a remote Channel.Close from
+        the broker (when a pending Channel.Close method is queued for
+        execution) or a Connection's cleanup logic after sudden connection
+        loss. We use this opportunity to transition to CLOSED state, clean up
+        the channel, and dispatch the on-channel-closed callbacks.
 
         :param int reply_code: The reply code to pass to on-close callback
         :param str reply_text: The reply text to pass to on-close callback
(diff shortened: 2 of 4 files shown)


pika/channel.py
```diff
@@ -1066,13 +1066,14 @@ def _on_close(self, method_frame):
...
         if self.is_closing:
-            # Since we already sent Channel.Close, we need to wait for CloseOk
+            # Since we may have already sent Channel.Close, we need to wait for CloseOk
             # before cleaning up to avoid a race condition whereby our channel
             # number might get reused before our CloseOk arrives
 
             # Save the details to provide to user callback when CloseOk arrives
             self._closing_code_and_text = (method_frame.method.reply_code,
                                            method_frame.method.reply_text)
+            self._drain_blocked_methods_on_remote_close()
```
vitaly-krugl: Please add a code comment describing what prevents the blocked `Channel.Close` from becoming unblocked and sent to broker on an open connection so that it must be discarded this way manually.

pika/channel.py
```diff
@@ -1270,6 +1272,27 @@ def _on_synchronous_complete(self, _method_frame_unused):
...
+        """This is called when the broker sends a Channel.Close. It checks the
+        blocked method queue for pending client-initiated Channel.Close methods
+        and ensures their callbacks are processed
+
+        """
+        LOGGER.debug('Draining %i blocked frames due to remote Channel.Close', len(self._blocked))
+        self._blocking = None
+        while self._blocked and self._blocking is None:
+            method = self._blocked.popleft()[0]
+            if method.NAME == spec.Channel.Close.NAME:
```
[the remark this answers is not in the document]
lukebakken: I was just trying to think of a way to do that 😖

pika/adapters/base_connection.py
```diff
@@ -307,8 +306,6 @@ def _handle_ioloop_stop(self):
         """
         if self.stop_ioloop_on_close and self.ioloop:
             self.ioloop.stop()
-        elif self.WARN_ABOUT_IOLOOP:
-            LOGGER.warning('Connection is closed but not stopping IOLoop')
```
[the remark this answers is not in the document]
lukebakken: Yes, the person who reported this issue mentions it specifically:

> I'll also point out that the "Connection is closed but not stopping IOLoop" is annoying; the IOLoop does not belong to pika, so it isn't pika's place to say whether the IOLoop should be running or stopped.

It could be changed to a debug message, but I don't see what value it adds since the message doesn't apply to other IO loop types.

pika/channel.py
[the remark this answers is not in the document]
lukebakken: Ah, yes it would

pika/channel.py
lukebakken: How about this comment?

https://github.com/pika/pika/blob/pika-945/pika/channel.py#L1286-L1289

lukebakken:
@vitaly-krugl - all set for review #2, thanks

pika/channel.py
vitaly-krugl: @lukebakken - that comment doesn't explain what is being asked.

pika/channel.py
```diff
@@ -1267,9 +1269,36 @@ def _on_synchronous_complete(self, _method_frame_unused):
...
+
+        """
+        LOGGER.debug('Draining %i blocked frames due to remote Channel.Close', len(self._blocked))
+        self._blocking = None
+        # self._blocking must be checked here as a callback could
+        # potentially change the state of that variable during an
+        # iteration of the while loop
+        while self._blocked and self._blocking is None:
+            method = self._blocked.popleft()[0]
+            if isinstance(method, spec.Channel.Close):
```
vitaly-krugl: There should never be more than one `spec.Channel.Close`  in `self._blocked` because `Channel.close` sets the `CLOSING` state and will raise an exception if called again in that state. So can probably break out of the loop immediately after calling `self._on_close_meta`. Might be good to include an explanation to this effect as code commend before breaking out.

tests/acceptance/async_adapter_tests.py
```diff
@@ -162,7 +163,30 @@ def on_exchange_declared(self, frame):
 
     def on_bad_result(self, frame):
         self.channel.exchange_delete(self.name)
-        raise AssertionError("Should not have received a Queue.DeclareOk")
+        raise AssertionError("Should not have received an Exchange.DeclareOk")
+
+
+class TestPassiveExchangeDeclareWithConcurrentClose(AsyncTestCase, AsyncAdapters):
```
vitaly-krugl: Was this test reproducing the failure prior to your fix?

lukebakken: Sure does, I'll add a comment about it to clarify, since it doesn't really look like it tests anything.

pika/channel.py
lukebakken: What do you think about continuing the loop just so that any other blocked methods are debug-logged?

pika/channel.py
lukebakken: It sure doesn't, because it's the wrong set of lines! 😄 I meant to link to this:

https://github.com/pika/pika/blob/pika-945/pika/channel.py#L1292-L1295

I'll move the comment and make it clearer what the scenario is.

lukebakken:
@vitaly-krugl - all set for review 3 🎣

pika/channel.py
```diff
@@ -1267,9 +1269,40 @@ def _on_synchronous_complete(self, _method_frame_unused):
...
+        # this method once due to the self.is_closing check, but we
+        # will iterate over all blocked methods for debug logging purposes
+        while self._blocked and self._blocking is None:
+            method = self._blocked.popleft()[0]
+            if isinstance(method, spec.Channel.Close):
+                # in this case, the Channel.Close method has not yet been
+                # sent to the broker but the channel has been
+                # closed by the broker by a remote Channel.Close. We do not
+                # want to send this method, but process it as though it had
+                # been sent
```
vitaly-krugl: > We do not want to send this method, but process it as though it had been sent

Just FYI: technically, there would be no harm in sending the blocked `Channel.Close` to the server, since AMQP 0.9.1 specification obliges the server to respond with `Close-OK` in this case, while dropping all other incoming methods, save for `Close-OK`

pika/channel.py
vitaly-krugl: Sorry, no -- it explains the side-effect, but not the root cause of the problem that we're working around with this code, which prevents reviewers/maintainers from gaining a valuable insight.

Here is what I consider to be the root cause of what this PR is fixing, which would be helpful to explain in a comment above the call to `self._drain_blocked_methods_on_remote_close`:

The server closes the channel before responding to an outstanding synchronous method (that possibly triggered the channel error and which isn't `Channel.Close`). AMQP 0.9.1 obliges the server to drop all methods arriving on a closed channel other than `Close-OK` and `Channel.Close`. Since the response to that synchronous method never arrives, the channel never becomes unblocked, and the `Channel.Close`, if any, in the blocked queue has no opportunity to be sent, and thus it's completion callback of the `Channel.Close` would never be called.

pika/channel.py
lukebakken: That's interesting, because that's not what happens with RabbitMQ. What I saw is `504` error response where RabbitMQ says it expects a `channel.open`. Maybe because in that case the test used the same channel number?

lukebakken:
@vitaly-krugl thanks for providing the comment you would like to see. I added it to the docstring for `_drain_blocked_methods_on_remote_close`. Let me know what you think.

pika/channel.py
vitaly-krugl: > What do you think about continuing the loop just so that any other blocked methods are debug-logged?

May be marginal, because: 1. if channel errors out, the developer should reasonably expect that some initiated requests would not complete (also since they didn't get completion callbacks); and 2. The one(s) that made it into the transport buffer and/or socket would not show up in these debug statements.

pika/channel.py
```diff
@@ -1267,9 +1269,40 @@ def _on_synchronous_complete(self, _method_frame_unused):
...
+        # potentially change the state of that variable during an
+        # iteration of the while loop
         while self._blocked and self._blocking is None:
             self._rpc(*self._blocked.popleft())
 
+    def _drain_blocked_methods_on_remote_close(self):
+        """This is called when the broker sends a Channel.Close while the
+        client is in CLOSING state. This method checks the blocked method
+        queue for a pending client-initiated Channel.Close method and
+        ensures its callbacks are processed, but does send the method to
```
vitaly-krugl: Should be "but does \*not\* send"

pika/channel.py
```diff
@@ -1267,9 +1269,40 @@ def _on_synchronous_complete(self, _method_frame_unused):
...
+        the broker.
+
+        The broker may close the channel before responding to outstanding
+        (blocked) synchronous methods, or even before these methods have
+        been sent to the broker. AMQP 0.9.1 obliges the server to drop all
+        methods arriving on a closed channel other than Channel.CloseOk and
+        Channel.Close. Since the response to the blocked synchronous
+        method never arrives, the channel never becomes unblocked, and the
+        Channel.Close, if any, in the blocked queue has no opportunity to
+        be sent, and thus its completion callback would never be called.
```
vitaly-krugl: Corrections:

        The broker may close the channel before responding to outstanding in-transit
        synchronous methods, or even before these methods have
        been sent to the broker. AMQP 0.9.1 obliges the server to drop all
        methods arriving on a closed channel other than Channel.CloseOk and
        Channel.Close. Since the response to a synchronous method that blocked
        the channel never arrives, the channel never becomes unblocked, and the
        Channel.Close, if any, in the blocked queue has no opportunity to
        be sent, and thus its completion callback would never be called.

vitaly-krugl: @lukebakken - we're almost there. Just comment changes remain.

lukebakken:
@vitaly-krugl OK comments addressed. Thank you for the thorough review.

lukebakken:
🎉

PR #964, opened 2018-02-17
Eliminate the `stop_ioloop_on_close` arg from all async connection adapters

### LINKED ISSUE #961: Remove stop_ioloop_on_close arg from adapter constructors for the 1.0 release
Some time ago, the `stop_ioloop_on_close=True` constructor arg slipped into all the asynchronous adapters. I think that the presence of this arg is overly-opinionated. An asynchronous app is likely to have other logic besides a single pika connection running on the same IOLoop (tornado, asyncio, etc.), so having ioloop-stopping logic in pika adapters is too pika-centric, unnecessary, non-orthogonal, and likely error-prone. Especially when it's brain-dead easy for the app to pass `on_close_callback`- having the signature `on_close_callback(connection, reason_code, reason_text)` - to the adapter constructor and call `connection.ioloop.stop()` itself if needed.

I think we should remove the `stop_ioloop_on_close` args everywhere in pika as well as any logic dealing with stopping the ioloop in pika adapters.

vitaly-krugl:
Closes #961

### DIFF
--- a/docs/examples/asynchronous_consumer_example.rst
+++ b/docs/examples/asynchronous_consumer_example.rst
@@ -55,8 +55,7 @@ consumer.py::
             """
             LOGGER.info('Connecting to %s', self._url)
             return pika.SelectConnection(pika.URLParameters(self._url),
-                                         self.on_connection_open,
-                                         stop_ioloop_on_close=False)
+                                         self.on_connection_open)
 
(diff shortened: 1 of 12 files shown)


lukebakken:
Reviewing this now :smiley_cat:

Issue #779, opened 2016-10-26
pika.exceptions.ConnectionClosed


NicolasDuran:
Using **RabbitMQ 3.6.5** on **Debian 8**
I create a queue containing the numbers for 1 to 100000.
I have workers running on docker (based on debian latest), containing pika.version = 0.10.0
```
import pika

conn = pika.URLParameters('amqp://xxx:xxx@spotted-monkey.rmq.cloudamqp.com/ctuzjcxs')
connection = pika.BlockingConnection(conn)

channel = connection.channel()
channel.queue_declare(queue='task_queue', durable=True)

def callback(ch, method, properties, body):
    print('GOT:' + str(body))
    ch.basic_ack(delivery_tag = method.delivery_tag)

channel.basic_qos(prefetch_count=1)
channel.basic_consume(callback, queue='task_queue')
channel.start_consuming()
```
Some of the workers works perfectly, some other fails very quickly with errors such as :
 
```
 3> ERROR:pika.adapters.base_connection:Socket Error: 104
 3> WARNING:pika.adapters.base_connection:Socket closed when connection was open
 3> WARNING:pika.connection:Disconnected from RabbitMQ at spotted-monkey.rmq.cloudamqp.com:5672 (0): Not specified
 3> CRITICAL:pika.adapters.blocking_connection:Connection close detected
 3> Traceback (most recent call last):
 3>   File "receiver-class2.py", line 25, in <module>
 3>     channel.start_consuming()
 3>   File "/usr/local/lib/python3.4/site-packages/pika/adapters/blocking_connection.py", line 1681, in start_consuming
 3>     self.connection.process_data_events(time_limit=None)
 3>   File "/usr/local/lib/python3.4/site-packages/pika/adapters/blocking_connection.py", line 647, in process_data_events
 3>     self._flush_output(common_terminator)
 3>   File "/usr/local/lib/python3.4/site-packages/pika/adapters/blocking_connection.py", line 426, in _flush_output
 3>     raise exceptions.ConnectionClosed()
 3> pika.exceptions.ConnectionClosed
```
or 
```
 0> Traceback (most recent call last):
 0>   File "/usr/local/lib/python3.4/site-packages/pika/adapters/base_connection.py", line 384, in _handle_read
 0>     data = self.socket.recv(self._buffer_size)
 0> ConnectionResetError: [Errno 104] Connection reset by peer
 0> 
 0> During handling of the above exception, another exception occurred:
 0> 
 0> Traceback (most recent call last):
[...]

PR #975, opened 2018-02-24
Clean up connection socket error handling

vitaly-krugl:
`BasicConnection._handle_events` took redundant args with default values that are actually not passed by anything calling it, yet it passed the `error` arg (defaulting to None and not set explicitly by any of the callers) to `_handle_error` upon ERROR socket event, which was resulting in the confusing log message "Tried to handle an error where no error existed".

In fact, `_handle_events` shouldn't even react to the ERROR event individually, since when there is an error on the TCP/IP stream, an IO event (READ and/or WRITE) will also be indicated and a subsequent attempt by pika to perform the corresponding I/O on the socket will generate the appropriate error condition and be handled appropriately by read/write handler(s). In fact, you could have an ERROR event (e.g., HUP) indicated on the socket, yet a socket read could succeed if incoming data was still available on the socket buffer. So it's better to handle the error once attempted I/O trips up on it.

From https://github.com/pika/pika/issues/779#issuecomment-368005734:
```
[ERROR] Error event 25, None
[CRITICAL] Tried to handle an error where no error existed
[ERROR] Socket Error: 104
[INFO] Disconnected from RabbitMQ at blablabla.com:5672 (-1): ConnectionResetError(104, 'Connection reset by peer')
[ERROR] Connection close detected; result=BlockingConnection__OnClosedArgs(connection=<SelectConnection CLOSED socket=None params=<ConnectionParameters host=blablabla.com port=5672 virtual_host=/ ssl=False>>, reason_code=-1, reason_text="ConnectionResetError(104, 'Connection reset by peer')")
```

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -304,7 +304,7 @@ class BaseConnection(connection.Connection):
         # called), etc., etc., etc.
         self._manage_event_state()
 
-    def _handle_connection_sock_error(self, exception):
+    def _handle_connection_socket_error(self, exception):
         """Internal error handling method. Here we expect a socket error
         coming in and will handle different socket errors differently.
 
@@ -350,7 +350,7 @@ class BaseConnection(connection.Connection):
         only have the socket in a blocking state during connect."""
         LOGGER.warning("Unexpected socket timeout")
 
-    def _handle_connection_sock_events(self, fd, events):
+    def _handle_connection_socket_events(self, fd, events):
         """Handle IO/Event loop events, processing them.
 
         :param int fd: The file descriptor for the events
@@ -405,12 +405,12 @@ class BaseConnection(connection.Connection):
                 # ssl wants more data but there is nothing currently
                 # available in the socket, wait for it to become readable.
                 return 0
-            return self._handle_connection_sock_error(error)
+            return self._handle_connection_socket_error(error)
 
         except SOCKET_ERROR as error:
             if error.errno in (errno.EAGAIN, errno.EWOULDBLOCK):
                 return 0
-            return self._handle_connection_sock_error(error)
+            return self._handle_connection_socket_error(error)
 
         if not data:
             # Remote peer closed or shut down our input stream - disconnect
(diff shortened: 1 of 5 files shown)


pika/adapters/base_connection.py
```diff
@@ -382,20 +365,19 @@ def _handle_events(self, fd, events, error=None, write_only=False):
             self._handle_write()
             self._manage_event_state()
 
-        if self.socket and not write_only and (events & self.READ):
```
vitaly-krugl: `write_only` was most likely a vestige of the original `BlockingConnection` implementation prior to my rewrite of `BlockingConnection`.

pika/adapters/base_connection.py
```diff
@@ -315,63 +304,57 @@ def _flush_outbound(self):
...
-                           repr(error_value))
+                           repr(exception))
 
     def _handle_timeout(self):
         """Handle a socket timeout in read or write.
         We don't do anything in the non-blocking handlers because we
         only have the socket in a blocking state during connect."""
         LOGGER.warning("Unexpected socket timeout")
 
-    def _handle_events(self, fd, events, error=None, write_only=False):
```
vitaly-krugl: `write_only` was most likely a vestige of the original `BlockingConnection` implementation prior to my rewrite of `BlockingConnection`.

vitaly-krugl:
@lukebakken - this is ready for code review

pika/adapters/base_connection.py
```diff
@@ -382,27 +365,27 @@ def _handle_events(self, fd, events, error=None, write_only=False):
...
+            LOGGER.debug('Sock error events %r; sock error code %r', events,
+                         error)
 
     def _handle_read(self):
         """Read from the socket and call our on_data_available with the data."""
         try:
             while True:
                 try:
                     if self.params.ssl:
+                        # TODO Why using read instead of recv on ssl socket?
```
lukebakken: I suspect it is due to [this bug](https://bugs.python.org/issue3890) which has been resolved for a long time.

lukebakken:
Thanks @vitaly-krugl !!!

PR #978, opened 2018-02-26
pika.compat should be a module, not package

lukebakken:
Fixes the issue raised here:

https://github.com/pika/pika/issues/945#issuecomment-366148312


anton-ryzhov:
@vitaly-krugl because `distutils` installs only **packages** listed under `packages` keyword in `setup.py`. `compat.py` was a part of `pika` package, but `compat/__init__.py` is a new `pika.compat` package, different than just `pika`.
If you add new [sub]packages, you must list it in `setup.py`. Or use http://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages

vitaly-krugl:
> because distutils installs only packages listed under packages keyword

So, one option was to list `compat/__init__.py` in setup.py instead of turning it into `compat` module? Just curious.

anton-ryzhov:
Yes, sure, like this:
```python
setuptools.setup(
    …
    packages=['pika', 'pika.adapters', 'pika.compat'],
    …
)
```
or even better (autodiscovery):
```python
from setuptools import find_packages

setuptools.setup(
    …
    packages=find_packages(exclude=['*.tests', '*.tests.*', 'tests.*', 'tests']),
    …
)
```

lukebakken:
@vitaly-krugl apologies that I missed your question and thank you @anton-ryzhov for answering it.

vitaly-krugl:
Thanks @anton-ryzhov!

@lukebakken - I like the `find_packages()` recommendation for auto-discovery. What do you think?

lukebakken:
I doubt Pika will add packages in the future. The issue around `compat` was all my mistake in thinking it had to be a package.

PR #983, opened 2018-02-28
WIP: Enable TLS/SSL testing on CI infrastructure

### LINKED ISSUE #981: Enable RabbitMQ to accept SSL connections in addition to plaintext on all Travis-CI and AppVeyor builds
I would like to fix/clean up SSL support in pika. Could you please help by enabling RabbitMQ to accept SSL connections in addition to plaintext on all Travis-CI and AppVeyor builds? Also provide rudimentary documentation where the client test code can find the necessary artifacts (e.g., public cert).

Please make sure that the certs never expire or at least don't expire for many years.

See https://github.com/pika/pika/blob/master/docs/examples/tls_mutual_authentication.rst and https://github.com/pika/pika/blob/master/docs/examples/tls_server_uathentication.rst for examples.

We would also need some reasonable strategy for users who run the pika test suite in their own environments. Perhaps the tests discover the location of the test  cert via a specific well-named environment variable and skip if said environment variable is not defined or some such thing?

lukebakken:
Fixes #981


tests/acceptance/async_adapter_tests.py
```diff
@@ -612,7 +612,7 @@ def begin(self, channel):
         timer1 = channel.connection.add_timeout(0, self.on_my_timer)
         timer2 = channel.connection.add_timeout(0, self.on_my_timer)
 
-        self.assertNotEqual(timer1, timer2)
```
vitaly-krugl: Were you seeing this fail too?

lukebakken: Yes, since the timer class implements value equality this would evaluate to equal as both timers have `0` as their deadline.

vitaly-krugl:
The test "Setup 5 timeout handlers and observe them get invoked one by one" has been flaky. I am guessing that the test machines are very slow which sometimes causes a conflict with the selected granularity of timers. Since it's a unit test, it should be mocking time.time() for control and speed of execution instead of slowing down the testing by making the test wait for the actual timeouts in real time.

PR #987, opened 2018-03-04
Use standard ssl.SSLContext instead of homegrown values in pika.SSLOptions

### LINKED ISSUE #986: Change pika to accept ssl.SSLContext instead of dict or homegrown SSLOptions attributes for 1.0.0
Pre 1.0.0, pika accepts either a dict or a homegrown `pika.connection.SSLOptions` instance and then [converts it to `ssl.SSLContext` (in case of `SSLOptions`) or passes the dict as kwargs to `ssl.wrap_socket()`](https://github.com/pika/pika/blob/09270712bf5912566df8af10ae68a1ddaea4f4e1/pika/adapters/base_connection.py#L512-L539).

I can't think of a good reason to march on to 1.0.0 with this status quo. It's unnecessary and limiting. User's should just be able to pass an `ssl.SSLContext` instance which gives them most flexibility, including the ability to cache ssl contexts. For example, `asyncio` in Python 3 accepts only `ssl.SSLContext` instances.

Proposal for pika 1.0.0:
1. Modify `pika.connection.SSLOptions` constructor to accept only a non-None `ssl.SSLContext` instance and an optional `server_hostname` arg
2. `server_hostname` defaults to None and is used for CN validation - see discussion in issue #464.

vitaly-krugl:
Closes #986

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -3,7 +3,6 @@
 # pylint: disable=C0302
 
 import ast
-import sys
 import collections
 import copy
 import logging
@@ -13,11 +12,6 @@ import platform
 import socket
 import warnings
 import ssl
-
-if sys.version_info > (3,):
-    import urllib.parse as urlparse  # pylint: disable=E0611,F0401
-else:
-    import urlparse
 
 from pika import __version__
 from pika import callback as pika_callback
@@ -29,6 +23,7 @@ from pika import heartbeat as pika_heartbeat
 
 from pika import spec
 
+import pika.compat
 from pika.compat import (xrange, basestring, # pylint: disable=W0622
                          url_unquote, dictkeys, dict_itervalues,
                          dict_iteritems)
@@ -519,13 +514,14 @@ class Parameters(object):  # pylint: disable=R0902
     def tcp_options(self):
         """
         :returns: None or a dict of options to pass to the underlying socket
+        :rtype: dict|None
         """
         return self._tcp_options
 
     @tcp_options.setter
     def tcp_options(self, value):
         """
-        :param bool value: None or a dict of options to pass to the underlying
+        :param dict|None value: None or a dict of options to pass to the underlying
             socket. Currently supported are TCP_KEEPIDLE, TCP_KEEPINTVL, TCP_KEEPCNT
             and TCP_USER_TIMEOUT. Availability of these may depend on your platform.
         """
@@ -759,7 +755,7 @@ class URLParameters(Parameters):
 
         # TODO Is support for the alternative http(s) schemes intentional?
 
-        parts = urlparse.urlparse(url)
+        parts = pika.compat.urlparse(url)
 
         if parts.scheme == 'https':
             # Create default context which will get overridden by the
@@ -792,7 +788,7 @@ class URLParameters(Parameters):
 
         # Handle query string values, validating and assigning them
 
-        self._all_url_query_values = urlparse.parse_qs(parts.query)
+        self._all_url_query_values = pika.compat.url_parse_qs(parts.query)
 
         for name, value in dict_iteritems(self._all_url_query_values):
             try:
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -709,8 +709,10 @@ class URLParameters(Parameters):
         - locale:
             Override the default `en_US` locale value
         - ssl_options:
-            None for plaintext or dict of arguments passed to
-            :meth:`ssl.wrap_socket` as kwargs
+            None for plaintext; for SSL: dict of public ssl context-related
+            arguments that may be passed to :meth:`ssl.SSLSocket` as kwargs,
+            except `sock`, `server_side`,`do_handshake_on_connect`, `family`,
+            `type`, `proto`, `fileno`.
         - retry_delay:
             The number of seconds to sleep before attempting to connect on
             connection failure.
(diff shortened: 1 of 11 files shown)


vitaly-krugl:
cc @lukebakken

pika/connection.py
```diff
@@ -947,66 +906,60 @@ def _set_url_socket_timeout(self, value):
...
+        if options is None:
+            if self.ssl_options is not None:
+                raise ValueError(
+                    'Specified ssl_options=None URL arg is inconsistent with '
+                    'the specified https URL scheme.')
+        else:
+            # Convert options to pika.SSLOptions via ssl.SSLSocket()
+            sock = socket.socket()
+            try:
+                ssl_sock = ssl.SSLSocket(sock=sock, **options)
```
vitaly-krugl: `ssl.SSLSocket` supports the important `server_hostname` arg that `ssl.wrap_socket` doesn't.

vitaly-krugl:
Ready for review.

cc @lukebakken

lukebakken:
@vitaly-krugl awesome, I'll review today or tomorrow

michaelklishin:
@vitaly-krugl thank you for your ongoing (and seemingly never ending :)) contributions!

vitaly-krugl:
On a related note, what's the typical use case for URLParameters? Is this configuration style widely used? The reason I ask is if you need to throw in things like `ssl_options`, then the nesting makes the resulting query args pretty messy for someone to compose by hand. For example, this set or parameters represented as python dict:

```
{'blocked_connection_timeout': 10.5,
 'channel_max': 3,
 'connection_attempts': 2,
 'frame_max': 40000,
 'heartbeat': 7,
 'locale': 'en_UK',
 'retry_delay': 3,
 'socket_timeout': 100.5,
 'ssl_options': {'ca_certs': None,
  'cert_reqs': 0,
  'certfile': None,
  'ciphers': None,
  'keyfile': None,
  'npn_protocols': None,
  'server_hostname': 'blah.blah.com',
  'ssl_version': 2},
 'tcp_options': {'TCP_KEEPIDLE': 60, 'TCP_USER_TIMEOUT': 1000}}

... looks like this when represented as query args; note the ton of escaping that deals with the nested dicts towards the end:
```
In [50]: urlencode(qa)
Out[50]: 'blocked_connection_timeout=10.5&channel_max=3&connection_attempts=2&frame_max=40000&heartbeat=7&locale=en_UK&retry_delay=3&socket_timeout=100.5&ssl_options=%7B%27keyfile%27%3A+None%2C+%27certfile%27%3A+None%2C+%27ssl_version%27%3A+2%2C+%27ca_certs%27%3A+None%2C+%27cert_reqs%27%3A+0%2C+%27npn_protocols%27%3A+None%2C+%27ciphers%27%3A+None%2C+%27server_hostname%27%3A+%27blah.blah.com%27%7D&tcp_options=%7B%27TCP_USER_TIMEOUT%27%3A+1000%2C+%27TCP_KEEPIDLE%27%3A+60%7D'
```

Is this really convenient/readable for users? Wouldn't something like JSON or YAML be less error-prone, more readable, and easier for users to maintain?

**Comment by michaelklishin** [comment]
Shovel, Federation and some clients use URI parameters for [a subset of connection options](http://www.rabbitmq.com/uri-spec.html). It's not strictly a requirement.

**Comment by vitaly-krugl** [comment]
> Shovel, Federation and some clients use URI parameters ...

Thank you for the link

File: pika/connection.py [hunk 2]
```diff
@@ -947,66 +902,60 @@ def _set_url_socket_timeout(self, value):
...
-        self.cafile = cafile
-        self.capath = capath
-        self.cadata = cadata
-        self.do_handshake_on_connect = do_handshake_on_connect
-        self.suppress_ragged_eofs = suppress_ragged_eofs
-        self.ciphers = ciphers
+    # Protect against accidental assignment of an invalid attribute
+    __slots__ = ('context', 'server_hostname')
+
+    def __init__(self, context, server_hostname=None):
```
lukebakken: We're keeping `SSLOptions` around just for `server_hostname`. We could get rid of this class and add `ssl_server_hostname` to `ConnectionParameters`. Just an idea.

vitaly-krugl:
I prefer grouping information that belongs together in an object. This way
we have a natural place (i.e., a namespace) to add other bits in the
future, it facilitates clean consistency checking, and it also eliminates
some confusing situations, such as the user specifying the server hostname
without the SSL context, making it easier to understand and less
error-prone.

As for the URL case, flattening the query args might make it simpler for
someone to use it. However, I haven't seen complaints yet, and this stuff
has been this way for a long time. So, either no one really cares about
using URL Parameters with pika or they are fine with the status quo. :)

On Mar 6, 2018 10:55 AM, "Luke Bakken" <notifications@github.com> wrote:

*@lukebakken* commented on this pull request.
------------------------------

In pika/connection.py
<https://github.com/pika/pika/pull/987#discussion_r172621853>:

> -        self.keyfile = keyfile
-        self.key_password = key_password
-        self.certfile = certfile
-        self.server_side = server_side
-        self.verify_mode = verify_mode
-        self.ssl_version = ssl_version
-        self.cafile = cafile
-        self.capath = capath
-        self.cadata = cadata
-        self.do_handshake_on_connect = do_handshake_on_connect
-        self.suppress_ragged_eofs = suppress_ragged_eofs
-        self.ciphers = ciphers
+    # Protect against accidental assignment of an invalid attribute
+    __slots__ = ('context', 'server_hostname')
+
+    def __init__(self, context, server_hostname=None):

We're keeping SSLOptions around just for server_hostname. We could get rid
of this class and add ssl_server_hostname to ConnectionParameters. Just an
idea.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/pika/pika/pull/987#pullrequestreview-101670756>, or mute
the thread

vitaly-krugl:
Thank you @lukebakken!

Issue #912, opened 2018-01-03
BlockingConnection.close() hangs on Exception treatment


galindro:
The console hangs if you close a BlockingConnection on treating an exception triggered by exchange_declare() method from channel instance. To exit the program, you need to press Ctrl+C or close the connection from RabbitMQ Server.

## Env. Info:
* Pika version: 0.11.2
* Python version: 3.6.4
* RabbitMQ version: rabbitmq:3.7.2-management-alpine (docker image)

## How to reproduce
1. Start a rabbitmq docker container 
```bash
docker run -it --rm -p 15672:15672 -p 5672:5672 rabbitmq:3.7.2-management-alpine
```

2. Run the bellow script:
```python
import pika
import logging
import sys

logger = logging.getLogger()
h = logging.StreamHandler(sys.stdout)
FORMAT = f'%(asctime)s - %(levelname)s: "%(message)s"'
h.setFormatter(logging.Formatter(FORMAT))
logger.addHandler(h)
logger.setLevel("DEBUG")

rmq_credentials = pika.PlainCredentials('guest', 'guest')
rmq_params = pika.ConnectionParameters(host='localhost',
                                       port='5672',
                                       credentials=rmq_credentials,
                                       blocked_connection_timeout=5)
rmq_connection = pika.BlockingConnection(rmq_params)
rmq_channel = rmq_connection.channel()

try:
  rmq_channel.exchange_declare(exchange=1, exchange_type='topic', internal=True, durable=True)
except Exception as e:
  rmq_connection.close()
```

3. You should see the bellow logs and the console will hang. To exit, you will need to press Ctrl+C
```bash
2018-01-03 16:52:00,481 - DEBUG: "Using EPollPoller"
2018-01-03 16:52:00,481 - DEBUG: "Added: {'callback': <bound method Connection._on_connection_start of <SelectConnection CLOSED socket=None params=<ConnectionParameters host=localhost port=5672 virtual_host=/ ssl=False>>>, 'one_shot': True, 'only': None, 'arguments': None, 'calls': 1}"
2018-01-03 16:52:00,482 - DEBUG: "Added: {'callback': <bound method Connection._on_connection_close of <SelectConnection CLOSED socket=None params=<ConnectionParameters host=localhost port=5672 virtual_host=/ ssl=False>>>, 'one_shot': True, 'only': None, 'arguments': None, 'calls': 1}"
[...]

PR #991, opened 2018-03-07
Request marshaling error should not corrupt a channel

### LINKED ISSUE #990: BlockingChannel.queue_declare hanging on non-string queue parameters
Under Python 3.6.4 and Pika 0.11.2, the `BlockingChannel.queue_declare` method hangs when setting its `queue` parameter to a value that is not of `str` type (e.g., `int`, `bool`, `list`, `dict`, `tuple`, `NoneType`).

Input:

```
$ python3 <<EOF
import pika
with pika.BlockingConnection() as connection:
    channel.queue_declare(queue=[1, 2, 3])
EOF
```

lukebakken:
Fixes #990 and #912

### DIFF
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -1152,7 +1152,7 @@ class Channel(object):
             # a meta-close (see `_on_close_meta()` which fakes receipt of
             # `Channel.CloseOk` and dispatches the `'_on_channel_close'`
             # callbacks.
-            self._drain_blocked_methods_on_close()
+            self._drain_blocked_methods_on_remote_close()
         else:
             self._transition_to_closed()
 
@@ -1330,10 +1330,9 @@ class Channel(object):
         while self._blocked and self._blocking is None:
             self._rpc(*self._blocked.popleft())
 
-    def _drain_blocked_methods_on_close(self):
+    def _drain_blocked_methods_on_remote_close(self):
         """This is called when the broker sends a Channel.Close while the
-        client is in CLOSING state, or when a local Channel.Close is requested
-        while self._blocking is true. This method checks the blocked method
+        client is in CLOSING state. This method checks the blocked method
         queue for a pending client-initiated Channel.Close method and
         ensures its callbacks are processed, but does not send the method
         to the broker.
@@ -1348,7 +1347,7 @@ class Channel(object):
         sent, and thus its completion callback would never be called.
 
         """
-        LOGGER.debug('Draining %i blocked frames due to Channel.Close',
+        LOGGER.debug('Draining %i blocked frames due to broker-requested Channel.Close',
                      len(self._blocked))
         while self._blocked:
             method = self._blocked.popleft()[0]
@@ -1401,13 +1400,6 @@ class Channel(object):
         if self.is_closed:
             self._raise_if_not_open()
 
-        if self.is_closing and self._blocking:
-            LOGGER.debug(
-                'In blocking state but close requested, '
-                'so discarding blocked method %s', self._blocking)
-            self._blocking = None
-            self._drain_blocked_methods_on_close()
-
         # If the channel is blocking, add subsequent commands to our stack
         if self._blocking:
             LOGGER.debug('Already in blocking state, so enqueueing method %s; '
@@ -1417,17 +1409,10 @@ class Channel(object):
             return
 
         # Note: _send_method can throw exceptions if there are framing errors
-        # or invalid data passed in. Call it here, before the acceptable_replies
-        # block to prevent callbacks from being registered if an exception is
-        # thrown
-        try:
-            self._send_method(method)
-        except Exception as err:
-            if self._blocking:
-                LOGGER.error(
-                    "send_method failed for blocking method %s with %s, will discard",
-                    self._blocking, type(err).__name__)
-            raise
+        # or invalid data passed in. Call it here to prevent self._blocking
+        # from being set if an exception is thrown. This also prevents
+        # acceptable_replies registering callbacks when exceptions are thrown
+        self._send_method(method)
 
         # If acceptable replies are set, add callbacks
         if acceptable_replies:
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -1152,7 +1152,7 @@ class Channel(object):
             # a meta-close (see `_on_close_meta()` which fakes receipt of
             # `Channel.CloseOk` and dispatches the `'_on_channel_close'`
             # callbacks.
-            self._drain_blocked_methods_on_remote_close()
+            self._drain_blocked_methods_on_close()
         else:
             self._transition_to_closed()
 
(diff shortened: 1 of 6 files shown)


vitaly-krugl:
@lukebakken, I haven't forgotten - will catch up in the next couple of days.

lukebakken:
@vitaly-krugl no hurry at all! Thanks again.

lukebakken:
@vitaly-krugl the `queue.declare` method never makes it to RabbitMQ. An `AssertionError` is thrown [here](https://github.com/pika/pika/blob/master/pika/spec.py#L1003-L1004) which gums up the works when the `with` clause tries to exit.

vitaly-krugl:
Would this alone punch at the heart of the problem [here](https://github.com/pika/pika/blob/b7f27983cfbcbaf34a06b6fc9259a7fd50b8838d/pika/channel.py#L1382)?
```
try:
    self._send_method(method)
except Exception:
    self._blocking = None
    raise
```

lukebakken:
I'll try that out.

vitaly-krugl:
Don't try it out just yet, I missed something there (and the fix in the PR did, too, I think)

vitaly-krugl:
The [if acceptable_replies:](https://github.com/pika/pika/blob/b7f27983cfbcbaf34a06b6fc9259a7fd50b8838d/pika/channel.py#L1361) block sets `self._blocking` and also registers a number of callbacks. If `self._send_method(method)` (as in this case), we really don't want any of those registered callbacks to remain there either.

Since a successful `self._send_method(method)` call will ultimately just enqueue some data on the output write buffer, it should be possible to move the (https://github.com/pika/pika/blob/b7f27983cfbcbaf34a06b6fc9259a7fd50b8838d/pika/channel.py#L1361) block after [self._send_method(method)](https://github.com/pika/pika/blob/b7f27983cfbcbaf34a06b6fc9259a7fd50b8838d/pika/channel.py#L1382).

Furthermore, to ensure that an incomplete message doesn't get placed in the output buffer (due to marshaling failure of one of its subframes), [Connection._send_message()](https://github.com/pika/pika/blob/b7f27983cfbcbaf34a06b6fc9259a7fd50b8838d/pika/connection.py#L2273) needs to be modified to pre-marshal all of its parts and then append them to the output frame buffer only after all marshaling is done, updating the stats and finally calling `self._flush_outbound() ` and `self._detect_backpressure()` like [Connection._send_frame()](https://github.com/pika/pika/blob/b7f27983cfbcbaf34a06b6fc9259a7fd50b8838d/pika/connection.py#L2251-L2257). To this end,  `Connection._send_message()` and `Connection._send_frame()` should share a method (e.g.,  `Connection._output_marshaled_frame()` that updates `self.bytes_sent` and `self.frames_sent` and appends the marshaled frame data to `self.outbound_buffer` .

lukebakken:
@vitaly-krugl - ready for re-review. Thanks!

lukebakken:
@vitaly-krugl - I have merged in the tests you provided and this is ready for another review. Thanks!

vitaly-krugl:
@lukebakken, I renamed this PR "Request marshaling error should not corrupt a channel", which reflects issues #990 and #912 more accurately.

pika/channel.py
```diff
@@ -1397,6 +1397,12 @@ def _rpc(self, method, callback=None, acceptable_replies=None):
         if self.is_closed:
             self._raise_if_not_open()
 
+        if self.is_closing and self._blocking:
+            # Note: This is normally called when a synchronous command is
+            # completed. It will undo the blocking state and send all the
+            # frames that stacked up while we were in the blocking state.
+            self._on_synchronous_complete(None)
```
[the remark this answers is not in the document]
vitaly-krugl: The problem we're trying to fix is "Request marshaling error should not corrupt a channel". The only reason the channel becomes corrupted is the marshaling issue, which the rest of this PR addresses.

pika/channel.py
```diff
@@ -1327,9 +1327,10 @@ def _on_synchronous_complete(self, _method_frame_unused):
         while self._blocked and self._blocking is None:
             self._rpc(*self._blocked.popleft())
 
-    def _drain_blocked_methods_on_remote_close(self):
```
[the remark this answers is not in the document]
lukebakken: > If we think we need an "emergency channel-close" method that purges queued-up requests (I don't think we do

Draining blocked methods on a broker-initiated close was introduced in #957 - please check that PR out again. I still think it's necessary.

vitaly-krugl:
On broker's Channel.Close, the draining is necessary because ANQP says to
ignore all incoming  requests after  channel is closed except
Channel.Close. So, draining in that case helps break the gridlock.

However, in the case the client is closing the channel with some blocking
requests still pending normally, we have a perfectly healthy channel and
nothing special is needed. The normal course of events will see it through.

vitaly-krugl:
@lukebakken, is this PR ready for re-review?

vitaly-krugl:
I think I might not have ended that review ??

lukebakken:
@vitaly-krugl thanks! Sorry I missed the previous comment about that test.

tests/acceptance/blocking_adapter_test.py
```diff
@@ -271,6 +270,16 @@ def test(self):
         self.assertFalse(ch._impl._consumers)
 
 
+class TestUsingInvalidQueueArgument(BlockingTestCaseBase):
+    def test(self):
+        """BlockingConnection raises expected connection when invalid queue parameter is used
```
vitaly-krugl: "expected connection" ==> "expected exception"

tests/acceptance/blocking_adapter_test.py
```diff
@@ -271,6 +270,16 @@ def test(self):
...
+        """
+        connection = self._connect()
+        ch = connection.channel()
+        with self.assertRaises(AssertionError):
```
vitaly-krugl: Surprising that the underlying implementation raises AssertionError instead of TypeError/ValueError. But no need to fix that.

pika/connection.py
```diff
@@ -2266,8 +2261,14 @@ def _send_message(self, channel_number, method_frame, content):
 
         """
         length = len(content[1])
-        self._send_frame(frame.Method(channel_number, method_frame))
-        self._send_frame(frame.Header(channel_number, length, content[0]))
+        marshaled_body_frames = []
+
+        # Note: we construct the Method and Header objects, marshal them
```
vitaly-krugl: ... Method, Header, and Content objects, ...

lukebakken:
@vitaly-krugl - changes made. I'll merge this once builds complete. Thanks!

lukebakken:
@vitaly-krugl I made your latest suggested changes and rebased against `master`. I'm waiting on the builds at the moment. Thanks.

lukebakken:
Thanks @vitaly-krugl

PR #993, opened 2018-03-12
(Bug fix) Wrong condition check when trying to set self.virtual_host

daryasary:
Wrong condition check when trying to set self.virtual_host in URLParameters, this will fix "list index out of range" error.

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -784,7 +784,7 @@ class URLParameters(Parameters):
                                                                  url_unquote(parts.password))
 
         # Get the Virtual Host
-        if len(parts.path) > 1:
+        if len(parts.path.split('/')) > 1:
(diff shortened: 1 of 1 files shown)


lukebakken:
Hello! Would you mind also creating a test for this?

lukebakken:
What URL did you use to trigger the `list index out of range` error?

lukebakken:
Aha, your URL must have been like this -

```
>>> urlparse("amqp://foo.bar?baz&bat").path.split('/')[1]
IndexError: list index out of range
```

lukebakken:
I can't reproduce the `list index out of range` error. If you could please provide the Python and Pika version you're using, as well as a way to reproduce this error, that would be appreciated.

PR #984, opened 2018-03-02
Enable TLS/SSL testing in Travis CI and AppVeyor

lukebakken:
Fixes pika/pika#981

PR #994, opened 2018-03-12
Add a few more URLParameters tests

lukebakken:
I tried to reproduce what was reported in #993, but couldn't, but added a few more tests.

### DIFF
--- a/tests/unit/connection_parameters_tests.py
+++ b/tests/unit/connection_parameters_tests.py
@@ -695,6 +695,34 @@ class URLParametersTests(_ParametersTestsBase):
         self.assertEqual(parameters.virtual_host,
                          pika.URLParameters.DEFAULT_VIRTUAL_HOST)
 
+    def test_uses_default_virtual_host_via_encoded_slash_upcase(self):
+        parameters = pika.URLParameters('amqp://myserver.mycompany.com/%2F')
+        self.assertEqual(parameters.virtual_host,
+                         pika.URLParameters.DEFAULT_VIRTUAL_HOST)
+
+    def test_uses_default_virtual_host_via_encoded_slash_downcase(self):
+        parameters = pika.URLParameters('amqp://myserver.mycompany.com/%2f')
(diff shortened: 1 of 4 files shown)


tests/unit/connection_parameters_tests.py
```diff
@@ -695,6 +695,34 @@ def test_uses_default_virtual_host_if_only_slash_is_specified(self):
         self.assertEqual(parameters.virtual_host,
                          pika.URLParameters.DEFAULT_VIRTUAL_HOST)
 
+    def test_uses_default_virtual_host_via_encoded_slash_1(self):
+        parameters = pika.URLParameters('amqp://myserver.mycompany.com/%2F')
+        self.assertEqual(parameters.virtual_host,
+                         pika.URLParameters.DEFAULT_VIRTUAL_HOST)
+
+    def test_uses_default_virtual_host_via_encoded_slash_2(self):
```
vitaly-krugl: I think it may be better to use more descriptive test names instead of suffixing with 1,2,3,4. It seems like the variability is upper case vs lower case encoding and terminating slash versus no terminating slash.

vitaly-krugl:
> I tried to reproduce what was reported in #993, but couldn't

@daryasary, could you please provide a simple example that reproduces the issue that you tried to fix with #993?

vitaly-krugl:
@lukebakken, could you please take a look at the software installation failures that caused a number of the build jobs to fail? This may be related to the changes in PR #984. Perhaps software installations could use some hardening or a faster source repository.

lukebakken:
It looks like Erlang Solutions' package repository is slow.

vitaly-krugl:
> It looks like Erlang Solutions' package repository is slow

Is there a cache site that might be faster, more reliable?

lukebakken:
I'll try using the default Erlang package.

lukebakken:
The default Erlang package for Ubuntu Trusty is too old to use (`R16B03`) with the latest RabbitMQ.

PR #996, opened 2018-03-14
Fixed leaking of ChannelClosed exception back through the callback from SelectConnection that notified BlockingChannel about broker-initiated Channel.Close

vitaly-krugl:
This tests changes from pull request #886 that stop event processing and raise ChannelClosed upon channel failure (i.e., when Channel.Close is received from broker).

Also test `BlockingChannel.start_consuming()` and `BlockingChannel.consume()` encounters of Cancel from broker.

Fixed leaking of ChannelClosed exception (introduced in #886) back through the callback from SelectConnection that notified BlockingChannel about broker-initiated `Channel.Close` and on through the I/O loop..

# NOTABLE BEHAVIOR/API CHANGES:

1. ChannelClosed due to broker-initiated Channel.Close  is now raised for a channel only when the user interacts with that channel by invoking that channel's API that leads to I/O loop activity (e.g., sending a request to broker, running the BlockingChannel.consume() generator, running in BlockingChannel.start_consuming(). Simply calling BlockingConnection.sleep() won't raise ChannelClosed.
2. Added explicit `reply_code` and `reply_text` constructor args to the `ChannelClosed` exception class and provided corresponding `reply_code` and `reply_text` properties to retrieve them.
3. Standardized raising of the ChannelClosed exception via `Channel._raise_if_not_open()` that supplies reasonable `reply_code` and `reply_text` args based on current channel state to the `ChannelClosed` exception.
4. Added class `pika.channel. ClientChannelErrors` that defines client-side channel reply codes for use in `ChannelClosed.reply_code`.
5. When dispatching `'_on_channel_close'` callbacks after user-initiated closing of the channel, pass the `reply_code` and `reply_text` that were provided by user to `Channel.close()` and `BlockingChannel.close()` instead of the dummy `(0, '')`, which was counter to the original intention and documentation described [here](https://github.com/pika/pika/blob/b7f27983cfbcbaf34a06b6fc9259a7fd50b8838d/pika/channel.py#L85-L87).
6. Implemented new public property `Channel.is_closed_by_broker` that may be used in conjunction with `Channel.is_closed` to glean the source of the channel's closing.

### DIFF
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -20,6 +20,19 @@ LOGGER = logging.getLogger(__name__)
 MAX_CHANNELS = 65535  # per AMQP 0.9.1 spec.
 
 
+class ClientChannelErrors(object):
+    """Client-side channel error codes using negative values to avoid collision
+     with AMQP-defined reply-code values. This values may be used as
+     `reply_code` in `ChannelClosed` exception.
+
+     """
+    # Our codes will be -1001, -1002, -1003, etc. to avoid collision
+    # with `Connection`-defined internal client error codes
+    FIRST = -1000
+    OPEN_PENDING = -1001  # channel is not usable yet; wait until it opens
+    CLOSE_PENDING = -1002  # channel is closing
+
+
 class Channel(object):
     """A Channel is the primary communication method for interacting with
     RabbitMQ. It is recommended that you do not directly invoke
@@ -81,13 +94,19 @@ class Channel(object):
         self._on_openok_callback = on_open_callback
         self._state = self.CLOSED
 
+        # Will be set to True upon receipt of `Channel.Close` from broker.
+        self._closed_by_broker = False
+
         # We save the closing reason code and text to be passed to
         # on-channel-close callback at closing of the channel. Channel.close
         # stores the given reply_code/reply_text if the channel was in OPEN or
         # OPENING states. An incoming Channel.Close AMQP method from broker will
         # override this value. And a sudden loss of connection has the highest
-        # prececence to override it.
-        self._closing_code_and_text = (0, '')
+        # precedence to override it.
+        #
+        # When initiating or receving 'Channel.Close' from broker, set to
+        # the two-tuple: (int reply_code, str reply_text)
+        self._closing_code_and_text = None
 
         # opaque cookie value set by wrapper layer (e.g., BlockingConnection)
         # via _set_cookie
@@ -202,8 +221,7 @@ class Channel(object):
                               acknowledgement of all outstanding messages.
 
         """
-        if not self.is_open:
-            raise exceptions.ChannelClosed()
+        self._raise_if_not_open()
         return self._send_method(spec.Basic.Ack(delivery_tag, multiple))
 
     def basic_cancel(self, consumer_tag='', callback=None):
@@ -225,7 +243,7 @@ class Channel(object):
         :raises ValueError:
 
         """
-        self._validate_channel()
+        self._raise_if_not_open()
         nowait = self._validate_rpc_completion_callback(callback)
 
         if consumer_tag in self._cancelled:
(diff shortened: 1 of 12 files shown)


vitaly-krugl:
@lukebakken - I implemented a new test that detected unwanted/dangerous behavior in BlockingConnection concerning leaking exceptions back into the I/O loop/SelectConnection.

As the result, this PR's tests are now failing.

More info: BlockingConnection's underlying asynchronous connection adapter (SelectConnection) uses callbacks to communicate with its user (BlockingConnection in this case). If BlockingConnection leaks exceptions back into the I/O loop or the asynchronous connection adapter, it interrupts their normal workflow and introduces a high likelihood of state inconsistency within those components.

pika/channel.py
```diff
@@ -225,7 +243,7 @@ def basic_cancel(self, consumer_tag='', callback=None):
         :raises ValueError:
 
         """
-        self._validate_channel()
+        self._raise_if_not_open()
```
vitaly-krugl: @lukebakken, here I renamed the method to make it explicit/self-documenting. The old name `_validate_channel` sounded more like a consistency-checker for the channel.

tests/acceptance/blocking_adapter_test.py
```diff
@@ -205,7 +251,7 @@ def test(self):
 
         # Declare a new queue
         ch.queue_declare(q_name, auto_delete=True)
-        self.addCleanup(self._connect().channel().queue_delete, q_name)
+        self.addCleanup(lambda: self._connect().channel().queue_delete(q_name))
```
vitaly-krugl: I changed these to lambda when I discovered that tests that fail with timeout also fail during cleanup - exploding the nosetests log size and creating unnecessary confusion during failure analysis - since the timeout timer in these addCleanup `self._connect()` calls also starts ticking at the same time as the connection used in each test.

tests/unit/channel_tests.py
```diff
@@ -330,7 +334,7 @@ def test_basic_consume_consumers_rpc_with_no_completion_callback(self, rpc, _unu
         expectation = spec.Basic.Consume(
             queue='test-queue',
             consumer_tag=consumer_tag,
-            auto_ack=False,
+            no_ack=False,
```
vitaly-krugl: The programmatically-generated code in pika/spec.py uses the AMQP `no_ack` versus the new name `auto_ack` that we adopted recently in the connection adapters.

tests/unit/channel_tests.py
```diff
@@ -590,6 +601,7 @@ def test_confirm_delivery_with_bad_callback_raises_value_error(self):
                           'bad-callback')
 
     def test_confirm_delivery_raises_channel_closed(self):
+        self.obj._closing_code_and_text = (123, "here is why it was closed")
```
vitaly-krugl: Need to set these up since `Channel. _closing_code_and_text` now defaults to `None`.

tests/unit/channel_tests.py
```diff
@@ -549,7 +559,7 @@ def test_close_in_opening_state(self, _unused):
             spec.Channel.Close(200, 'Got to go', 0, 0), self.obj._on_closeok,
             [spec.Channel.CloseOk])
 
-        self.assertEqual(self.obj._closing_code_and_text, (0, ''))
+        self.assertEqual(self.obj._closing_code_and_text, (200, 'Got to go'))
```
vitaly-krugl: `Channel.close()` member function now saves the passed-in values in `_closing_code_and_text`, so comparing against the bogus `(0, '')` no longer works.

vitaly-krugl:
Ready for review.

cc @lukebakken

vitaly-krugl:
Had a transient timeout running "Create and delete and exchange (AsyncioConnection)" test over SSL and python 3.7-dev on travis-ci: https://travis-ci.org/pika/pika/jobs/354112927

lukebakken:
@vitaly-krugl I'll take a look again today. Thanks!

vitaly-krugl:
Thanks @lukebakken!

PR #1010, opened 2018-04-06
Don't attempt graceful AMQP Connection.Close upon heartbeat timeout

vitaly-krugl:
cc @lukebakken 

Once heartbeat logic concludes that the connection is dead due to missed heartbeats, there is no point attempting a graceful Connection.Close, including Channel.Close requests if there were open channels.


michaelklishin:
The idea makes sense. Good catch, @vitaly-krugl!

PR #1011, opened 2018-04-10
Add support for type "x"

masell:
Sender is java spring application, and on python side i get this exception:

```
  File "/usr/lib/python3.6/asyncio/events.py", line 127, in _run
    self._callback(*self._args)
  File ".venv/lib/python3.6/site-packages/pika/adapters/base_connection.py", line 395, in _handle_events
    self._handle_read()
  File ".venv/lib/python3.6/site-packages/pika/adapters/base_connection.py", line 449, in _handle_read
    self._on_data_available(data)
  File ".venv/lib/python3.6/site-packages/pika/connection.py", line 1934, in _on_data_available
    consumed_count, frame_value = self._read_frame()
  File ".venv/lib/python3.6/site-packages/pika/connection.py", line 2080, in _read_frame
    return frame.decode_frame(self._frame_buffer)
  File ".venv/lib/python3.6/site-packages/pika/frame.py", line 250, in decode_frame
    out = properties.decode(frame_data[12:])
  File ".venv/lib/python3.6/site-packages/pika/spec.py", line 2112, in decode
    (self.headers, offset) = data.decode_table(encoded, offset)
  File ".venv/lib/python3.6/site-packages/pika/data.py", line 186, in decode_table
    value, offset = decode_value(encoded, offset)
  File ".venv/lib/python3.6/site-packages/pika/data.py", line 308, in decode_value
    raise exceptions.InvalidFieldTypeException(kind)
pika.exceptions.InvalidFieldTypeException: b'x'
```

Docs here:
https://www.rabbitmq.com/releases/rabbitmq-dotnet-client/v3.6.0/rabbitmq-dotnet-client-3.6.0-client-htmldoc/html/type-RabbitMQ.Client.BinaryTableValue.html

### DIFF
--- a/tests/unit/data_tests.py
+++ b/tests/unit/data_tests.py
@@ -48,6 +48,15 @@ class DataTests(unittest.TestCase):
             ('bytes', b'foobar'),
         ])
 
+    def test_decode_bytes(self):
+        input = (
+                b'\x00\x00\x00\x01'
+                b'\x05bytesx\x00\x00\x00\x06foobar'
+            )
+        result = data.decode_table(input, 0)
+        self.assertEqual(result, ({'bytes': b'foobar'}, 21))
+
+
     def test_encode_table(self):
         result = []
         data.encode_table(result, self.FIELD_TBL_VALUE)


lukebakken:
Thank you! Could you please include a test?

tests/unit/data_tests.py
```diff
@@ -15,7 +15,7 @@
 class DataTests(unittest.TestCase):
 
     FIELD_TBL_ENCODED = (
-        b'\x00\x00\x00\xcb'
+        b'\x00\x00\x00\xd9'
```
masell: I dont know what this is...

masell:
@lukebakken Added test case, and encode also (Needed for test.. :) )

masell:
@lukebakken The coverage job is failing due to some credentials stuff... I can't fix that i assume?

lukebakken:
@masell - yep, don't worry about that! I'll review this PR within a week. Thanks again.

tests/unit/data_tests.py
```diff
@@ -7,40 +7,48 @@
...
-        b'\x0bdecimal_tooD\x00\x00\x00\x00d'
-        b'\x07dictvalF\x00\x00\x00\x0c\x03fooS\x00\x00\x00\x03bar'
-        b'\x06intvalI\x00\x00\x00\x01'
-        b'\x06bigint\x6c\x00\x00\x00\x00\x9a\x7e\xc8\x00'
-        b'\x07longval\x6c\x00\x00\x00\x00\x36\x65\x26\x55'
-        b'\x04nullV'
-        b'\x06strvalS\x00\x00\x00\x04Test'
-        b'\x0ctimestampvalT\x00\x00\x00\x00Ec)\x92'
-        b'\x07unicodeS\x00\x00\x00\x08utf8=\xe2\x9c\x93')
+    FIELD_TBL_ENCODED = b"".join((
```
vitaly-krugl: Although more elegant, changing all the existing lines makes it difficult to review.

masell:
@vitaly-krugl reverted the tbl_encoded.

masell:
Please note that in python 2.X, a byte (type X) is a str as they do not distinguish the types. while in 3.X its as bytes b"". This the reason for checking if its PY3 in the test.

```
Python 2.7.14 (default, Sep 23 2017, 22:06:14) 
[GCC 7.2.0] on linux2
>>> b"a" == u"a"
True
```

```
Python 3.6.3 (default, Oct  3 2017, 21:45:48) 
[GCC 7.2.0] on linux
>>> b"a" == u"a"
False
```

hairyhum:
@masell it looks like there is no test to check the `x` type parsing with 2.7. Can you add one?

masell:
@hairyhum Same test, different fixture for 2.7

https://github.com/pika/pika/pull/1011/files#diff-6d92c2a692fc77fd16dcd1bc9b922392R35

hairyhum:
@masell yes, but it uses type `S`, not `x`. Or do I miss something?

masell:
@hairyhum 
Added new test for decoding.
https://github.com/pika/pika/pull/1011/commits/fca2ec9b6a04273a3a1d7102d2d3c36d100b7281

lukebakken:
@vitaly-krugl - good point. @masell links to [this document](https://www.rabbitmq.com/releases/rabbitmq-dotnet-client/v3.6.0/rabbitmq-dotnet-client-3.6.0-client-htmldoc/html/type-RabbitMQ.Client.BinaryTableValue.html) which mentions this non-standard extension:

> The sole reason for the existence of this class is to permit encoding of byte[] as 'x' in AMQP field tables, an extension to the specification that is part of the tentative JMS mapping implemented by QPid

@masell also says that the sending application he is using is a "java spring application". It looks as though the [RabbitMQ Java client](https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/ValueWriter.java#L190-L194) supports the `x` encoding as well: [link](https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/ValueWriter.java#L190-L194)

Type `x` is also supported in the Ruby `amq-protocol` project: [link](https://github.com/ruby-amqp/amq-protocol/blob/master/lib/amq/protocol/type_constants.rb#L18).

It seems valid to add it to Pika as well, IMHO.

vitaly-krugl:
@lukebakken, so what's the semantic difference between `byte[]` as represented by 'x' and `raw byte[]` as represented by 'S' ?

masell:
@vitaly-krugl 
"S" is string.
"x" is byte[].

"S" is byte[] on the wire, that is unicode encoded. Decoded would be str() in python3/2, System.String in java.
"x" is byte[] with no encoding, a data blob. Can even be a JPEG if you wish. Decoded would be a bytes() in python3, a byte[] in java. Python2 lacks a byte representation AFAIK, so its the same as str in python2.

vitaly-krugl:
@lukebakken, would you mind completing this review? Thx!

lukebakken:
@vitaly-krugl I'll get it today. I've had a cold this week so things are going slowly.

lukebakken:
@vitaly-krugl :+1: from me, could you update your review?

vitaly-krugl:
@lukebakken  - done!

PR #1016, opened 2018-04-13
Use Travis CI RabbitMQ 

lukebakken:
See travis-ci/travis-ci#6302

I have contacted both Erlang Solutions and Travis CI about this. I'm not sure what's up and I can't reproduce the issue using my local Ubuntu Trusty VMs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,24 @@ language: python
 
 sudo: false
 
-addons:
-  apt:
-    sources:
-      - sourceline: deb https://packages.erlang-solutions.com/ubuntu trusty contrib
-        key_url: https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc
-    packages:
-      - erlang-nox
+services:
+  - rabbitmq
+
+# TODO:
+# set 'sudo: false' and uncomment once "Hash Sum mismatch" failure in 'apt-get update' is fixed
+# addons:
+#   apt:
+#     sources:
+#       - sourceline: deb https://packages.erlang-solutions.com/ubuntu trusty contrib
+#         key_url: https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc
+#     packages:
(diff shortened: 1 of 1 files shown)


vitaly-krugl:
Oh man, that's a pain :(. By the way, have you come across these?
* https://askubuntu.com/questions/329450/e-some-index-files-failed-to-download-they-have-been-ignored-or-old-ones-used 
* https://ubuntuforums.org/showthread.php?t=1654068

lukebakken:
@vitaly-krugl I did see something similar but I'm not sure we can do anything about it. I'll add those links to  travis-ci/travis-ci#6302

vitaly-krugl:
@lukebakken, would this help: https://www.rabbitmq.com/install-debian.html#bintray-apt-repo ?

.travis.yml
```diff
@@ -2,19 +2,24 @@ language: python
...
-    sources:
-      - sourceline: deb https://packages.erlang-solutions.com/ubuntu trusty contrib
-        key_url: https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc
-    packages:
-      - erlang-nox
+services:
+  - rabbitmq
+
+# TODO:
+# set 'sudo: false' and uncomment once "Hash Sum mismatch" failure in 'apt-get update' is fixed
```
vitaly-krugl: Where is `apt-get update` called?

.travis.yml
```diff
@@ -27,31 +32,31 @@ stages:
 - name: deploy
   if: tag IS present
 
-cache:
-  apt: true
```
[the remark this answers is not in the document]
vitaly-krugl: I tried apt: false and removing the apt line from cache, but it didn't help.

.travis.yml
lukebakken: We don't need it now as we're not installing anything with these changes and using the already-installed Erlang and RabbitMQ

PR #1017, opened 2018-04-14
Fix TypeError not prevent running the example

CharlesCCC:
Fix `TypeError: basic_consume() got an unexpected keyword argument 'callback'`


vitaly-krugl:
@CharlesCCC: thank you for your contribution, but the example code in head of pika master branch matches `Channel.basic_consume()` args in head of pika master branch. The pika APIs in master are getting normalized for the v1.0.0 release.

You need to use examples from the release label that is compatible with the version of pika that you're using.

PR #1026, opened 2018-04-27
Cleaned up "on channel closed" callback and ChannelClosed exception scenarios

vitaly-krugl:
## Backward-incompatible improvements for v1.0.0:
- Raise `ChannelWrongStateError` instead of `ChannelClosed` and `ChannelAlreadyClosing` when requesting channel operation incompatible with channel state. For consistency with `ConnectionWrongStateError` scenarios. Also foregoes having to invent fake reason codes.
- Got rid of `pika.channel.ClientChannelError` enumeration. The fake channel error codes aren't needed any more.
- `"_on_channel_close"` callback registered via `Channel.add_on_close_callback()` passes
exception as reason instead of reply_code/reply_text. For consistency with "on connection closed" callback.
- Removed `Channel.is_closed_by_broker` property (which was added not long ago for v1.0.0). Users can now simply check whether the reason exception passed to the close callback is an instance of `ChannelClosedByBroker` exception.

## Other improvements for v1.0.0:
- Added ChannelClosed-based exceptions ChannelClosedByClient and ChannelClosedByBroker to help user distinguish between the scenarios.

### DIFF
--- a/docs/examples/asynchronous_consumer_example.rst
+++ b/docs/examples/asynchronous_consumer_example.rst
@@ -142,7 +142,7 @@ consumer.py::
             LOGGER.info('Adding channel close callback')
             self._channel.add_on_close_callback(self.on_channel_closed)
 
-        def on_channel_closed(self, channel, reply_code, reply_text):
+        def on_channel_closed(self, channel, reason):
             """Invoked by pika when RabbitMQ unexpectedly closes the channel.
             Channels are usually closed if you attempt to do something that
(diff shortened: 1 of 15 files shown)


vitaly-krugl:
@lukebakken, it's ready for review once again: I resolved merge conflicts against latest master, made `BlockingChannel` context manager compatible with the channel exception changes, and added corresponding `BlockingChannel` acceptance tests.

vitaly-krugl:
Thanks @michaelklishin

PR #1029, opened 2018-05-02
Privatize non-blocking adapters' methods add_timeout(), remove_timeout(), and add_connection_threadsafe() for v1.0.0

vitaly-krugl:
## Backward-incompatible cleanup for v1.0.0:
- Privatize non-blocking adapters' add_timeout(), remove_timeout(), add_connection_threadsafe().
  - Rationale: pika.connection.Connection defines the abstract I/O methods
for services that it and generic connection tests need which may change without notice. But they are not for clients, who should instead be using the native methods for timers and thread-safe callbacks provided by their ioloops. In other words, a non-blocking connection instance is not an ioloop.

## Other changes:
- Added blurb in README describing how to safely ACK a mesage from another thread.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -95,18 +95,19 @@ the connection adapter's `create_connection()` class method.
 
 Requesting message ACKs from another thread
 -------------------------------------------
-The single-threaded usage constraint of inidivdual Pika connection adapter
-instances may result in a dropped AMQP connection due to heartbeat timeout in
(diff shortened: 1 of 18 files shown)


vitaly-krugl:
Resolved merge conflicts with master branch.

vitaly-krugl:
Thank you!

PR #1030, opened 2018-05-02
Update the client and server configuration examples for TLS communication

maggyero:
- ``ssl.PROTOCOL_TLSv1`` is now deprecated (https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLSv1), ``ssl.PROTOCOL_TLS`` should be preferred, which is used by ``ssl.create_default_context`` among other parameters which provide a higher security level (https://docs.python.org/3/library/ssl.html#ssl.create_default_context).
- Checking server hostnames prevents man-in-the-middle attacks (https://github.com/pika/pika/issues/464).
- RabbitMQ 3.7.0 introduced a new systemctl syntax for rabbitmq.conf (https://www.rabbitmq.com/configure.html#config-file).
- TLS server configuration for HTTPS for the ``rabbitmq-management`` plugin might as well be demonstrated.

### DIFF
--- a/docs/examples/tls_mutual_authentication.rst
+++ b/docs/examples/tls_mutual_authentication.rst
@@ -1,7 +1,7 @@
 TLS parameters example
 ======================
 
-This examples demonstrates a TLS session with RabbitMQ using mutual authentication. It was tested against RabbitMQ 3.7.4, using Python 3.6.5 and Pika 1.0.0b1.
+This example demonstrates a TLS session with RabbitMQ using mutual authentication (server and client authentication). It was tested against RabbitMQ 3.7.4, using Python 3.6.5 and Pika 1.0.0b1.
 
 See https://www.rabbitmq.com/ssl.html for certificate generation and RabbitMQ TLS configuration.
 
@@ -16,14 +16,13 @@ tls_example.py::
         cafile="/Users/me/tls-gen/basic/testca/cacert.pem")
     context.load_cert_chain("/Users/me/tls-gen/basic/client/cert.pem",
                             "/Users/me/tls-gen/basic/client/key.pem")
-    server_hostname = "example.com"
     ssl_options = pika.SSLOptions(context=context,
-                                  server_hostname=server_hostname)
+                                  server_hostname="example.com")
     conn_params = pika.ConnectionParameters(ssl_options=ssl_options)
     
     with pika.BlockingConnection(conn_params) as conn:
         ch = conn.channel()
-        print(ch.queue_declare("foobar"))
+        ch.queue_declare("foobar")
         ch.publish("", "foobar", "Hello, world!")
         print(ch.basic_get("foobar"))
 
--- a/docs/examples/tls_mutual_authentication.rst
+++ b/docs/examples/tls_mutual_authentication.rst
@@ -1,60 +1,45 @@
 TLS parameters example
-=============================
-This examples demonstrates a TLS session with RabbitMQ using mutual authentication.
+======================
 
-It was tested against RabbitMQ 3.6.10, using Python 3.6.1 and pre-release Pika `0.11.0`
+This examples demonstrates a TLS session with RabbitMQ using mutual authentication. It was tested against RabbitMQ 3.7.4, using Python 3.6.5 and Pika `1.0.0b1`.
 
-Note the use of `ssl_version=ssl.PROTOCOL_TLSv1`. The recent verions of RabbitMQ disable older versions of
-SSL due to security vulnerabilities.
-
-See https://www.rabbitmq.com/ssl.html for certificate creation and rabbitmq SSL configuration instructions.
-
+See https://www.rabbitmq.com/ssl.html for certificate generation and RabbitMQ SSL configuration instructions.
 
 tls_example.py::
 
-    import ssl
-    import pika
     import logging
+    import pika
+    import ssl
 
     logging.basicConfig(level=logging.INFO)
-
-    context = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
-    context.verify_mode = ssl.CERT_REQUIRED
-    context.load_verify_locations('/Users/me/tls-gen/basic/testca/cacert.pem')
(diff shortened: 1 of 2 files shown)


docs/examples/tls_mutual_authentication.rst
```diff
@@ -1,60 +1,45 @@
...
-    ch = conn.channel()
-    print(ch.queue_declare("sslq"))
-    ch.publish("", "sslq", "abc")
-    print(ch.basic_get("sslq"))
-
+    context = ssl.create_default_context(
+        cafile="/Users/me/tls-gen/basic/testca/cacert.pem")
+    context.load_cert_chain("/Users/me/tls-gen/basic/client/cert.pem",
+                            "/Users/me/tls-gen/basic/client/key.pem")
+    server_hostname = "example.com"
```
vitaly-krugl: A single-use variable containing a simple value is unnecessary. Use `'example.com'` directly in the `pika.SSLOptions()` instantiation below.

docs/examples/tls_mutual_authentication.rst
```diff
@@ -1,60 +1,45 @@
...
+    
+    with pika.BlockingConnection(conn_params) as conn:
+        ch = conn.channel()
+        print(ch.queue_declare("foobar"))
+        ch.publish("", "foobar", "Hello, world!")
+        print(ch.basic_get("foobar"))
 
 rabbitmq.config::
 
-    %% Both the client and rabbitmq server were running on the same machine, a MacBookPro laptop.
```
vitaly-krugl: @lukebakken or @michaelklishin : need you to review the rabbitmq sample config

docs/examples/tls_mutual_authentication.rst
vitaly-krugl: Please apply the suggested changes in this function as well: https://github.com/pika/pika/blob/9fb234c0e91b709b5935f4d97432c5991597c61f/tests/acceptance/async_test_base.py#L120

docs/examples/tls_mutual_authentication.rst
```diff
@@ -1,60 +1,44 @@
...
-          %%
-          {ssl_options, [{cacertfile,           "/Users/me/tls-gen/basic/testca/cacert.pem"},
-                         {certfile,             "/Users/me/tls-gen/basic/server/cert.pem"},
-                         {keyfile,              "/Users/me/tls-gen/basic/server/key.pem"},
-                         {verify,               verify_peer},
-                         {fail_if_no_peer_cert, true}]}
-        ]
-      }
-    ].
+    # Enable A.M.Q.P.S.
```
vitaly-krugl: Please make sure these are consistent with https://github.com/pika/pika/blob/master/testdata/rabbitmq.conf.in

vitaly-krugl: @maggyero, sorry, I should have been more specific - I meant the properties being set, not the values (e.g., file paths) themselves. Thank you.

vitaly-krugl:
See also issue #1035.

docs/examples/tls_mutual_authentication.rst
maggyero: To me the other properties in the suggested rabbitmq.conf.in (``listeners.tcp.default = 5672``, ``num_acceptors.tcp = 10``, ``num_acceptors.ssl = 10``, ``reverse_dns_lookups = false``, ``loopback_users.guest = true``, ``log.console = false``, ``log.console.level = debug``) are distracting for the reader who would expect in an example like this only the necessary parameters for setting up TLS.

michaelklishin: I agree with @maggyero.

docs/examples/tls_mutual_authentication.rst
```diff
@@ -1,60 +1,44 @@
...
+    listeners.ssl.default = 5671
+    ssl_options.cacertfile = PIKA_DIR/testdata/certs/ca_certificate.pem
+    ssl_options.certfile = PIKA_DIR/testdata/certs/server_certificate.pem
+    ssl_options.keyfile = PIKA_DIR/testdata/certs/server_key.pem
+    ssl_options.verify = verify_peer
+    ssl_options.fail_if_no_peer_cert = true
+
+    # Enable H.T.T.P.S.
```
michaelklishin: Can we change this to just "HTTPS" (this is not a music album cover ;)) and add a comment that says "See http://www.rabbitmq.com/ssl.html to learn more"? The rest looks reasonable to me.

maggyero: From [Wikipedia](https://en.wikipedia.org/wiki/Acronym#Punctuation):

> In English, abbreviations have traditionally been written with a full stop/period/point in place of the deleted part to show the ellipsis of letters – although the colon and apostrophe have also had this role – and with a space after full stops (e.g. "A. D."). In the case of most acronyms, each letter is an abbreviation of a separate word and, in theory, should get its own termination mark. Such punctuation is diminishing with the belief that the presence of all-capital letters is sufficient to indicate that the word is an abbreviation.
> Some influential style guides, such as that of the BBC, no longer require punctuation to show ellipsis; some even proscribe it. Larry Trask, American author of The Penguin Guide to Punctuation, states categorically that, in British English, "this tiresome and unnecessary practice is now obsolete".
> Nevertheless, some influential style guides, many of them American, still require periods in certain instances. For example, The New York Times' guide recommends following each segment with a period when the letters are pronounced individually, as in K.G.B., but not when pronounced as a word, as in NATO. The logic of this style is that the pronunciation is reflected graphically by the punctuation scheme.

Since periods are disappearing, I’ve just removed them (but note that they’re *not* wrong). Done.

maggyero: > add a comment that says "See http://www.rabbitmq.com/ssl.html to learn more"?

It’s actually already there at the beginning of the example so it would be a repetition:

> See https://www.rabbitmq.com/ssl.html for certificate generation and RabbitMQ TLS configuration.

michaelklishin:
Thank you!

PR #1033, opened 2018-05-03
Fix race condition between BlockingConnection.add_callback_threasafe() and BlockingConnection._cleanup()

vitaly-krugl:
Fix race condition between `BlockingConnection.add_callback_threasafe()` and `BlockingConnection._cleanup()`.

`BlockingConnection.add_callback_threasafe()` raises `ConnectionWrongStateError` when called during/after blocking connection is closed.


michaelklishin:
Thank you!

PR #1039, opened 2018-05-08
Fix setup.py included packages

guillaume-michel:
Current setup.py does not install pika module correctly due to a bug in find_pakages syntax. This merge request fixes that.


michaelklishin:
I'm curious why the implementation we had did install successfully from git via pip?

guillaume-michel:
I don't know, maybe you use editable mode?
Thanks for merging the branch so quickly.

lukebakken:
@guillaume-michel what version of Python and `setuptools` are you using? This has never been a problem before. We need to make sure this doesn't affect older Python version.

PR #1045, opened 2018-05-14
WIP: fix issue #1044

### LINKED ISSUE #1044: Performance of add_callback_threadsafe() relative to add_timeout()?
I am seeing some unexpected performance issues with `BlockingConnection.add_callback_threadsafe()` relative to` BlockingConnection.add_timeout()` in Pika 0.12.0b2, and I am wondering if this is a known issue or not.

Details follow...

I have some (working) Pika code that does the following using the BlockingConnection API:
```
connection.add_timeout(0.5, process_outbox)
channel.start_consuming()
```
Then, `process_outbox()` looks something like this:
```
def process_outbox():
    # If any outgoing messages are waiting in the "outbox", send them via channel.basic_publish()
    ....
    connection.add_timeout(0.5, process_outbox)
```
As a result, Pika calls `process_outbox()` every half second, and that's when pending messages are sent.  So far, so good.

But that approach involves up to a half-second of latency, and it spends a lot of time checking the outbox when there are no messages to send.  So rewrote my code to use `BlockingConnection.add_callback_threadsafe()` from Pika 0.12.0b2, expecting it would be more efficient.  In this new version, I just enter `channel.start_consuming()` right away (i.e., without setting a timeout).  Then, when I need to send a message, I do something like the following:
```
def send(msg):
    # Put the msg in the outbox (a synchronized queue.Queue)
    ....
    connection.add_callback_threadsafe(process_outbox)
```
  where, now, `process_outbox()` is as before, but it doesn't call `add_timeout()`:
```
def process_outbox():
    # If any outgoing messages are waiting in the "outbox", send them via channel.basic_publish()
    ....
```
I have a couple test programs that I used to compare the two approaches (both using Pika 0.12.0b2).  One program queues up 3 messages to be sent ASAP, and another program recieves them.
* In the `add_timeout()` case, I typically see a delay of about half a second until the `process_outbox()` callback is called, and then the messages are all sent and received at once -- as one might expect.
* In the `add_callback_threadsafe()` case, though, I see a delay of multiple SECONDS before the `process_outbox()` callback is called and the messages go out.

From what I can see in the code, `add_callback_theadsafe()` calls `add_timeout()` with a time delay of 0 seconds (via` _on_threadsafe_callback()`), so this is a little surprising... although the `add_callback_threadsafe()` call comes from another thread, so I suppose that there may be thread locking issues.  (I can't find where that is or confirm it via a profiler, though.)

My question is: Is this the expected behavior of  `add_callback_theadsafe()`?  And, are their any suggestions to make its usage more efficient?

I've attached a sample program that, on my system, consistently shows a delay of about 4 seconds before the callback specified by `add_callback_threadsafe()` is called.
[threadsafe_callback.py.txt](https://github.com/pika/pika/files/1992745/threadsafe_callback.py.txt)

lukebakken:
Fixes #1044 

This PR removes the use of `_thread_id` in `select_connection.py` and will always wake up the IOloop for cross-thread operations. Implemented as per @vitaly-krugl's [comment](https://github.com/pika/pika/issues/1044#issuecomment-388541195)

TODO: docstrings

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -940,9 +940,7 @@ class SelectPoller(_PollerBase):
                     # NOTE When called without any FDs, select fails on
                     # Windows with error 10022, 'An invalid argument was
                     # supplied'.
-                    time.sleep(self._get_max_wait())
                     read, write, error = [], [], []
-
                 break
             except _SELECT_ERRORS as error:
                 if _is_resumable(error):
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -335,9 +335,6 @@ class IOLoop(AbstractSelectorIOLoop):
         # Callbacks requested via `add_callback`
         self._callbacks = collections.deque()
 
-        # Identity of this IOLoop's thread
-        self._thread_id = None
-
         self._poller = self._get_poller(self._get_remaining_interval,
                                         self.process_timeouts)
 
@@ -439,9 +436,9 @@ class IOLoop(AbstractSelectorIOLoop):
 
         # NOTE: `deque.append` is atomic
         self._callbacks.append(callback)
-        if threading.current_thread().ident != self._thread_id:
-            # Wake up the IOLoop running in another thread
-            self._poller.wake_threadsafe()
+
+        # Wake up the IOLoop running in another thread
+        self._poller.wake_threadsafe()
 
         LOGGER.debug('add_callback_threadsafe: added callback=%r', callback)
 
@@ -508,7 +505,6 @@ class IOLoop(AbstractSelectorIOLoop):
         exit. See `IOLoop.stop`.
 
         """
-        self._thread_id = threading.current_thread().ident
         self._poller.start()
 
     def stop(self):
@@ -521,18 +517,12 @@ class IOLoop(AbstractSelectorIOLoop):
             `ioloop.add_callback_threadsafe(ioloop.stop)`
 
         """
-        if (self._thread_id is not None and
-                threading.current_thread().ident != self._thread_id):
-            LOGGER.warning('Use add_callback_threadsafe to request '
-                           'ioloop.stop() from another thread')
-
         self._poller.stop()
 
     def activate_poller(self):
         """[Extension] Activate the poller
 
         """
-        self._thread_id = threading.current_thread().ident
         self._poller.activate_poller()
 
     def deactivate_poller(self):
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -937,10 +937,8 @@ class SelectPoller(_PollerBase):
                         self._fd_events[PollEvents.ERROR],
                         self._get_max_wait())
                 else:
-                    # NOTE When called without any FDs, select fails on
-                    # Windows with error 10022, 'An invalid argument was
-                    # supplied'.
-                    read, write, error = [], [], []
+                    # https://github.com/pika/pika/issues/1044#issuecomment-388514004
+                    raise RuntimeError('All _fd_events arrays are empty')
                 break
             except _SELECT_ERRORS as error:
                 if _is_resumable(error):
@@ -949,7 +947,6 @@ class SelectPoller(_PollerBase):
                     raise
 
         # Build an event bit mask for each fileno we've received an event for
-
         fd_event_map = collections.defaultdict(int)
         for fd_set, evt in zip(
                 (read, write, error),


pika/adapters/select_connection.py
```diff
@@ -439,9 +436,9 @@ def add_callback_threadsafe(self, callback):
 
         # NOTE: `deque.append` is atomic
         self._callbacks.append(callback)
-        if threading.current_thread().ident != self._thread_id:
-            # Wake up the IOLoop running in another thread
-            self._poller.wake_threadsafe()
+
+        # Wake up the IOLoop running in another thread
```
vitaly-krugl: => `# Wake up the IOLoop *potentially* running in another thread`

pika/adapters/select_connection.py
```diff
@@ -937,12 +927,8 @@ def poll(self):
                         self._fd_events[PollEvents.ERROR],
                         self._get_max_wait())
                 else:
-                    # NOTE When called without any FDs, select fails on
-                    # Windows with error 10022, 'An invalid argument was
-                    # supplied'.
-                    time.sleep(self._get_max_wait())
-                    read, write, error = [], [], []
-
```
vitaly-krugl: I would prefer to leave this time.sleep() code in place. The reason being that the poller should be able to work without file descriptors. The fact that we presently always have at least one file descriptor is an artifact of the current design. In particular, pushing the running of event loop and implementation of wakeup all the way down to the poller layer is arguably misplaced. As the result, there are several up/down delegations of API calls that don't belong. I am planning to refactor such that running of the event loop and wakeup logic is contained in the IOLoop class, while pollers are only responsible for the polling abstraction, which I believe to be more orthogonal. In that case, poller's implementation won't be guaranteed that it will always be used with >= 1 fd.

pika/adapters/select_connection.py
```diff
@@ -439,9 +436,9 @@ def add_callback_threadsafe(self, callback):
```
vitaly-krugl: Also need to add the efficiency blurb in `IOLoop.add_callback_threadsafe()` docstring (just like the one you added in `BlockingConnection.add_callback_threadsafe()`). Except that in IOLoop, it's `call_later()` instead of `add_timeout()` and `delay` instead of `deadline`.

pika/adapters/select_connection.py
```diff
@@ -937,12 +927,8 @@ def poll(self):
...
                         self._get_max_wait())
                 else:
-                    # NOTE When called without any FDs, select fails on
-                    # Windows with error 10022, 'An invalid argument was
-                    # supplied'.
-                    time.sleep(self._get_max_wait())
-                    read, write, error = [], [], []
-
+                    # https://github.com/pika/pika/issues/1044#issuecomment-388514004
+                    raise RuntimeError('All _fd_events arrays are empty')
```
vitaly-krugl: By the way, my understanding is that `RuntimeError` is intended to be used as the base class of the Python runtime errors, not our errors encountered during runtime.

lukebakken: I referred to [this documentation](https://docs.python.org/2/library/exceptions.html#exceptions.RuntimeError). It seemed like the best choice.

pika/adapters/select_connection.py
lukebakken: I don't understand why the `time.sleep` call is there, though. Won't it block everything for 5 seconds, guaranteed?

pika/adapters/select_connection.py
lukebakken: Before `1.0.0` is released should these three methods all have the same name and signature?

pika/adapters/select_connection.py
lukebakken: It seems like 5 seconds is too long for a sleep, though, since it will block all fd activity, right?

vitaly-krugl:
@lukebakken:

regarding `time.sleep()` in select_connection's poller code...
> It seems like 5 seconds is too long for a sleep, though, since it will block all fd activity, right?

`select_connection.IOLoop`'s asynchronous behavior is governed only by 1. timers; 2. thread-safe callbacks; and 3. file descriptors. No other behaviors are possible with this IOLoop. If no file descriptors are registered, it leaves only timers and thread-safe callbacks, in which case the next time that the I/O loop would need to wake up is governed by `self._get_max_wait()`. 

`SelectPoller` - which uses `select.select()` for polling - cannot call `select.select()` without file descriptors on Windows (windows raises an error in this case, although linux/unix do not). Functionally, calling `time.sleep(self._get_max_wait())` is equivalent to calling `select.slect([], [], [], self._get_max_wait())`.

The other pollers that are based on `poll()`, `epoll()`, and OS X kernel queues don't have the Windows problem and act equivalently to `time.sleep()` when no file descriptors are registered.

PR #1057, opened 2018-05-31
Heartbeats should be sent at half the specified interval

### LINKED ISSUE #1055: HeartbeatChecker is confused about heartbeat timeouts

I see two problems here:

lukebakken:
Fixes #1055

### DIFF
--- a/pika/heartbeat.py
+++ b/pika/heartbeat.py
@@ -12,21 +12,24 @@ class HeartbeatChecker(object):
     intervals.
 
     """
+    DEFAULT_INTERVAL = 60
     MAX_IDLE_COUNT = 2
+
     _STALE_CONNECTION = "Too Many Missed Heartbeats, No reply in %i seconds"
 
-    def __init__(self, connection, interval, idle_count=MAX_IDLE_COUNT):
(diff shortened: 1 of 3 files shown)


pika/heartbeat.py
```diff
@@ -12,21 +12,24 @@ class HeartbeatChecker(object):
     intervals.
 
     """
+    DEFAULT_INTERVAL = 600
     MAX_IDLE_COUNT = 2
+
     _STALE_CONNECTION = "Too Many Missed Heartbeats, No reply in %i seconds"
 
-    def __init__(self, connection, interval, idle_count=MAX_IDLE_COUNT):
+    def __init__(self, connection, interval=DEFAULT_INTERVAL, idle_count=MAX_IDLE_COUNT):
```
lukebakken: @vitaly-krugl I didn't want to do too many changes to this class but we could remove these two parameters and use default values within the class if you prefer.

pika/heartbeat.py
```diff
@@ -12,21 +12,24 @@ class HeartbeatChecker(object):
     intervals.
 
     """
+    DEFAULT_INTERVAL = 600
```
michaelklishin: It's been 60 seconds by default in the broker and most clients for a good couple of years now.

PR #1058, opened 2018-06-03
 Add and correct the implementations of comparison methods

maggyero:
This pull request:

a) adds `__eq__` and `__ne__` methods to the `pika.Parameters` class;
b) corrects the existing implementations of the `__eq__`, `__lt__`, `__le__` and `__ne__` methods of the `pika.PlainCredentials`, `pika.ExternalCredentials` and `pika._Timeout` classes;
c) adds and corrects the corresponding unit tests.

In the a) case, I had a need at my company to compare two instances of the `pika.ConnectionParameters` class when unit testing a function that takes a dictionary of connection parameters as input and returns an instance of the `pika.ConnectionParameters` as output.

In the b) case, the implementations of `__eq__`, `__lt__` and `__le__` were incorrect as they were *not commutative* when the other operand is of a different class. The implementation of `__ne__` was also incorrect as *it failed to replicate* the Python 3 default implementation of `__ne__` which fallbacks on the `__ne__` of the other operand when `__eq__` returns `NotImplemented`.

See this Stackoverflow [community post](https://stackoverflow.com/a/25176504/2326961) for more details on the canonical implementations of `__eq__` and `__ne__`, or the Python [documentation](https://docs.python.org/3/library/functools.html#functools.total_ordering) for an authoritative example (with a weaker constraint on the operands).


pika/connection.py
```diff
@@ -157,6 +157,32 @@ def __repr__(self):
                 (self.__class__.__name__, self.host, self.port,
                  self.virtual_host, bool(self.ssl_options)))
 
+    def __eq__(self, other):
+        if isinstance(self, other.__class__):
+            return (
+                self._backpressure_detection == other._backpressure_detection
```
michaelklishin: RabbitMQ identifies connections using a 4-tuple: client hostname, client port, server hostname, server port. Should the same logic be applied here? Of course it would assume that the connection has an open socket.

maggyero: Do you mean we should only check the equality of these 4 parameters instead of all the parameters?

lukebakken: It makes sense to use the same comparison logic for connections that RabbitMQ does.

maggyero: Alright. I’m going to correct this and make some tests.

maggyero: @michaelklishin @lukebakken Sorry I misread. I am not sure I understand now. Which connection parameters do you suggest we compare exactly? You are talking about *client* and *server* hostnames and ports, but as far as I know we only have access to the server hostnames and ports (`self._host` and `self._port`).

lukebakken: Oh, right. Hmm it would be nice to have access to the ephemeral ports in use as that would identify the connection. That seems like the correct way to compare two connections.

maggyero:
@lukebakken:

> Thank you. Please provide tests that fail prior to these changes and succeed after applying them.

Done.

maggyero:
@lukebakken:

> Thank you. Please provide tests that fail prior to these changes and succeed after applying them.

Done. For the moment I restricted the comparison of the `Parameters` instances to the server `_host`and `_port` parameters. If the client parameters are needed, they can be easily added.

lukebakken:
@vitaly-krugl if you have a moment I'd like to see what you think. Thanks!

maggyero:
@lukebakken With the last update from the master branch (previously a non PR-related test on Python 2 failed), all tests have passed. So I think you can re-accept this PR now.

ottohirr:
I happened to be perusing the credential code and found this.

```
class PlainCredentials(object):
    ...
    def __eq__(self, other):
        if isinstance(other, PlainCredentials):
            return (self.username == other.username
                    and self.password == other.password
                    and self.erase_on_connect == other.erase_on_connect)
        return NotImplemented
```

Semantically I believe this gives the wrong _impression_.

If the `erase_on_connect` is true, then there is no way to determine if the credentials are the same if they have been used in a connection because the `username` and `password` are set to `None`.

Example:

```
cred_a = pika.PlainCredentials( 'abc', 'def', erase_on_connect=True )
cred_b = pika.PlainCredentials( 'ghi', 'jkl', erase_on_connect=True )

# cred_a == cred_b would be false

# use cred_a and cred_b in connection

# cred_a == cred_b would be true
```

I believe that is not what equality should identify.

Maybe the more appropriate approach is if `username`/`password` is `None` and `erase_on_connect` is `True`, then it should return `NotImplemented` or some other indicator of being unknowable.

Regards,
..Otto

PR #1061, opened 2018-06-06
WIP: Validate parameters to API methods that have changed

### LINKED ISSUE #977: 1.0.0 API changes could raise helpful exceptions
If an API change resulted in a parameter order change where the user could pass a callable where it is not intended, Pika could raise `TypeError` with a helpful message.

See the discussion at the end of issues #963 and #945

lukebakken:
Fixes #977

Raise helpful exceptions if legacy parameter ordering is used after upgrading to Pika 1.0.0

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,19 @@ Version History
 
 - `BlockingConnection.consume` now returns `(None, None, None)` when inactivity timeout is reached (`PR <https://github.com/pika/pika/pull/899>`_)
 
+API changes from previous versions:
+
+pika/channel.py
+
+def basic_consume(self,
+                  queue,
+                  on_message_callback,
+                  auto_ack=False,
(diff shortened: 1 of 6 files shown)


CHANGELOG.rst
```diff
@@ -8,6 +8,19 @@ Version History
...
+pika/channel.py
+
+def basic_consume(self,
+                  queue,
+                  on_message_callback,
+                  auto_ack=False,
+                  exclusive=False,
+                  consumer_tag=None,
+                  arguments=None,
+                  callback=None)
```
lukebakken: @vitaly-krugl how would you like API changes documented? I can't really think of a good way, except maybe having old version next to the new one in a table, perhaps.

vitaly-krugl: > except maybe having old version next to the new one in a table, perhaps

@lukebakken - this sounds reasonable

PR #1064, opened 2018-06-06
Add Pika version to connection logging

### LINKED ISSUE #1060: Include the Pika version number in the log when opening a connection
See @vitaly-krugl's comment [here](https://github.com/pika/pika/issues/1046#issuecomment-392585166)

lukebakken:
Fixes #1060

### DIFF
--- a/pika/adapters/utils/connection_workflow.py
+++ b/pika/adapters/utils/connection_workflow.py
@@ -14,6 +14,7 @@ import socket
 import pika.compat
 import pika.exceptions
 import pika.tcp_socket_opts
+from pika import __version__
 
 
 _LOG = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


pika/adapters/utils/connection_workflow.py
```diff
@@ -186,7 +187,8 @@ def start(self, addr_record, conn_params, on_done):
         self._sock.setblocking(False)
 
         addr = self._addr_record[4]
-        _LOG.info('Connecting to AMQP broker at %r', addr)
+        _LOG.info('Pika version %s connecting to AMQP broker at %r',
```
michaelklishin: Can we make this say "node"? "AMQP" is used to refer to two protocols and the target could be a proxy or load balancer, too.

PR #1066, opened 2018-06-06
Fixup for #1055

### LINKED ISSUE #1055: HeartbeatChecker is confused about heartbeat timeouts

lukebakken:
Note: even though we send heartbeats in half the specified interval, the broker will be sending them to us at the specified interval. This means we will be checking for an idle connection twice as many times as the broker will send heartbeats to us, so we need to double the max idle count.

The changes in `example/consume.py` back-port the example changes from the `master` branch.

Fixes #1055

### DIFF
--- a/examples/consume.py
+++ b/examples/consume.py
@@ -1,17 +1,15 @@
+import functools
+import logging
 import pika
 
-def on_message(channel, method_frame, header_frame, body):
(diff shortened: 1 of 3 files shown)


lukebakken:
Thank you @michaelklishin

Issue #1031, opened 2018-05-02
Remove adapters/twisted_connection.py due to lack of test coverage. (pika v1.0.0 release target)


vitaly-krugl:
The `twisted_connection` module was added in 2011, but to this date still has no meaningful test coverage. It's unmaintainable without comprehensive test coverage and therefore pika cannot release v1.0.0 with `twisted_connection` in it.

We literally don't know whether it works at all at this point and which API methods are working/broken.

I am hoping that this issue will inspire someone with interest in the twisted adapters to add comprehensive test coverage shortly. Barring that, `twisted_connection` must be removed from the package.

See the following for examples of pika acceptance test suites:
* https://github.com/pika/pika/blob/master/tests/acceptance/async_adapter_tests.py
* https://github.com/pika/pika/blob/master/tests/acceptance/blocking_adapter_test.py

See also issues #615 and #616.

vitaly-krugl:
cc @lukebakken

wulczer:
Hey, as the original author of the adapter I'm OK with removing it if no one (including me) finds the time to add test coverage.

I'll do my best to get something written, but I can't promise anything and if you decide it's time to cut a 1.0.0 and I haven't gotten my act together, go ahead and get rid of it.

cellscape:
I no longer use pika with twisted, removing is fine by me.

Zephor5:
Thx for the information, it's ok for me. Maybe I'll get something done in the future.

glyph:
Does #1069 cover the code that resolves this?

vitaly-krugl:
Yes, it does.

glyph:
Thanks so much to @abompard for keeping the Twisted dream alive! 🎉

wulczer:
Thank you indeed @abompard - and apologies for not keeping up with the maintenance of the adapter. Three cheers for the Twisted community!

PR #1069, opened 2018-06-07
Add tests for the Twisted adapter

### LINKED ISSUE #615: Implement acceptance tests for TwistedProtocolConnection adapter
There are presently no acceptance tests for TwistedProtocolConnection adapter, and this adapter has been broken for a while as a result (see PR #613 and #614)

abompard:
This commit does not add test to the `TwistedConnection`, only to the `TwistedProtocolConnection`. I think that the `TwistedConnection` can be deprecated and/or dropped, because is uses the blocking `socket` module and that's discouraged in Twisted.
When those pieces are removed, the tests coverage is almost 100% (only the `NotImplementedError`s are uncovered).

The commit also fixes a couple bugs in the Twisted adapter that likely come from evolving code in the rest of the project.

Fixes #615, #1031


lukebakken:
@abompard thanks a lot! I have taken the liberty of removing `TwistedConnection` and pushed those updates to your branch. I'm running tests and reviewing all this now.

pika/adapters/twisted_connection.py
```diff
@@ -397,7 +310,7 @@ def _adapter_get_write_buffer_size(self):
         """Implement pure virtual
         :py:ref:meth:`pika.connection.Connection._adapter_emit_data()` method.
 
-        TODO: this method only belongs in SelectConnection, none others needs it
+        TODO: this method only belongs in SelectConnection, no others need it
```
lukebakken: Hmm, I don't see an implementation of `_adapter_get_write_buffer_size` in `SelectConnection` ... @vitaly-krugl is that something that should be there?

vitaly-krugl:
@abompard, thank you for the contribution. I haven't check, but just in case, please make sure that all the affected modules pass pylint (with pika's pylintrc) cleanly. Cheers!

abompard:
I have the following warnings, that I've disabled:
* W0703, line 124 & 182, "Catching too general exception Exception (broad-except)": we don't know what the exception can be in the wrapped function so I think it's OK to be broad.
* W0221, line 241, "Parameters differ from overridden 'channel' method (arguments-differ)": that's intended because the second argument is forced on the overridden method.

But that's only my opinion, I can adapt if you disagree.

abompard:
Oh by the way, now that you've removed `TwistedConnection`, you can also remove `_TwistedIOServicesAdapter` and `_TwistedDeferredIOReference` as they were only used by `TwistedConnection`.

vitaly-krugl:
@abompard  - Re.
> I have the following warnings, that I've disabled:
>
> W0703, line 124 & 182, "Catching too general exception Exception (broad-except)": we don't know what the exception can be in the wrapped function so I think it's OK to be broad.
W0221, line 241, "Parameters differ from overridden 'channel' method (arguments-differ)": that's intended because the second argument is forced on the overridden method.
But that's only my opinion, I can adapt if you disagree.


W0703: I don't know what else can be done differently for it, so just have to live with it.

W0221: I understand that the native Twisted adaptation needs to have a different API because of Futures, for example. However, I think that the API overriding approach take by the implementation of the TwistedProtocolConnection (TPC after this) is suboptimal.
* Here, we can use the litmus test for subclassing versus containing: "is a" versus "is implemented in terms of". In my mind, TPC is not a Pika Connection Adapter in the sense that it's a Twisted Protocol Adapter and its ABI differs from Pika's Connection and BaseConnection. Just like Pika's BlockingConnection class that's implemented in terms of SelectConnection but doesn't subclass from it (see BlockingConnection._impl and BlockingChannel._impl), TPC should not subclass from Connection, but instead should contain an instance of an implementation class that derives from `pika.Connection`.
* And TPC should have its own explicitly-defined API without the `*args`/`*kwargs` funny business that falls short in terms of self-documenting code and when used with tab-completion-enabled IDE's. Looking at `*args`/`*kwargs`, I would have a bear of a time navigating/using this API as a developer.

Example of what I have in mind; it should not be much effort to refactor current implementation, and partinioning of responsibilities would make the implementation more maintainable too:

```
class _TwistedConnectionAdapter(pika.Connection):
    # Here, implement `pika.Connection`'s pure virtual methods - all that presently lives in 
    # `TwistedProtocolConnection`.
    # Also implement whatever accessors are going to be needed by the new 
    # `TwistedProtocolConnection` implementation
    
    def __init__(...<whatever is needed by your implementation>...):
        ....

class TwistedProtocolConnection(object):

    def __init__(...<whatever makes sense for twisted protocol adaptation>...):
        """"
        ....
        :param <type> ...: ....
        :param <type> ...: ....
        """"        self._impl = _TwistedConnectionAdapter(...)

   def channel(...<explicit args that make sense for twisted adaptation>...):
        """"
        :param <type> ...: ....
        :param <type> ...: ....
         
        :returns: <whatever makes sense for Twisted adaptation (Future probably in this case)>
        :rtype: ....
        """"
        .....

    def ....


class TwistedProtocolChannel(object):
    def __init__(self, channel_impl, <plus whatever is needed by TwistedProtocolConnection>):
        self._impl = channel_impl

    def basic_get(<whatever makes sense for Twisted adaptation>):
            """"
        :param <type> ...: ....
        :param <type> ...: ....
         
        :returns: <whatever makes sense for Twisted adaptation>
        :rtype: ....
        """"
        .....

     def ....

```

abompard:
Thanks for your suggestions @vitaly-krugl . I tried to do this refactoring in the last commit, please tell me if you like it better that way.

lukebakken:
@abompard yep, we're aware of that, but thanks for noticing.

pika/adapters/twisted_connection.py
```diff
@@ -397,416 +337,125 @@ def _adapter_get_write_buffer_size(self):
...
-                family,
-                socktype,
-                proto,
-                flags),
-            on_done)
-
-    def set_reader(self, fd, on_readable):
-        """Implement
-        :py:meth:`.utils.nbio_interface.AbstractFileDescriptorServices.set_reader()`.
+    def channel(self, channel_number=None):  # pylint: disable-msg=W0221
```
vitaly-krugl: @lukebakken, this has been bugging me for some time in Pika core - it provides a seemingly unnecessary arg channel_number that lets the user specify his own channel_number instead of letting pika do it automagically. Is there a real use case for this? Do any of the other clients in RabbitMQ echo-system let you do that?

pika/adapters/twisted_connection.py
```diff
@@ -105,35 +114,79 @@ def channel_closed(self, _channel, reason):
         self.__calls = set()
         self.__consumers = {}
```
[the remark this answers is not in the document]
abompard: I'm not the original author of that but if I understand correctly the point was to avoid conflicting with one of the wrapped method (since `__getattr__` is subclassed here). However none of the wrapped method start with an underscore and I don't think we will want to wrap one that does in the future, so a single underscore prefix should be sufficient. I also don't expect the `TwistedChannel` class to be subclassed. I'll change it.

pika/adapters/twisted_connection.py
```diff
@@ -105,35 +114,79 @@ def channel_closed(self, _channel, reason):
...
+                    (queue_obj, frame.method.consumer_tag)
+                ),
+            )
+        except Exception:  # pylint: disable-msg=W0703
             return defer.fail()
 
-        return defer.succeed((queue, consumer_tag))
+        return d
 
     def queue_delete(self, *args, **kwargs):
```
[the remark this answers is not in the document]
abompard: Sure, I'll do it, but we must be careful to update `TwistedChannel`'s wrapper methods whenever we update the main `Channel`'s methods. I guess that's a trade-off in favor of more advanced editors :-)

pika/adapters/twisted_connection.py
```diff
@@ -397,416 +337,125 @@ def _adapter_get_write_buffer_size(self):
...
+    that the host, port and ssl values of the connection parameters are ignored
+    because, yet again, it's Twisted who manages the connection.
 
-        """
-        check_callback_arg(callback, 'callback')
-        return _TimerHandle(self._reactor.callLater(delay, callback))
+    """
+    def __init__(self,
+                 parameters=None,
+                 on_close_callback=None):
```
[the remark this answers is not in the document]
abompard: You're right. I'll implement a mechanism similar to the `ready` attribute.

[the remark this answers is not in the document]
abompard: Sure, will do.

pika/adapters/twisted_connection.py
```diff
@@ -312,38 +271,19 @@ class TwistedProtocolConnection(pika.connection.Connection):
     """
     def __init__(self,
                  parameters=None,
+                 on_open_callback=None,
+                 on_open_error_callback=None,
                  on_close_callback=None,
```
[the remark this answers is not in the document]
abompard: Sure!

abompard:
Thanks a lot for your in-depth review @vitaly-krugl ! I've implemented the proposed changes in the last two commits. It seems to be working, Pylint is silent and coverage is as before (99% because of the two unimplemented methods).

lukebakken:
@vitaly-krugl I can do some long-running tests with code that uses this PR, but for a deep dive someone more familiar with twisted should check this out, or perhaps @gmr if he has time. We can also alert the twisted community when `1.0.0b1` is released with this code to test it out.

pika/adapters/twisted_connection.py
```diff
@@ -84,89 +94,172 @@ class TwistedChannel(object):
...
-        self.__calls = set()
-        self.__consumers = {}
+        self._channel = channel
+        self._closed = None
+        self._calls = set()
+        self._consumers = {}
 
         channel.add_on_close_callback(self.channel_closed)
 
     def channel_closed(self, _channel, reason):
```
vitaly-krugl: `channel_closed()` appears to be for internal use, so should be named with a leading underscore as a private method

pika/adapters/twisted_connection.py
```diff
@@ -183,126 +276,30 @@ def single_argument(*args):
...
+        self._consumers.pop(queue_name, None)
         return ret
 
-    def __clear_call(self, ret, d):
-        self.__calls.discard(d)
+    def _clear_call(self, ret, d):
+        self._calls.discard(d)
         return ret
 
     def __getattr__(self, name):
```
vitaly-krugl: @abompard, are you planning to add the remaining explicit Channel methods?

pika/adapters/twisted_connection.py
abompard: Right, I'll fix this.

pika/adapters/twisted_connection.py
abompard: I didn't think so, I thought you meant explicit signatures for the redefined methods only. But if you prefer not using `__getattr__`, I can copy all the methods in `WRAPPED_METHODS` in this class.
I'm just a bit worried it will be more maintenance in the long run, since they will have to be adapted in this class too whenever they are changed in the main Channel class. But I'll follow you if you think it's worthwhile.

vitaly-krugl:
I think it would be easier for developers to use if the methods we're all
there with explicit args. We're close to v1.0, so I don't anticipate method
signatures changing for a long time.

@lukebakken, does this work for you?

Thank you,
Vitaly

abompard:
I'm currently copying the other methods and I realize I should amend my last comment: all the public methods in `pika.Channel` must be copied, not only the redefined or wrapped ones. `TwistedChannel` only wraps the methods that have a callback, but without `__getattr__` even methods such as `basic_ack` must be redefined, only to return `self._channel.<parent method>`.

As I said I don't mind doing it, copying is easy, I'm just worried about the maintenance. Changes in the docstrings will have to be propagated too.

lukebakken:
@abompard this hopefully will be the only time this needs to be done. One goal of `1.0.0` is to get as close to a final API as possible. Thank you for all your hard work, this is great!

abompard:
Hey @lukebakken , did you get a chance to work on this? I don't want to be pushy, I understand other things can get in the way, it's just a friendly reminder :-)

lukebakken:
I was out on vacation last week, I'll see if I can find time for Pika this week.

lukebakken:
Tests pass, these changes appear to be good.

@jeremycline - I see you responded to this PR. Would you mind taking it for a spin?

lukebakken:
@abompard I merged this PR but noticed that tests fail on Python 2.7.15. Would you please address this? I couldn't figure out why this is happening, nor a fix. Transcript is attached here:

[gh-1069-test-transcript.txt](https://github.com/pika/pika/files/2246866/gh-1069-test-transcript.txt)

abompard:
I'll check it out, thanks for the merge!

abompard:
OK I found why, it's because `functools.wraps` in Python2 requires the `__name__` attribute on functions and `mock.Mock` does not have that. In Python3, `functools.wraps` is protected against missing attributes with a try..except block.

I have a patch, I'll make a PR (with a couple minor fixes to the Twisted adapter) as soon as I can pull from master. At the moment Github does not show me this PR as merged in master (a hiccup maybe?)

PR #1070, opened 2018-06-08
Update README.rst

ponugoti:
Remove erraneous character as part test message


lukebakken:
Does the example code not work with `b'Test Message'`?

lukebakken:
Using `b'Test Message'`, `u'Test Message'` or just `'Test Message'` sends the same 12 bytes to RabbitMQ. I'm going to close this for now but please reply with the reason you think this should be changed.

PR #1071, opened 2018-06-09
Changes from PR #1057 review

### LINKED ISSUE #1055: HeartbeatChecker is confused about heartbeat timeouts

lukebakken:
Fixes #1055, related to #1057

### DIFF
--- a/pika/heartbeat.py
+++ b/pika/heartbeat.py
@@ -8,37 +8,32 @@ LOGGER = logging.getLogger(__name__)
 
 class HeartbeatChecker(object):
     """Checks to make sure that our heartbeat is received at the expected
-    intervals.
+    timeouts.
 
     """
-    DEFAULT_INTERVAL = 60
-    MAX_IDLE_COUNT = 2
+    # Note: even though we're sending heartbeats in half the specified
+    # timeout value, the broker will be sending them to us at the specified
+    # value. This means we'll be checking for an idle connection via send_and_check
+    # twice as many times as the broker will send heartbeats to us,
+    # so we need to set max idle count to 4 here
+    _MAX_IDLE_COUNT = 4
     _CONNECTION_FORCED = 320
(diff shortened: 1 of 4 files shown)


vitaly-krugl:
@lukebakken, let's work out one of these parallel pull requests first, then create the second one after that.

pika/heartbeat.py
```diff
@@ -7,38 +7,38 @@
...
-    DEFAULT_INTERVAL = 60
-    MAX_IDLE_COUNT = 2
     _CONNECTION_FORCED = 320
+    _MAX_IDLE_COUNT = 2
     _STALE_CONNECTION = "Too Many Missed Heartbeats, No reply in %i seconds"
 
-    def __init__(self, connection, interval=DEFAULT_INTERVAL, idle_count=MAX_IDLE_COUNT):
-        """Create a heartbeat on connection sending a heartbeat frame every
-        interval seconds.
+    def __init__(self, connection, interval):
```
michaelklishin: Since the doc string says "timeout" and not "interval" and interval is a half of the timeout value per spec (2 missed heartbeats are considered to be a timeout), should we rename the argument and properties?

michaelklishin:
This [rabbitmq-users thread](https://groups.google.com/d/msg/rabbitmq-users/Fmfeqe5ocTY/0fxMMVsSAgAJ) is worth mentioning.

pika/heartbeat.py
michaelklishin: @vitaly-krugl

lukebakken: I will update the documentation. We are sending heartbeats at a specified interval (half of what's passed in) so I think that's an appropriate name for it.

lukebakken: @michaelklishin updated and pushed, let me know what you think. I already have a link to that mailing list thread in a comment as well.

PR #1072, opened 2018-06-09
Changes from PR #1057 review

### LINKED ISSUE #1055: HeartbeatChecker is confused about heartbeat timeouts

lukebakken:
Fixes #1055, related to #1057

(cherry picked from commit 69a239d3495ff268f0354f4fb5a11c7b746a7066)

### DIFF
--- a/pika/heartbeat.py
+++ b/pika/heartbeat.py
@@ -12,22 +12,18 @@ class HeartbeatChecker(object):
     heartbeat is received before the expected timeout expires.
 
     """
-    # Note: even though we're sending heartbeats in half the specified
-    # timeout value, the broker will be sending them to us at the specified
-    # value. This means we'll be checking for an idle connection via send_and_check
-    # twice as many times as the broker will send heartbeats to us,
-    # so we need to set max idle count to 4 here
     _MAX_IDLE_COUNT = 4
     _STALE_CONNECTION = "Too Many Missed Heartbeats, No reply in %i seconds"
 
-    def __init__(self, connection, timeout):
+    def __init__(self, connection, interval):
         """Create a heartbeat on the connection that sends two heartbeat frames
         within the specified timeout window. Also checks to ensure heartbeats
         are received from the broker.
 
         :param pika.connection.Connection: Connection object
-        :param int timeout: Heartbeat check timeout. Note: heartbeats will
-                            actually be sent at timeout / 2 frequency.
+        :param int interval: Heartbeat check interval. Note: heartbeats will
+                             actually be sent at interval / 2 frequency, and
+                             heartbeak checks at (interval * 3) / 2.
 
         """
         self._connection = connection
--- a/pika/heartbeat.py
+++ b/pika/heartbeat.py
@@ -8,8 +8,10 @@ LOGGER = logging.getLogger(__name__)
 
 
 class HeartbeatChecker(object):
-    """Sends heartbeats to the broker and checks to make sure that the broker's
-    heartbeat is received before the expected timeout expires.
+    """Sends heartbeats to the broker twice during the provided interval
+    and checks at a different interval to make sure that the broker's
+    heartbeat is received as expected. See the parameter list for more
+    details.
 
     """
     _MAX_IDLE_COUNT = 2
@@ -17,18 +19,19 @@ class HeartbeatChecker(object):
 
     def __init__(self, connection, interval):
         """Create a heartbeat on the connection that sends two heartbeat frames
-        within the specified timeout window. Also checks to ensure heartbeats
-        are received from the broker.
+        within the specified interval. Also checks to ensure heartbeats are
+        received from the broker at a different interval explained below.
 
         :param pika.connection.Connection: Connection object
         :param int interval: Heartbeat check interval. Note: heartbeats will
                              actually be sent at interval / 2 frequency, and
-                             heartbeat checks made at (interval * 3) / 2. Default
+                             heartbeat checks made at interval. The default
                              interval is 60 seconds so Pika will send heartbeats
-                             at 30 second intervals and will check every 90 seconds.
+                             at 30 second intervals and will check every 60 seconds.
                              If no heartbeat is received from the broker, nor data
-                             activity on the connection after 180 seconds the
-                             connection will be assumed dead and closed.
+                             activity on the connection after 120 seconds (two
+                             intervals, what RabbitMQ uses) the connection will be
+                             assumed dead and closed.
 
         """
         self._connection = connection
@@ -37,8 +40,13 @@ class HeartbeatChecker(object):
         # https://www.rabbitmq.com/heartbeats.html#heartbeats-timeout
         # https://github.com/pika/pika/pull/1072
         # https://groups.google.com/d/topic/rabbitmq-users/Fmfeqe5ocTY/discussion
+        # There is a certain amount of confusion around how client developers
+        # interpret the spec. The spec talks about 2 missed heartbeats as a
+        # *timeout*, plus that any activity on the connection counts for a
+        # heartbeat. This is to avoid edge cases and not to depend on network
+        # latency.
         self._send_interval = float(interval) / 2
-        self._check_interval = (float(interval) * 3) / 2
+        self._check_interval = interval
 
         # Initialize counters
         self._bytes_received = 0
@@ -49,18 +57,8 @@ class HeartbeatChecker(object):
 
         self._send_timer = None
         self._check_timer = None
-
-        self._setup_timers()
-
-    @property
-    def active(self):
-        """Return True if the connection's heartbeat attribute is set to this
-        instance.
-
-        :rtype True
-
-        """
-        return self._connection._heartbeat_checker is self  # pylint: disable=W0212
+        self._start_send_timer()
+        self._start_check_timer()
 
     @property
     def bytes_received_on_connection(self):
@@ -84,7 +82,7 @@ class HeartbeatChecker(object):
         LOGGER.debug('Received heartbeat frame')
         self._heartbeat_frames_received += 1
 
-    def send_heartbeat(self):
+    def _send_heartbeat(self):
         """Invoked by a timer to send a heartbeat when we need to.
 
         """
@@ -92,7 +90,7 @@ class HeartbeatChecker(object):
         self._send_heartbeat_frame()
         self._start_send_timer()
 
-    def check_heartbeat(self):
+    def _check_heartbeat(self):
         """Invoked by a timer to check for broker heartbeats. Checks to see
         if we've missed any heartbeats and disconnect our connection if it's
         been idle too long.
@@ -149,7 +147,7 @@ class HeartbeatChecker(object):
         :rtype: bool
 
         """
-        return not self._bytes_received == self.bytes_received_on_connection
+        return self._bytes_received != self.bytes_received_on_connection
 
     @staticmethod
     def _new_heartbeat_frame():
@@ -169,40 +167,17 @@ class HeartbeatChecker(object):
             self._new_heartbeat_frame())
         self._heartbeat_frames_sent += 1
 
-    def _setup_timers(self):
-        """Use the connection objects _adapter_add_timeout function which is
-        implemented by the Adapter for calling our send and check functions
-        after interval seconds.
-
-        """
-        self._setup_send_timer()
-        self._setup_check_timer()
-
-    def _setup_send_timer(self): # pylint: disable=C0111
+    def _start_send_timer(self):
+        """Start a new heartbeat send timer."""
         self._send_timer = self._connection._adapter_add_timeout(  # pylint: disable=W0212
             self._send_interval,
-            self.send_heartbeat)
-
-    def _setup_check_timer(self): # pylint: disable=C0111
-        self._check_timer = self._connection._adapter_add_timeout(  # pylint: disable=W0212
-            self._check_interval,
-            self.check_heartbeat)
-
-    def _start_send_timer(self):
-        """If the connection still has this object set for heartbeats, add a
-        new send timer.
-
-        """
-        if self.active:
-            self._setup_send_timer()
+            self._send_heartbeat)
 
     def _start_check_timer(self):
-        """If the connection still has this object set for heartbeats, add a
-        new check timer.
-
-        """
-        if self.active:
-            self._setup_check_timer()
+        """Start a new heartbeat check timer."""
+        self._check_timer = self._connection._adapter_add_timeout(  # pylint: disable=W0212
+            self._check_interval,
+            self._check_heartbeat)
 
     def _update_counters(self):
         """Update the internal counters for bytes sent and received and the
--- a/pika/heartbeat.py
+++ b/pika/heartbeat.py
@@ -57,7 +57,8 @@ class HeartbeatChecker(object):
         # to two times the heartbeat timeout value.
         # https://github.com/pika/pika/pull/1072#issuecomment-397850795
         # https://github.com/rabbitmq/rabbitmq-java-client/blob/b55bd20a1a236fc2d1ea9369b579770fa0237615/src/main/java/com/rabbitmq/client/impl/AMQConnection.java#L773-L780
-        self._check_interval = float(timeout) * 2;
+        # https://github.com/ruby-amqp/bunny/blob/3259f3af2e659a49c38c2470aa565c8fb825213c/lib/bunny/session.rb#L1187-L1192
+        self._check_interval = float(timeout) * 2
 
         LOGGER.debug('timeout: %f send_interval: %f check_interval: %f',
                      self._timeout,
--- a/pika/heartbeat.py
+++ b/pika/heartbeat.py
@@ -9,37 +9,32 @@ LOGGER = logging.getLogger(__name__)
 
 class HeartbeatChecker(object):
     """Checks to make sure that our heartbeat is received at the expected
-    intervals.
+    timeouts.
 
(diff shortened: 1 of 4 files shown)


pika/heartbeat.py
```diff
@@ -9,37 +9,32 @@
 
 class HeartbeatChecker(object):
     """Checks to make sure that our heartbeat is received at the expected
```
vitaly-krugl: ==> "Send heartbeats to the broker and checks to make sure that broker's heartbeat is received before the expected timeout expires." ?

pika/heartbeat.py
```diff
@@ -9,37 +9,32 @@
...
-    DEFAULT_INTERVAL = 60
-    MAX_IDLE_COUNT = 2
-
+    # Note: even though we're sending heartbeats in half the specified
+    # timeout value, the broker will be sending them to us at the specified
+    # value. This means we'll be checking for an idle connection via send_and_check
+    # twice as many times as the broker will send heartbeats to us,
+    # so we need to set max idle count to 4 here
+    _MAX_IDLE_COUNT = 4
+    _CONNECTION_FORCED = 320
```
vitaly-krugl: _CONNECTION_FORCED is not used.

pika/heartbeat.py
```diff
@@ -9,37 +9,32 @@
...
     """
-    DEFAULT_INTERVAL = 60
-    MAX_IDLE_COUNT = 2
-
+    # Note: even though we're sending heartbeats in half the specified
+    # timeout value, the broker will be sending them to us at the specified
+    # value. This means we'll be checking for an idle connection via send_and_check
+    # twice as many times as the broker will send heartbeats to us,
+    # so we need to set max idle count to 4 here
+    _MAX_IDLE_COUNT = 4
```
vitaly-krugl: What made you switch from 2 to 4? Not saying it's wrong, just curious?

pika/heartbeat.py
```diff
@@ -9,37 +9,32 @@
...
+    _CONNECTION_FORCED = 320
     _STALE_CONNECTION = "Too Many Missed Heartbeats, No reply in %i seconds"
 
-    def __init__(self, connection, interval=DEFAULT_INTERVAL, idle_count=MAX_IDLE_COUNT):
+    def __init__(self, connection, timeout):
         """Create a heartbeat on connection sending a heartbeat frame every
```
vitaly-krugl: This description sounds wrong. It should send a heartbeat multiple times within the timeout window. And it will also be checking for timeout of heartbeats from the broker.

pika/heartbeat.py
```diff
@@ -9,37 +9,32 @@
...
         # https://www.rabbitmq.com/heartbeats.html#heartbeats-timeout
-        self._interval = float(interval / 2)
-
-        # Note: even though we're sending heartbeats in half the specified
-        # interval, the broker will be sending them to us at the specified
-        # interval. This means we'll be checking for an idle connection
-        # twice as many times as the broker will send heartbeats to us,
-        # so we need to double the max idle count here
-        self._max_idle_count = idle_count * 2
+        self._timeout = float(timeout / 2)
```
vitaly-krugl: The math counts don't work out with this PR: division by 2 versus _MAX_IDLE_COUNT. `self._timeout` is just timeout without any division. Cast it to a float, though, to make division correct on python 2.x later when calculating the delay value to pass to _adapter_add_timeout(). In other words: 
```
self._timeout = float(timeout)
```

Or, if you want to store pre-computed "check interval delay", then 
```
self._check_interval = float(timeout) / HeartbeatChecker._MAX_IDLE_COUNT
```
Note that having float wrapping `timeout` turns the division into floating point op on python 2.x too.

In subsequent comments, I will assume you are using `self._check_interval`

pika/heartbeat.py
```diff
@@ -172,7 +167,7 @@ def _setup_timer(self):
 
         """
         self._timer = self._connection._adapter_add_timeout(  # pylint: disable=W0212
-            self._interval,
+            self._timeout,
```
vitaly-krugl: Pass `self._check_interval` instead of `self._timeout`.

pika/heartbeat.py
```diff
@@ -129,7 +124,7 @@ def _close_connection(self):
         """Close the connection with the AMQP Connection-Forced value."""
         LOGGER.info('Connection is idle, %i stale byte intervals',
                     self._idle_byte_intervals)
-        duration = self._max_idle_count * self._interval
+        duration = HeartbeatChecker._MAX_IDLE_COUNT * self._timeout
```
vitaly-krugl: Don't need this since will be using `self._check_interval` in _adapter_add_timeout() below.

tests/unit/heartbeat_tests.py
```diff
@@ -55,31 +55,28 @@ def _adapter_remove_timeout(self):
 
 class HeartbeatTests(unittest.TestCase):
 
-    INTERVAL = 60
-    HALF_INTERVAL = INTERVAL / 2
+    TIMEOUT = 60
```
vitaly-krugl: Need to fix up tests to match the corrected implementation.

pika/heartbeat.py
lukebakken: Here's what I observed. A 60 second negotiated timeout will result in RabbitMQ sending heartbeats every 60 seconds. Since we changed Pika to send heartbeats every `timeout / 2` seconds (to prevent edge cases that can end up with RMQ closing a connection), heartbeats will be sent to RMQ every 30 seconds. We also check for an idle connection in `send_and_check` in Pika, so we'll be checking twice as often. Once, when testing my change, the timing was just right that the idle count reached 2 and Pika erroneously thought the connection was idle. So, I think it makes sense to basically double the idle count to go with halving the heartbeat timeout.

pika/heartbeat.py
lukebakken: Great, `_check_interval` is a much better name.

lukebakken:
@vitaly-krugl this should be all set. I tried to fix this pylint error but the hassle involved in getting the mocks fixed for the tests is taking too much time -

```
$ pylint pika/heartbeat.py
Using config file /home/lbakken/development/pika/pika/pylintrc
************* Module pika.heartbeat
W0212,  61:15 - Access to a protected member _heartbeat_checker of a client class (protected-access)
```

vitaly-krugl:
@lukebakken - It looks like RabbitMQ's heartbeat-emission functionality is broken as it's inconsistent with RabbitMQ's own documentation, and implements the heartbeat protocol asymmetrically. I posted on rabbitmq-users about it: https://groups.google.com/forum/#!topic/rabbitmq-users/Fmfeqe5ocTY.

I recommend adopting the following strategy - use two timers in HeartbeatChecker:
1. Heartbeat Transmit timer: fires every `float(timeout) / 2` seconds (note the float wrapped only around  timeout, not the entire quotient :) ). This is compatible with https://www.rabbitmq.com/heartbeats.html#heartbeats-timeout: "Heartbeat frames are sent about every timeout / 2 seconds"
2. Data received from broker timer: fires every `float(timeout) * 3 / 2` seconds. Each time this timer fires, HertbeatChecker checks whether any data has been received since last time. This provides for sufficient additional time buffer to make sure we don't miss RabbitMQ's heartbeat to work around the noted flaw in RabbitMQ when there isn't any other data traffic from the broker without taking the full 2x time to detect dead connections.

pika/heartbeat.py
lukebakken: I will re-run my test with Wireshark, but I'm pretty sure my observation is correct when a 60 second heartbeat is used with an app that uses the Java client.

pika/heartbeat.py
```diff
@@ -8,38 +8,34 @@
...
 
     """
-    DEFAULT_INTERVAL = 60
-    MAX_IDLE_COUNT = 2
-
+    # Note: even though we're sending heartbeats in half the specified
+    # timeout value, the broker will be sending them to us at the specified
+    # value. This means we'll be checking for an idle connection via send_and_check
+    # twice as many times as the broker will send heartbeats to us,
+    # so we need to set max idle count to 4 here
```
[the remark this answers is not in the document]
lukebakken: I understand what the docs say, but I am trying to match what the Java client does.

lukebakken:
@vitaly-krugl - thanks, I'll implement that heartbeat solution.

pika/heartbeat.py
```diff
@@ -8,38 +8,37 @@
...
-    DEFAULT_INTERVAL = 60
-    MAX_IDLE_COUNT = 2
-
+    _MAX_IDLE_COUNT = 2
     _STALE_CONNECTION = "Too Many Missed Heartbeats, No reply in %i seconds"
 
-    def __init__(self, connection, interval=DEFAULT_INTERVAL, idle_count=MAX_IDLE_COUNT):
-        """Create a heartbeat on connection sending a heartbeat frame every
-        interval seconds.
+    def __init__(self, connection, interval):
```
vitaly-krugl: @lukebakken - help, I am confused! :) We have been discussing this for some time and agreed to rename the arg interval to timeout, and it's now back to interval. Interval is meaningless here because we shouldn't use it as interval between anything, and timeout has the correct semantics.

pika/heartbeat.py
```diff
@@ -165,23 +169,40 @@ def _send_heartbeat_frame(self):
...
+        after interval seconds.
+
+        """
+        self._setup_send_timer()
+        self._setup_check_timer()
+
+    def _setup_send_timer(self): # pylint: disable=C0111
+        self._send_timer = self._connection._adapter_add_timeout(  # pylint: disable=W0212
+            self._send_interval,
+            self.send_heartbeat)
```
vitaly-krugl: send_hearbeat and check_heartbeat should be private methods with leading underscore in the names.

pika/heartbeat.py
```diff
@@ -165,23 +169,40 @@ def _send_heartbeat_frame(self):
...
-    def _start_timer(self):
+    def _start_check_timer(self):
         """If the connection still has this object set for heartbeats, add a
-        new timer.
+        new check timer.
 
         """
         if self.active:
-            self._setup_timer()
+            self._setup_check_timer()
```
vitaly-krugl: There is unnecessary complication from the super-granular separation between _start_check_timer/_setup_check_timer and _start_send_timer/_setup_send_timer as well as _setup_timers. Let's just have _start_check_timer and _start_send_timer and no `self.active` check (as explained in the following paragraph)

And the `self.active` checks are unnecessary since `Connection` calls the heartbeat's `stop()` method when it's not needed any more, and the stop() cancels the timeouts. So, let's get rid of the redundant `HearbeatChecker.active` property as well.

These unnecessary legacy things are an unwelcome distraction for maintainers. :)

pika/heartbeat.py
```diff
@@ -8,38 +8,37 @@
...
-    """Checks to make sure that our heartbeat is received at the expected
-    intervals.
+    """Sends heartbeats to the broker and checks to make sure that the broker's
+    heartbeat is received before the expected timeout expires.
 
     """
-    DEFAULT_INTERVAL = 60
-    MAX_IDLE_COUNT = 2
-
+    _MAX_IDLE_COUNT = 2
```
vitaly-krugl: There is no use for _MAX_IDLE_COUNT any longer since we configure our check interval as `self._check_interval = (float(timeout) * 3) / 2` (i.e.   timeout * 1.5) to account for latencies already. Otherwise, with _MAX_IDLE_COUNT of 2, a negotiated timeout of 10 minutes becomes a 30-minute timeout (10 * 1.5 * 2), a bit much, right? If you're an app developer and you limited timeout to x seconds to detect dead connections in some reasonable time, but pike obliged 3x later, wouldn't you and your SLA be fuming right now? :) So, no more _MAX_IDLE_COUNT.

pika/heartbeat.py
```diff
@@ -8,38 +8,37 @@
...
-        # Note: even though we're sending heartbeats in half the specified
-        # interval, the broker will be sending them to us at the specified
-        # interval. This means we'll be checking for an idle connection
-        # twice as many times as the broker will send heartbeats to us,
-        # so we need to double the max idle count here
-        self._max_idle_count = idle_count * 2
+        # https://github.com/pika/pika/pull/1072
+        # https://groups.google.com/d/topic/rabbitmq-users/Fmfeqe5ocTY/discussion
+        self._send_interval = float(interval) / 2
+        self._check_interval = (float(interval) * 3) / 2
```
vitaly-krugl: Not sure why I originally suggested the `* 3 / 2` math. My bad. It would be more readable as `timeout * 1.5` and floating point math would be happy on both 2.x and 3.x too.

lukebakken: Alright, and if there's no activity when a check happens, close the connection.

pika/heartbeat.py
lukebakken: OK, `send_and_check` used to be public, so I kept it that way.

pika/heartbeat.py
lukebakken: Maybe we have a different understanding of an interval. To me, an "interval of 60 seconds" means that every 60 seconds something happens, which describes what we do - send and check heartbeats (using our differently-computed values). Is that not your understanding of "interval"?

michaelklishin: RabbitMQ server will treat the negotiated period of time as a timeout (2 intervals), regardless of what it's named in the client.

lukebakken: Thanks @michaelklishin I will use the same logic here. Two "heartbeat check" intervals without a heartbeat from the server or activity on the channel will be considered a timeout.

michaelklishin:
The changes move Pika much closer to the Java client and look reasonable. I tried the following script, which is a modification of tutorial 1's consumer:

``` python
#!/usr/bin/env python
import pika

connection = pika.BlockingConnection(pika.ConnectionParameters(
        host='localhost', heartbeat = 10))

channel.queue_declare(queue='hello')

def callback(ch, method, properties, body):
    print(" [x] Received %r" % body)

channel.basic_consume('hello', callback, auto_ack = True)

print(' [*] Waiting for messages. To exit press CTRL+C')
channel.start_consuming()
```

Connection heartbeat timeout is negotiated as 10 but after 40 to 60 seconds I reliably get a missed heartbeat exception from Pika:

```
PYTHONPATH=. ./debug/idle_consumer.py                                                                                                                                                                 1 ↵
 [*] Waiting for messages. To exit press CTRL+C
  File "./debug/idle_consumer.py", line 17, in <module>
    channel.start_consuming()
  File "/Users/antares/Development/RabbitMQ/pika.git/pika/adapters/blocking_connection.py", line 1886, in start_consuming
    self._process_data_events(time_limit=None)
  File "/Users/antares/Development/RabbitMQ/pika.git/pika/adapters/blocking_connection.py", line 2048, in _process_data_events
    self.connection.process_data_events(time_limit=time_limit)
  File "/Users/antares/Development/RabbitMQ/pika.git/pika/adapters/blocking_connection.py", line 840, in process_data_events
    self._flush_output(common_terminator)
  File "/Users/antares/Development/RabbitMQ/pika.git/pika/adapters/blocking_connection.py", line 533, in _flush_output
    raise self._closed_result.value.error
pika.exceptions.AMQPHeartbeatTimeout: No activity or too many missed meartbeats in the last 10 seconds
```

Here's a [capture of the heartbeat frames sent on the connection](https://gist.github.com/michaelklishin/26936bd64d97c897aa79158519993f8c) in text form.

lukebakken:
@michaelklishin I think I know what's up. There's a slim chance that due to timing Pika will think it's missed an heartbeat from RMQ when the heartbeat was just about to arrive. I'll see what I can do.

lukebakken:
@michaelklishin [this](https://github.com/rabbitmq/rabbitmq-java-client/blob/b55bd20a1a236fc2d1ea9369b579770fa0237615/src/main/java/com/rabbitmq/client/impl/AMQConnection.java#L773-L780) is the code in the Java client that detects missed server heartbeats. Note that the interval used is **twice** the negotiated timeout value ("two complete heartbeat setting intervals"). So, my latest commit matches this behavior. If a user sets a 10 minute heartbeat timeout with this change a dead connection won't be detected for 20 minutes (just as the Java client does).

@vitaly-krugl has commented that this behavior is a bit undesirable. Another alternative is, instead of doubling the negotiated value, we just add a fixed amount like 5 seconds. I tested this and it resolves the issue reported [here](https://github.com/pika/pika/pull/1072#issuecomment-397850795).

michaelklishin:
My point was that both server and Pika had heartbeats enabled and for several timeout periods the connection was considered to be alive even without any activity (it definitely was fine). Bunny adds a fixed number and does not allow heartbeat timeout to be configured < ~ 1 second. It also can run for as long as I don't shut it down with a similar test.

lukebakken:
@michaelklishin well at least _I'm_ no longer confused by the `timeout` vs `interval` debate 😄

lukebakken:
@michaelklishin -

> Bunny adds a fixed number...

It looks like Bunny multiplies the negotiated heartbeat timeout by `2.2` [here](https://github.com/ruby-amqp/bunny/blob/3259f3af2e659a49c38c2470aa565c8fb825213c/lib/bunny/session.rb#L1187-L1192) to set the socket read timeout value, which is pretty much equivalent to the change I have proposed in this PR.

lukebakken:
Thanks @vitaly-krugl I think we came to the right set of changes.

vitaly-krugl:
Looks good. I wish RabbitMQ would change to a symmetric scheme. There would
be no harm if it sent heartbeats twice per negotiated heartbeat. :)

PR #1090, opened 2018-06-28
Merge stable into master

lukebakken:
This ensures all `stable` changes are in `master` and will allow us to more easily make future changes based on `stable` that are forward-ported to `master`.


lukebakken:
Thanks @michaelklishin

Issue #1084, opened 2018-06-23
connection.remove_timeout behavior in 0.12.0


wcpettus:
Hi pika,

I have used the `connection.remove_timeout` in a way that works as of 0.11.2 and earlier, but breaks in 0.12.0.  Not sure if this is my abusive use, intentional new behavior, or inadvertent new edge case.  I appreciate your comments and assistance.

```python
import pika
connection = pika.BlockingConnection()
connection.remove_timeout(None)
```

On 0.11.2 that just works.  On 0.12.0, I get the following traceback:
```python
>>> connection.remove_timeout(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../pika/adapters/blocking_connection.py", line 684, in remove_timeout
    self._impl.remove_timeout(timeout_id)
  File ".../pika/adapters/base_connection.py", line 137, in remove_timeout
    self.ioloop.remove_timeout(timeout_id)
  File ".../pika/adapters/select_connection.py", line 372, in remove_timeout
    self._timer.remove_timeout(timeout_id)
  File ".../pika/adapters/select_connection.py", line 216, in remove_timeout
    if timeout.callback is None:
AttributeError: 'NoneType' object has no attribute 'callback'
```

Mostly tested on python2.7, but the crash is reproducible in python3.

michaelklishin:
I'm not sure what the intent of your code is but `remove_timeout` assumes its argument is a timer. I somewhat doubt it is a part of the public API. Consider removing that line in your code or explaining why exactly you need it.

michaelklishin:
I now see in #1086 that the same practice that's used for all RabbitMQ distributions subprojects and some clients has been adopted for Pika: questions belong to the mailing list, either [`rabbitmq-users`](https://groups.google.com/forum/#!forum/rabbitmq-users) or [pika-python](https://groups.google.com/forum/?fromgroups#!forum/pika-python).

I do not see [any examples that use `remove_timeout`](https://github.com/pika/pika/search?q=remove_timeout&unscoped_q=remove_timeout). The most logical explanation I can think of is: it's not a part of the public API, even if the function happens to be public.

lukebakken:
Hi @wcpettus I would gladly review a pull request that addresses this issue. Please target the `stable` branch.

vitaly-krugl:
[...]

PR #1094, opened 2018-07-07
Allow None argument for remove_timeout

wcpettus:
Fix for #1084 

I tried to make the minimal modification while allowing a `timeout=None` argument to `_Timer.remove_timeout`.
- This seemed the best place to put it because this is the end of the `remove_timout` cascade of calls and where all the logic happens.
- A targeted exception for the NoneType seems most inline with the rest of the code.

yapf is happy, no coverage change


vitaly-krugl:
Thank you for your contribution, but we don't want to accept bogus values passed to the APIs. If the application uses None as a sentinel value, it's up to the application to check for it and not call `remove_timeout()` if it's None. It's the application's responsibility to do so. It's bad precedent and a slippery slope for an API to accept bogus values.

Issue #1093, opened 2018-07-06
heartbeat connection parameter has no effect


carlosefr:
The [docs](https://pika.readthedocs.io/en/latest/modules/parameters.html) have this to say about the `heartbeat` connection parameter:

> AMQP connection heartbeat timeout value for negotiation during connection tuning or callable which is invoked during connection tuning. None to accept broker’s value. 0 turns heartbeat off. Defaults to DEFAULT_HEARTBEAT_TIMEOUT.

However, `heartbeat` doesn't seem to work as expected. It has no effect whatsoever, but the timeout suggested by the server isn't being honored either.

Consider the test case below, a publisher publishing messages at the rate of about 1 msg/sec, and the following settings:

  * `HEARTBEAT=60` / `WAIT_TIME=50`: exits on the first message.
  * `HEARTBEAT=60` / `WAIT_TIME=20`: exits on the second message.
  * `HEARTBEAT=60` / `WAIT_TIME=10`: no longer exits

Setting `HEARTBEAT` to `0` or `3600` changes nothing: the behavior is the same as above. The same with using Pika **0.12.0** or **0.11.2**.

Also, changing the heartbeat value in RabbitMQ has no effect (and I've confirmed that the client gets the changed value by using a callable instead of an integer for the `heartbeat` parameter):

```
%% /etc/rabbitmq/rabbitmq.config
[
  {rabbit,
    [
      {heartbeat, 90}
    ]
  }
].
```

I'm using RabbitMQ **3.7.4** and Python **3.6.5** on Linux for all these tests.

## Test Case

```python3
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

import time
import pika

HEARTBEAT=60
WAIT_TIME=25

def receive_message(channel, method, properties, body):
    print("Got message! Sleeping %d seconds" % WAIT_TIME)
    time.sleep(WAIT_TIME)
    channel.basic_ack(delivery_tag=method.delivery_tag)
    print("Acknowledged message! Timeout NOT triggered.")

print("Connecting: heartbeat=%d" % HEARTBEAT)
cs = pika.PlainCredentials("guest", "guest")
ps = pika.ConnectionParameters(host="127.0.0.1", credentials=cs, heartbeat=HEARTBEAT)
conn = pika.BlockingConnection(ps)
chan = conn.channel()

chan.exchange_declare(exchange="example_exchange", exchange_type="topic", durable=True)
chan.queue_declare(queue="example_queue", durable=True)
[...]

PR #1097, opened 2018-07-09
Make it clear that heartbeats are not the only source of connection timeouts

carlosefr:
The `heartbeat` parameter documentation doesn't make it clear that the broker may still close a connection due to inactivity even if heartbeats are disabled (or large). Without prior knowledge, this behavior is kind of unexpected (but makes sense once you're aware of it).

This PR adds a note about this to the documentation. See also #1093 and #753.

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -331,6 +331,12 @@ class Parameters(object):  # pylint: disable=R0902
             connection tuning or callable which is invoked during connection tuning.
             None to accept broker's value. 0 turns heartbeat off. Defaults to
             `DEFAULT_HEARTBEAT_TIMEOUT`.
+
+        Note: The connection may still be timed out by the broker, regardless of the
+              heartbeat setting, if the client takes too long to accept new messages.
+              Setting `prefetch_count` to 1 on the channel is one way to prevent this
+              from happening. See the `Channel.basic_qos()` method.
+
         :rtype: integer, None or callable
 
         """


lukebakken:
Let's not have an explicit prefetch recommendation, and we could link to my comment [here](https://github.com/pika/pika/issues/753#issuecomment-318119222) or expand on that comment and add it to the Pika / RabbitMQ docs. What do you think @michaelklishin ?

pika/connection.py
```diff
@@ -331,6 +331,12 @@ def heartbeat(self):
             connection tuning or callable which is invoked during connection tuning.
             None to accept broker's value. 0 turns heartbeat off. Defaults to
             `DEFAULT_HEARTBEAT_TIMEOUT`.
+
+        Note: The connection may still be timed out by the broker, regardless of the
+              heartbeat setting, if the client takes too long to accept new messages.
+              Setting `prefetch_count` to 1 on the channel is one way to prevent this
```
vitaly-krugl: @michaelklishin, do you agree?

michaelklishin: I'm afraid I don't. Client responsiveness is not tied to a particular prefetch setting and 1 is a very conservative value. Again, I primarily dislike the incredible generality of the language here.

michaelklishin:
Closing since there is no consensus about this specific wording. We'd consider an alternative PR, of course.

PR #1108, opened 2018-08-01
Fix Twisted tests

abompard:
A couple fixes to the Twisted adapter and example. And a commit to make tests pass on Python2.

Follow-up of PR #1069 1069.


tests/acceptance/twisted_adapter_tests.py
```diff
@@ -108,6 +108,16 @@ class TwistedChannelTestCase(TestCase):
...
+        # This is only needed on Python2 for functools.wraps to work.
+        wrapped = (
+            "basic_cancel", "basic_get", "basic_qos", "basic_recover",
+            "exchange_bind", "exchange_unbind", "exchange_declare",
+            "exchange_delete", "confirm_delivery", "flow",
+            "queue_bind", "queue_declare", "queue_delete", "queue_purge",
+            "queue_unbind", "tx_commit", "tx_rollback", "tx_select",
+        )
+        for meth_name in wrapped:
+            getattr(self.pika_channel, meth_name).__name__ = meth_name
```
lukebakken: Aha! Thank you.

lukebakken:
Thanks again, @abompard

Zephor5:
hi, i see this breaks `TwistedProtocolConnection` ready defer result, from the connection instance to None, why ?

Zephor5:
i have a use case like:
``` python
    def __connect(self, retrying=False):
        params = self.__params
        cc = ClientCreator(self.loop, TwistedProtocolConnection, params)
        _d = cc.connectTCP(params.host, params.port, timeout=self.timeout_conn)
        _d.addCallback(lambda p: p.ready)
        _d.addCallbacks(self._in_pool,
                        lambda err: err if retrying or not self.retry else self.__connect(True))  # retry once when err
        return _d

    def _in_pool(self, conn):
```

that would make it difficult to pass the connection into `_in_pool` function

lukebakken:
@Zephor5 it would help a lot to provide a complete, runnable code sample to demonstrate your issue, and if you have a fix.

The people who maintain Pika do so as volunteers so anything you can do to help save time and remove guessing will help a lot. I am not as familiar with Twisted which makes a complete example that much more important.

PR #1111, opened 2018-08-21
Twisted channel: relay the consumer_tags property

abompard:
After the Twisted refactoring, a `consumer_tags` property was no longer exposed. This commit fixes it.


abompard:
I'm not sure why the tests failed on Python 2.7, it seems unrelated.

PR #1113, opened 2018-09-04
Add an example for Twisted authentication with x509 certs

jeremycline:
Add an example on how to perform mutual authentication with x509 client
certificates. 

@abompard and I had to fiddle a bit to get client authentication working with Twisted, so I thought it'd be good to add an example in the docs for it.


jeremycline:
Ahh, I lost track of this. It looks like @lukebakken added a note on setting up x509-based authentication (thanks!). Is there anything else you'd like included here?

lukebakken:
We're good! I was just waiting on the tests.

PR #1123, opened 2018-10-02
Examples using rabbitmq extensions

deslum:
I created pull requests with examples using rabbitmq extensions.

https://www.rabbitmq.com/extensions.html


lukebakken:
Thank you for using Pika!

While this PR is interesting, it also represents a lot of extra code we would have to maintain. I would rather see this distilled into a markdown document with small code snippets demonstrating each RabbitMQ "extension".

Issue #1126, opened 2018-10-15
Async is a keyword


Str-Gen:
Async is a keyword in Python 3.7.
the libev_connection.py file uses async in a manner that is no longer allowed, resulting in a crash.
I've read issue 1091 and a suitable fix is in pull request. This would re-add support for libev.

./adapters/libev_connection.py:105:        self.async = None
./adapters/libev_connection.py:146:            self.async = pyev.Async(self.ioloop, self._noop_callable)
./adapters/libev_connection.py:147:            self.async.start()

Str-Gen:
Unfortunately I currently don't have the time nor the familiarity with the project and its dependencies to provide the extensive testing code, requested by the authors.

Issue #1091, opened 2018-06-30
pika.adapters.libev_connection.LibevConnection is incompatible with Python 3.7


sky-code:
pika does not work with python 3.7.
```
  File "/Users/macbook/virtualenvs/natrix_env/lib/python3.7/site-packages/pika/__init__.py", line 22, in <module>
    from pika.adapters import BaseConnection
  File "/Users/macbook/virtualenvs/natrix_env/lib/python3.7/site-packages/pika/adapters/__init__.py", line 39, in <module>
    from pika.adapters.libev_connection import LibevConnection
  File "/Users/macbook/virtualenvs/natrix_env/lib/python3.7/site-packages/pika/adapters/libev_connection.py", line 105
    self.async = None
             ^
SyntaxError: invalid syntax
```
starting from python 3.7 `async` [is keyword](https://docs.python.org/3/whatsnew/3.7.html), rename this attribute for fix compatibility with 3.7

michaelklishin:
@sky-code that's a little light on details. I don't see LibevConnection mentioned anywhere in Pika master. Other adaptors don't seem to use `async` for attribute name or similar. Please provide a way to reproduce and save the maintainers some time.

sky-code:
this is because I install dramatiq library
```
pip install dramatiq[rabbitmq,watch]==1.2.0
```

lukebakken:
@sky-code if you have time to submit a PR that addresses new keywords in Python 3.7 like `async`, I would review it. Thanks!

sky-code:
where is this file ?
https://github.com/pika/pika/commits/d4bdaaa75805cb7899ed53cf684bbaa8bf026449/pika/adapters/libev_connection.py

why if i install clear `pika` package i not have this file in `site-packages/pika/adapters` but when pika installed with additional packages, this file appears from nowhere?

lukebakken:
@sky-code I apologize, I forgot that `libev` support was removed in 3ce691b01bc567531aee96c0fd55d94bc3226061

If there is enough interest, we would consider re-adding it to Pika `1.0.0` and would welcome a PR to do so.

PR #1127, opened 2018-10-15
Recreate libev_connection.py (python 3.7)

Str-Gen:
Follow-up on issue #1091 and issue #1126.
This would re-add libev support. The change is minor and results in proper execution on python 3.7.0.


lukebakken:
Please ensure that this code will be tested. Thanks.

PR #1129, opened 2018-10-18
pika: Don't automatically import twisted and tornado.

### LINKED ISSUE #1128: Dramatically improve import time for pika when `twisted` is available in the Python environment
First, since this is a long and complex issue report, I want to mention that I've been using Pika to interact with RabbitMQ for [Zulip](https://github.com/zulip/zulip/) since near the beginning, and it's been a great experience, so thanks for the library!

Currently, the import time for `pika` is really high in environments where large third-party libraries like Twisted are installed, because the `pika.adapters` system automatically imports those large libraries, even if they aren't actually being used.

In particular, on my very fast macine, in a virtualenv with both twisted and tornado installed, importing pika takes 200ms:
```
import time; x = time.time() ; import pika; print(time.time() - x)
0.19557762145996094
```

Whereas in a virtualenv without Twisted installed (but with Tornado installed), it's a bit over 50ms:
```
import time; x = time.time() ; import pika; print(time.time() - x)
0.053652048110961914
```

And in a virtualenv without either, it takes just 43ms, so 5x better.  

The cause of this is the code in `pika.adapters`, which automatically imports `twisted` and `tornado` if they are present as part of its auto-detection of Twisted/Tornado support.

This is a pretty significant issue for Zulip; we have Twisted and Tornado installed in our main virtualenv, but only use them for some small/side parts of the project.  So the import time for a Zulip process ends up being 150ms more than it needs to be because we end up importing both Twisted and Tornado unnecessarily (see: https://github.com/zulip/zulip/issues/9953 for more details on our import time optimization efforts).  While import time isn't a big deal for long-running processes, it is a big problem for running quick management commands and can cause problems when restarting a server, where one might be starting a dozen processes at the same time (in which case this 150ms becomes seconds of CPU time!).

What I'd like `pika` to do is only import `tornado` and `twisted` if the user actually attempts to import 

We've been able to work around this with the following ugly hack early in our startup process to make importing `pika` not result in importing `Twisted`:
```
    sys.modules['pika.adapters.twisted_connection'] = types.ModuleType(
        'pika.adapters.twisted_connection')
```

But it'd obviously be a lot better if there were a way to import pika without importing these adapters.  I haven't thought deeply about the options for doing this, but one that seems reasonable is to just require users to write `from pika.adapters.tornado_connection import TornadoConnection`, rather than `from pika.adapters import TornadoConnection` (by removing the dynamic auto-import logic in `pika/adapters/__init__.py`), or perhaps better, remove the similar imports from `pika.__init__.py` (so users can't write `from pika import TornadoConnection`, but still can write `from pika import TwistedConnection`).

timabbott:
Previously, the pika.adapters convenience import logic would end up
importing Twisted, Tornado, and AsyncIO if they are present in the
current Python environment, regardless of whether the user intended to
use them.

While the convenience names of e.g. `pika.TornadoConnection` (rather
than the more verbose pika.adapters.TornadoConnection`) were nice,
this had the downside of making importing pika expensive in those
environments; on my system, twisted added 150ms and tornado 10ms to
the based ~40ms import time for pika.

We fix this by removing those convenience imports; now, a user needs
to explicitly import pika.adapters.tornado_connection (aka plan to
actually use the relevant adapter feature) in order to end up
importing those large third-party modules.

Fixes #1128.

---------------------------

Pull request is still work-in-progress; while it's possible I didn't make any mistakes, this is at this point untested.  I opened a PR at this point in development just to get CI to run.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Version History
 
 `GitHub milestone <https://github.com/pika/pika/milestone/8>`_
 
+- `AsyncioConnection`, `TornadoConnection` and `TwistedProtocolConnection` are no longer auto-imported (`PR <https://github.com/pika/pika/pull/1129>`_)
 - `BlockingConnection.consume` now returns `(None, None, None)` when inactivity timeout is reached (`PR <https://github.com/pika/pika/pull/899>`_)
 
 0.12.0 2018-06-19
--- a/README.rst
+++ b/README.rst
@@ -135,11 +135,13 @@ adapter-specific mechanism:
       connection.add_callback_threadsafe(functools.partial(ack_message, channel, delivery_tag))
 
 - When using a non-blocking connection adapter, such as
-  :py:class:`pika.AsyncioConnection` or :py:class:`pika.SelectConnection`, you
-  use the underlying asynchronous framework's native API for requesting an
-  ioloop-bound callback from another thread. For example, `SelectConnection`'s
-  `IOLoop` provides `add_callback_threadsafe()`, `Tornado`'s `IOLoop` has
-  `add_callback()`, while `asyncio`'s event loop exposes `call_soon_threadsafe()`.
+:py:class:`pika.adapters.asyncio_connection.AsyncioConnection` or
+:py:class:`pika.SelectConnection`, you use the underlying asynchronous
+framework's native API for requesting an ioloop-bound callback from
+another thread. For example, `SelectConnection`'s `IOLoop` provides
+`add_callback_threadsafe()`, `Tornado`'s `IOLoop` has
+`add_callback()`, while `asyncio`'s event loop exposes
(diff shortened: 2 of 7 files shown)


timabbott:
OK, I've now tested that it indeed solves the intended problem in Zulip with this 2-commit branch: https://github.com/timabbott/zulip/tree/pika-testing; with that, importing pika is just 35ms on my system.  

https://github.com/timabbott/zulip/commit/014c261e4a64412c05885e2bbf236c55b31585ac is the part of that PR that is the migration for this change; the other commit is just reverting the hack I'd added to work around this issue in Zulip.

Still haven't seen whether CI passes here, so I'm leaving the WIP tag on this PR.

timabbott:
OK, it looks like CI passed (the only failure looks like a coverage infrastructure flake), so this is ready for a review.

PR #1130, opened 2018-10-18
Close consumer queue when the server cancels a consumer.

jothan:
This makes the server cancellation visible simply from running consumer.get().


jothan:
I seem to have duplicated existing code on master, sorry.

PR #1132, opened 2018-10-29
Better parsing of URL parameter "ssl_options"

### LINKED ISSUE #1107: Python 3.7.0 support
There appear to be non-compatible changes in the `ssl` module. Travis CI should be updated as well.

lukebakken:
Due to the fact that `wrap_socket` is deprecated in Python 3.7, we can no longer use the old method of translating URL options to valid SSLContext options.

Fixes #1107

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -9,8 +9,8 @@ import functools
 import logging
 import math
 import numbers
+import os
 import platform
-import socket
 import warnings
 import ssl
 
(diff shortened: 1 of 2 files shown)


vitaly-krugl:
@lukebakken - Shouldn't the following raise ValueError instead of just logging a warning? If I misconfigure something, I would hope to catch that right away via the exception , which is typical for Python code, instead of pretty much hiding it which is what just logging it amounts to:
* https://github.com/pika/pika/pull/1132/files#diff-ab2a915baf59d292f456225a7431ea11R999
* https://github.com/pika/pika/pull/1132/files#diff-ab2a915baf59d292f456225a7431ea11R1009
* https://github.com/pika/pika/pull/1132/files#diff-ab2a915baf59d292f456225a7431ea11R1018

pika/connection.py
```diff
@@ -963,27 +963,64 @@ def _set_url_ssl_options(self, value):
...
-                    self.ssl_options = pika.SSLOptions(
-                        context=ssl_sock.context,
-                        server_hostname=ssl_sock.server_hostname)
-                finally:
-                    ssl_sock.close()
-            finally:
-                sock.close()
-
-
+            # Note: this is the deprecated wrap_socket signature and info:
```
vitaly-krugl: @lukebakken - this comment is confusing. We were using `ssl.SSLSocket` previously, not `wrap_socket`. Did you mean that `ssl.SSLSocket` was deprecated?

pika/connection.py
```diff
@@ -963,27 +963,64 @@ def _set_url_ssl_options(self, value):
...
+            #     ciphers=None
+            cxt = None
+            if 'ca_certs' in opts:
+                opt_ca_certs = opts['ca_certs']
+                if os.path.isfile(opt_ca_certs):
+                    cxt = ssl.create_default_context(cafile=opt_ca_certs)
+                elif os.path.isdir(opt_ca_certs):
+                    cxt = ssl.create_default_context(capath=opt_ca_certs)
+                else:
+                    LOGGER.warning('ca_certs is specified via ssl_options but '
```
vitaly-krugl: Should raise ValueError exception instead.

pika/connection.py
```diff
@@ -963,27 +963,64 @@ def _set_url_ssl_options(self, value):
...
+                                   'is neither a valid file nor directory: "%s"',
+                                   opt_ca_certs)
+
+            if 'certfile' in opts:
+                if os.path.isfile(opts['certfile']):
+                    keyfile = opts.get('keyfile')
+                    password = opts.get('password')
+                    cxt.load_cert_chain(opts['certfile'], keyfile, password)
+                else:
+                    LOGGER.warning('certfile is specified via ssl_options but '
```
vitaly-krugl: Should raise ValueError exception instead.

pika/connection.py
```diff
@@ -963,27 +963,64 @@ def _set_url_ssl_options(self, value):
...
+                                   'is not a valid file: "%s"',
+                                   opts['certfile'])
+
+            if 'ciphers' in opts:
+                opt_ciphers = opts['ciphers']
+                if opt_ciphers is not None:
+                    cxt.set_ciphers(opt_ciphers)
+                else:
+                    LOGGER.warning('ciphers specified in ssl_options but '
```
vitaly-krugl: Should raise ValueError exception instead.

pika/connection.py
```diff
@@ -963,27 +963,64 @@ def _set_url_ssl_options(self, value):
...
+                                   'evaluates to None')
+
```
vitaly-krugl: Need to raise an exception if opts contains anything else that isn't supported. Otherwise, this change is very error-prone.

pika/connection.py
lukebakken: I log a warning because the previous code did not raise an exception in this case. I will take these comments into account for version `1.0.0`

lukebakken: If I remember correctly, you don't see an error / exception until later with the old "create a socket to evaluate ssl options" method.

pika/connection.py
lukebakken: The `ssl.SSLSocket` ctor we were using is deprecated and the old `wrap_socket`, too.

PR #1135, opened 2018-11-04
Fix examples for new API

claws:
The asynchronous consumer example does not work in the current master branch.

I installed the ``pika`` Python project after downloading the current master from Github.
``` console
(venv) $ python
Python 3.6.4 (v3.6.4:d48ecebad5, Dec 18 2017, 21:07:28) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pika
>>> pika.__version__
'1.0.0b1'
```

After starting an RabbitMQ server I then tried to run the examples ``asynchronous_consumer_example.py`` and ``asynchronous_publisher_example.py`` where I encountered errors that stopped the consumer from running. A snippet of the error is shown below.

``` python
Traceback (most recent call last):
  File "/path/to/my/venv/lib/python3.6/site-packages/pika/adapters/utils/io_services_utils.py", line 1037, in _on_socket_readable
    self._consume()
  File "/path/to/my/venv/lib/python3.6/site-packages/pika/adapters/utils/io_services_utils.py", line 796, in _consume
    self._protocol.data_received(data)
  File "/path/to/my/venv/lib/python3.6/site-packages/pika/adapters/base_connection.py", line 492, in _proto_data_received
    self._on_data_available(data)
  File "/path/to/my/venv/lib/python3.6/site-packages/pika/connection.py", line 2075, in _on_data_available
    self._process_frame(frame_value)
  File "/path/to/my/venv/lib/python3.6/site-packages/pika/connection.py", line 2224, in _process_frame
    if self._process_callbacks(frame_value):
  File "/path/to/my/venv/lib/python3.6/site-packages/pika/connection.py", line 2205, in _process_callbacks
    frame_value)  # Args
  File "/path/to/my/venv/lib/python3.6/site-packages/pika/callback.py", line 60, in wrapper
    return function(*tuple(args), **kwargs)
  File "/path/to/my/venv/lib/python3.6/site-packages/pika/callback.py", line 92, in wrapper
    return function(*args, **kwargs)
  File "/path/to/my/venv/lib/python3.6/site-packages/pika/callback.py", line 236, in process
    callback(*args, **keywords)
  File "asynchronous_consumer_example.py", line 229, in on_bindok
    self.start_consuming()
  File "asynchronous_consumer_example.py", line 244, in start_consuming
    callback=self.on_message)
TypeError: basic_consume() missing 1 required positional argument: 'on_message_callback'
```

[...]

PR #1139, opened 2018-11-18
Remove contributing instruction requesting use of yapf tool

claws:
I proposed some changes in #1135 and followed the contributing guide which requested that code be formatted using the yapf formatter. This introduced many file modifications that were not related to the actual change. The files in question likely had not been formatted with the yapf tool and there must not be style compliance checks in the CI chain.

These non-functional (i.e. style) changes add lots of noise to the proposed change making it harder for reviewers to consider the changes.

This merge request updates the contributing instructions to remove the request asking contributors to format their proposed change using yapf. This merge request is being raised as requested by this review [comment](https://github.com/pika/pika/pull/1135#issuecomment-436041657).

#### Recommendation
I personally think that a better long term solution would be to adopt a code formatter such as [black](https://black.readthedocs.io/en/stable/) and integrate it into the CI chain. The process would likely involve raising a separate merge request that:
- Adds ``black`` as a dev dependency
- Removes ``yapf`` as a dev dependency
- Applies the black formatter over the whole code base. It would use and accept the default 
  output (irrespective of whether you agree with all of its style, it is not worth wasting time 
  arguing about trivial style preferences). 
- Adds a stage in the CI chain, prior to executing tests, that runs the formatting check (e.g. 
  ``black --check``) to ensure code style is compliant. 

With such a process in place you don't need to worry about this aspect ever again.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,8 +59,3 @@ If you would like to run TLS/SSL tests, use the following procedure:
     PIKA_TEST_TLS=true nosetests
     ```
 
-
-## Code Formatting
-
-Please format your code using [yapf](http://pypi.python.org/pypi/yapf)
-with ``google`` style prior to issuing your pull request.


gmr:
I prefer pep8 style from yapf to black. I don't see how black's formatting preferences make substantive changes to the process from yapf's. If I were still the only maintainer, I would have closed this.

lukebakken:
Thanks for your input @gmr 

@claws we will be sticking with `yapf` / `pep8` and I will re-visit this PR on Monday. Several contributors have used `yapf` like you did (and as part of the instructions) only to re-format parts of the code not related to their contribution.

PR #1161, opened 2019-01-29
Remove warnings when invalid SSL URL params are used

### LINKED ISSUE #1156: Validate ssl_options earlier
See discussion in #1132

lukebakken:
Fixes #1156 

In short, this PR removes unnecessary file checks and warnings and instead lets the `SSLContext` instance throw exceptions if items are specified but not present.

Tested with the following script and RabbitMQ configs:

[tls_mutual_auth.config.txt](https://github.com/pika/pika/files/2805756/tls_mutual_auth.config.txt)
[tls_mutual_auth_ex.py.txt](https://github.com/pika/pika/files/2805757/tls_mutual_auth_ex.py.txt)

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -964,6 +964,13 @@ class URLParameters(Parameters):
                     'Specified ssl_options=None URL arg is inconsistent with '
                     'the specified https URL scheme.')
         else:
+            # Older versions of Pika would take the opts dict and pass it
+            # directly as kwargs to the deprecated ssl.wrap_socket method.
+            # Here, we take the valid options and translate them into args
+            # for various SSLContext methods.
+            #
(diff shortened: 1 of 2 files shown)


pika/connection.py
```diff
@@ -981,36 +988,22 @@ def _set_url_ssl_options(self, value):
             #     ca_certs=None,
             #     do_handshake_on_connect=True, # Not URL-supported
             #     suppress_ragged_eofs=True,    # Not URL-supported
-            #     ciphers=None
-            cxt = None
+            #     ciphers=None)
+            #
+            cxt = ssl.SSLContext()
```
vitaly-krugl: ctx is the standard abbreviation for context, not cxt as far as I know :)

PR #1162, opened 2019-01-29
Rename all_channels to global_qos

### LINKED ISSUE #1145: QoS all_channels?
I think `all_channels` parameter of `Channel.basic_qos` method is confusing because according to https://www.rabbitmq.com/consumer-prefetch.html it means applying QoS across all consumers on the channel (not "applying to all channels").

Should it not be called `global_` according to https://github.com/pika/pika/blob/e76e2eaeb1ea8b903934dc45bb843ff0fd315fc7/pika/spec.py#L1309-L1311

lukebakken:
Fixes #1145 

`global` is a keyword, FYI


decaz:
@lukebakken for consistency it could be called `global_` according to https://github.com/pika/pika/blob/e76e2eaeb1ea8b903934dc45bb843ff0fd315fc7/pika/spec.py#L1309-L1311

lukebakken:
I'd rather not have a parameter ending in underscore.

PR #1165, opened 2019-01-31
Renamed Qos.global_ to Qos.global_qos

decaz:
Renamed Qos.global_ to Qos.global_qos in addition to #1162.
There are also some cosmetic changes were added.


decaz:
@lukebakken done - #1166.

PR #1179, opened 2019-02-05
WIP: Remove custom RMQ installation

lukebakken:


tests/acceptance/io_services_tests.py
```diff
@@ -735,19 +735,6 @@ def start(self):
         self.check_cancel_connect(socket.AF_INET)
 
 
-class TestConnectSocketIPv6Success(SocketConnectorTestBase,
```
vitaly-krugl: Why remove IPV6 tests?

lukebakken: To remove the need for `sudo` which would make builds a lot faster. This PR is more of an experiment than anything, because we can't use custom RabbitMQ config when the package is used without also using `sudo`

PR #1180, opened 2019-02-05
Run pylint and yapf on examples

### LINKED ISSUE #617: Fix errors found by pylint in twisted_service.py example

lukebakken:
Docs will no longer have code pasted into them, instead links will be used

Fixes #617


jeremycline:
Hi @lukebakken,

I took a look at the warnings for examples/twisted_service.py as requested. I should say I never use pylint myself since I disagree with it so regularly, so... take my feedback with a grain of salt, I guess:

```
************* Module twisted_service
E1101,  60:19 - Module 'twisted.application.internet' has no 'SSLClient' member (no-member)
E1101,  66:19 - Module 'twisted.application.internet' has no 'TCPClient' member (no-member)
W0201, 200: 8 - Attribute 'factory' defined outside __init__ (attribute-defined-outside-init)
W0221, 204: 4 - Parameters differ from overridden 'clientConnectionLost' method (arguments-differ)
W0201,  69: 8 - Attribute 'service' defined outside __init__ (attribute-defined-outside-init)
R0201, 239: 4 - Method could be a function (no-self-use)
E1111, 254: 8 - Assigning result of a function call, where the function has no return (assignment-from-no-return)
E1121, 254:23 - Too many positional arguments for method call (too-many-function-args)
W0201, 255: 8 - Attribute 'amqp' defined outside __init__ (attribute-defined-outside-init)
```
The first two are wrong and can be safely ignored, but to be fair to pylint Twisted is [dynamically generating the classes](https://github.com/twisted/twisted/blob/132aa9f0f2c8f25a1386d518ddb92bd0e1e3b045/src/twisted/application/internet.py#L200) and that's... unusual.

```
W0201, 200: 8 - Attribute 'factory' defined outside __init__ (attribute-defined-outside-init)
W0201,  69: 8 - Attribute 'service' defined outside __init__ (attribute-defined-outside-init)
```
These two can be fixed by defining "factory" in the ``PikaProtocol.__init__`` method and "service" in the ``PikaFactory.__init__`` method respectively. Since PikaProtocol doesn't have its own ``__init__`` at the moment, it's tough to say whether fixing that warning is worth the additional boilerplate code. This is much more a style issue where everyone has their own opinion. 

```
W0221, 204: 4 - Parameters differ from overridden 'clientConnectionLost' method (arguments-differ)
```
This one, in my opinion, is silly. It's upset that the argument names differ - the parent class uses "unused_reason" because it doesn't use the reason, but the overridden method does and calls the argument "reason".

```
R0201, 239: 4 - Method could be a function (no-self-use)
```
Yeah, the task method could be a function outside the class. Seems fine to me as it is since it's just an example, but :man_shrugging:.

```
E1111, 254: 8 - Assigning result of a function call, where the function has no return (assignment-from-no-return)
E1121, 254:23 - Too many positional arguments for method call (too-many-function-args)
```
I think this is Twisted being too clever for pylint again, as the [function in question](https://twistedmatrix.com/documents/current/api/twisted.application.service.IServiceCollection.html) returns a service and takes one argument.

```
W0201, 255: 8 - Attribute 'amqp' defined outside __init__ (attribute-defined-outside-init)
```
This could be fixed by adding an ``__init__`` to the ``TestService`` and setting amqp to None.

lukebakken:
@jeremycline THANK YOU. The first two were driving me nuts and I couldn't figure it out via the docs, either.

jeremycline:
No problem, happy to help!

michaelklishin:
I disagree with various lint tools myself on a regular basis but they are still worth using in almost every case.

Thank you for the detailed explanation, @jeremycline!

lukebakken:
Merging since these are just cosmetic changes to example code.

PR #1182, opened 2019-02-06
Run yapf and pylint on pika/adapters code

lukebakken:


lukebakken:
Merging since all tests passed and these should be non-functional changes

PR #1188, opened 2019-02-07
Improve docstrings consistency

maggyero:
This PR will:

- use `:param <type> <name>: <description>` consistently instead of both `:param <type> <name>: <description>` and `:param <name>: <description>\n:type <name>: <type>`;
- use `:returns:` consistently instead of both `:return:` and `:returns:`;
- add missing `:rtype: <type>`.


lukebakken:
Thank you!

PR #1209, opened 2019-05-01
Implement comparable spec classes

ingwinlu:
## Proposed Changes

Implement equality function on amqp_objects as described in https://groups.google.com/forum/#!topic/pika-python/QgzuYN4Z6KY
.

## Types of Changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

ask on the

## Further Comments

If this is a relatively large or complex change, kick off the discussion
by explaining why you chose the solution you did and what alternatives
you considered, etc.

### DIFF
--- a/pika/amqp_object.py
+++ b/pika/amqp_object.py
@@ -46,8 +46,8 @@ class Method(AMQPObject):
         :param bytes body: The message body
 
         """
-        self._properties = properties # pylint: disable=W0201
-        self._body = body # pylint: disable=W0201
+        self._properties = properties  # pylint: disable=W0201
+        self._body = body  # pylint: disable=W0201
 
     def get_properties(self):
         """Return the properties if they are set.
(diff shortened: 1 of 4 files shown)


pika/amqp_object.py
```diff
@@ -21,6 +21,12 @@ def __repr__(self):
             return "<%s>" % self.NAME
         return "<%s(%s)>" % (self.NAME, sorted(items))
 
+    def __eq__(self, other):
```
lukebakken: How about adding this here:

```
if self is other:
    return True
```

ingwinlu: if self is other then the comparison of the dicts point to the same dict object. hence there should not be much of a difference.

lukebakken: `self.__dict__ == other.__dict__` is smart enough to do a reference comparison first, then? That makes sense.

PR #1233, opened 2019-07-15
Fix add_on_return_callback callback parameters

### LINKED ISSUE #1232: twisted channel _on_puback_message_returned has wrong number of args
_on_puback_message_returned in twisted channel has only 1 parameter, bug it's called with 4, is it a bug?

lukebakken:
Fixes #1232

@abompard hello! If you have time to review this I would appreciate it.


michaelklishin:
This signature matches what the classic blocking connection adapter uses 👍

lukebakken:
@michaelklishin thanks!

PR #1254, opened 2019-12-10
Fix RecursionError on Python 3.8

lukebakken:
Reported here:

https://groups.google.com/d/topic/pika-python/JL014cAQ0ic/discussion

Code to test:

```
import pika

connection = pika.BlockingConnection()
method_frame, header_frame, body = channel.basic_get('test')
if method_frame:
    print(method_frame, header_frame, body)
    print(repr(header_frame.headers["timestamp_in_ms"]))
    channel.basic_ack(method_frame.delivery_tag)
else:
    print('No message returned')
```

Steps:

* Create queue `test` and publish a message to it with a `timestamp_in_ms` header using a number value like `1234567890`
* Run the above script using Python `3.8.0`. Without these changes, it'll crash.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ addons:
 
 env:
   global:
-    - RABBITMQ_VERSION=3.8.0
+    - RABBITMQ_VERSION=3.8.2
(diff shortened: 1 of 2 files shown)


lukebakken:
Thanks @dcorbacho !!

PR #1263, opened 2020-02-19
docs: remove incorrect changelog entry

brianmaissy:
Small documentation fix: removed an incorrect line from the 1.0.0 change list

As discussed on the mailing list: https://groups.google.com/d/topic/pika-python/G24Q0HAs1eo/discussion


lukebakken:
@brianmaissy thanks for using Pika and this PR. Please check my follow-up commit. I think we still need to include PR 901 in the `0.12` and `1.0.0` sections because the behavior is quite different -

https://github.com/pika/pika/pull/1263/commits/71ea92cc5cf4cb420b886aad8ea602c45376c3ab

brianmaissy:
Sounds good, thanks!

PR #1270, opened 2020-05-06
use explicit instead of implicit arguments for connection in tls example

torvitas:
## Proposed Changes

since this is a tutorial I think it would be more helpful if done explicit

## Types of Changes

- [x] Documentation (correction or otherwise)

## Checklist

- [x] its still just a documentation change

## Further Comments
I like trains

### DIFF
--- a/docs/examples/tls_mutual_authentication.rst
+++ b/docs/examples/tls_mutual_authentication.rst
@@ -17,7 +17,8 @@ tls_example.py::
     context.load_cert_chain("PIKA_DIR/testdata/certs/client_certificate.pem",
                             "PIKA_DIR/testdata/certs/client_key.pem")
     ssl_options = pika.SSLOptions(context, "localhost")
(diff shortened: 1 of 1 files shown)


docs/examples/tls_mutual_authentication.rst
```diff
@@ -17,7 +17,8 @@ tls_example.py::
     context.load_cert_chain("PIKA_DIR/testdata/certs/client_certificate.pem",
```
lukebakken, line 17: Why not change all arguments to named arguments?

torvitas: All the arguments on the changed call have names.

lukebakken: I mean in all method calls in the example, for instance:

* `load_cert_chain`
* `SSLOptions`

PR #1277, opened 2020-06-15
Fixed minor function call overhead

AndreCimander:
## Proposed Changes

During some profiling I noticed a duplicated lambda call in the [call graph](https://04936760119988894920.googlegroups.com/attach/2a7cd6998e4b9/Screenshot_20200606_211952_normal_while.png?part=0.1&view=1), but I am not sure if this proposed change of a **while**-loop to a **do...while**-loop is side effect free as I don't have a full grasp of the internal polling flow, so don't hesitate to close this PR :)

## Types of Changes

## Checklist

ask on the

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -509,9 +509,11 @@ class BlockingConnection(object):
                     (not waiters or any(ready() for ready in waiters))))
 
         # Process I/O until our completion condition is satisfied
-        while not is_done():
+        while True:
             self._impl.ioloop.poll()
             self._impl.ioloop.process_timeouts()
+            if is_done():
+                break
 
         if self._closed_result.ready:
             try:


AndreCimander:
Looking at the tests I would argue it has side effects :) 

Is there a special reason why the conditionals of `is_done` are wrapped into a function? Since it isn't reused I would argue moving those into the while loop will eliminate a few unnecessary function calls.

AndreCimander:
I'll close this one as it starts to gather dust :)

lukebakken:
@AndreCimander please be patient. We appreciate the time you've taken in submitting this pull request. The people who maintain Pika (mostly me at this point) do so in their free time. My main job is working on RabbitMQ and I spend time here when I can.

AndreCimander:
@lukebakken no worries, this wasn't meant to put any pressure on anyone, sorry if I you felt that way :)

It's just a habit of mine to keep the open merge request count as low as possible, otherwise they seemingly start to multiply and projects become... unwieldy.

michaelklishin:
The comment in `_flush_output` suggests that there is a reason:

``` py
# Conditions for terminating the processing loop:
        #   connection closed
        #         OR
        #   empty outbound buffer and no waiters
        #         OR
        #   empty outbound buffer and any waiter is ready
        is_done = (lambda:
                   self._closed_result.ready or
                   ((not self._impl._transport or
                     self._impl._get_write_buffer_size() == 0) and
                    (not waiters or any(ready() for ready in waiters))))
```

The condition is computed based on a number of factors and I don't think we can simply replace it with a `True` without behavior changes.

Thank you for taking the time to submit this 👍

PR #1281, opened 2020-07-27
Changed the information that is logged.

priyabhosale:
## Proposed Changes
https://github.com/pika/pika/blob/master/pika/channel.py#L1237 ---> this logger is logging all the header info. It can expose sensitive information in the header.
Changed the information that is logged in logger statement. 

## Checklist

ask on the


michaelklishin:
The goal of this is likely to make it easy to spot what routing key was used (it's an [Unroutable Message return](https://www.rabbitmq.com/publishers.html#unroutable) after all). I have a better idea: let's make this statement a debug one.

No other client warns on a `basic.return` but many log a debug message.

PR #1287, opened 2020-10-13
Add Gevent adapter

jpgxs:
Add GeventConnection class for non-blocking connections using Gevent's native IO loop.

Also fixes imports in async adapter tests so that they can be executed directly with `python -m nose tests/acceptance/async_adapter_tests.py`

## Proposed Changes

This change makes Pika's non-blocking APIs available for Gevent users without monkey patching.

## Types of Changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist

- [X] I have read the `CONTRIBUTING.md` document
- [X] All tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works

## Further Comments

The Gevent acceptance tests are skipped on Windows as Gevent is segfaulting in the AppVeyor tests. It could be an issue with my code or with the pre-built Gevent package being pulled from PyPi - I'll try to confirm this week.


michaelklishin:
@jpgxs thank you for this non-trivial contribution. Sorry that's taken us a while to respond.

lukebakken (approved):

dabbler0606:
Hello, my python version is 2.7, but I want to use pika and gevent, is there a solution?

Looking at the documentation shows that 1.2.0 only supports python 3.4+

thanks

michaelklishin:
@dabbler0606 consider using GitHub Discussions instead of commenting on random merged PRs.
You have to find an older (potentially quite a bit older) version of Pika and use that on Python 2.7. There are no plans to re-introduced Python 2 support since it's out of support.

Issue #1285, opened 2020-10-02
Use of catch_warnings/filterwarnings can cause exceptions to be raised in other threads


asinha94:
While encoding an integer, all warning are filtered so that they will raise exceptions if they occur. [The context-manager is not thread-safe](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) and any warning raised in any other thread at the same time will be raised as an exception. My team has run into this a few times while pika is under heavy load. Offending code here https://github.com/pika/pika/blob/a3ae3e5aae3fd546e9ae73bff0175f5d61a9d39a/pika/data.py#L135-L146

Reading the #888 where it was introduced, it seems to be done to maintain backwards compatibility with Python 2.6, which uses the deprecated `__int__()` method. I'm not 100% sure on where the `DeprecationWarning` is even raised since `value` is an `int` (is it when `b'I'` is packed since it's a non-integer? If thats the case can't the `b'l'` also throw?)

Since Python 2.6 isn't supported anymore I think that can be removed and the code reverted back to 
https://github.com/pika/pika/blob/fca3c48c74c546915f829f1a211a30413e44bfbc/pika/data.py#L121-L127

asinha94:
If there are no other consequences for removing the `DeprecationWarning` I can submit a PR for this, or it could be combined with #1265

lukebakken:
Please submit a PR for this issue (with tests, if appropriate). If you have time to address #1265 in another PR that would be great.

lukebakken:
Fixed by #1290

PR #1290, opened 2020-11-18
Removed warnings filter while encoding integer

asinha94:
## Proposed Changes

From #1285, while encoding an integer, all warnings are filtered and raised as exceptions. This affects all threads, not just the current thread, which can cause warnings to be raised as exceptions in other threads, where they aren't expected.

The reason for this change in the first place was to ensure compatibility with Python 2.6, which raises a `DeprecationWarning` when an integer larger than a 32-bit signed integer is passed in. This has been changed from Python 2.7 to raise `struct.error` instead. Since Python 2.6 is no longer supported (by Pika and everyone else) this can be removed, and we can rely on catching `struct.error`.

Python 2.6.6 in Docker (WSL2)

```
>>> struct.pack('>ci', b'I', 2147483648)
__main__:1: DeprecationWarning: 'i' format requires -2147483648 <= number <= 2147483647
'I\x80\x00\x00\x00'
```

Python 3.6 on WSL2
```
>>> import struct
>>> import warnings
>>> warnings.filterwarnings('error')
>>> struct.pack('>ci', b'I', 2147483648)
  File "<input>", line 1, in <module>
    struct.pack('>ci', b'I', 2147483648)
struct.error: 'i' format requires -2147483648 <= number <= 2147483647
```
## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #1285)

## Checklist

ask on the

## Further Comments

My change is minor, and since I'm removing an error/race-condition, there isn't much I can really test. I can verify that the change is running at my workplace on multiple production systems and fixes the issues we were having with this race-condition.

### DIFF
--- a/pika/data.py
+++ b/pika/data.py
@@ -2,7 +2,6 @@
 import struct
 import decimal
 import calendar
-import warnings
 
 from datetime import datetime
 
@@ -133,16 +132,14 @@ def encode_value(pieces, value): # pylint: disable=R0911
         pieces.append(struct.pack('>cq', b'l', value))
         return 9
     elif isinstance(value, int):
-        with warnings.catch_warnings():
-            warnings.filterwarnings('error')
-            try:
-                packed = struct.pack('>ci', b'I', value)
-                pieces.append(packed)
-                return 5
-            except (struct.error, DeprecationWarning):
-                packed = struct.pack('>cq', b'l', long(value))
-                pieces.append(packed)
-                return 9
+        try:
+            packed = struct.pack('>ci', b'I', value)
+            pieces.append(packed)
+            return 5
+        except struct.error:
+            packed = struct.pack('>cq', b'l', long(value))
+            pieces.append(packed)
+            return 9
     elif isinstance(value, decimal.Decimal):
         value = value.normalize()
         if value.as_tuple().exponent < 0:


lukebakken (changes requested):
How can we indicate that Python versions prior to 2.7 will no longer work correctly?

lukebakken:
Thank you, I have one question in my first review.

asinha94:
> How can we indicate that Python versions prior to 2.7 will no longer work correctly?

Since its officially unsupported is there anything more that needs to be done? We could also alter the statement slightly [(like you mentioned in your initial PR](https://github.com/pika/pika/pull/888#issuecomment-340871718)) so that it still works on prior versions.

```
elif isinstance(value, int):
    if -2147483648 <= value <= 2147483647:
        packed = struct.pack('>ci', b'I', value)
    else:
        packed = struct.pack('>cq', b'l', long(value))
    
    pieces.append(packed)
    return len(packed)
```

michaelklishin:
@asinha94 @lukebakken I vote for simply proceeding with not doing anything about Python 2.6 since, well, it's been out of support for a while.

lukebakken:
Thanks everyone!

asinha94:
Thanks!

PR #1292, opened 2020-12-07
Add explicit check for if timeout_heap has been set to None before do…

### LINKED ISSUE #1291: Should check if self._timeout_heap is None before pushing on heap
https://github.com/pika/pika/blob/a3ae3e5aae3fd546e9ae73bff0175f5d61a9d39a/pika/adapters/select_connection.py#L258
The first argument to heaq.heappush must be a list. If the close is called on the Timer object, then self._timeout_heap will be set to None, which will cause exceptions with the heappush. Adding a check for "if self.timeout_heap is not None" at the beginning of the function should fix the issue.

prot0man:
…ing any heap actions with _timeout_heap.

Fixes #1291

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -247,6 +247,9 @@ class _Timer(object):
         :raises ValueError, TypeError
 
         """
+        if self._timeout_heap is None:
+            raise ValueError("Timeout closed before call")
+
         if delay < 0:
             raise ValueError(
                 'call_later: delay must be non-negative, but got %r' % (delay,))


michaelklishin:
@prot0man why do we need this change? Some context,  test cases and a bit of justification would help. Thank you.

prot0man:
https://github.com/pika/pika/issues/1291

prot0man:
> @prot0man why do we need this change? Some context, test cases and a bit of justification would help. Thank you.

https://github.com/pika/pika/issues/1291

prot0man:
As you see from the issue, @lukebakken told me to do a pull request, but then never really took an action or commented on my pull request.

lukebakken:
I will add a test based on [this comment](https://github.com/pika/pika/issues/1291#issuecomment-748346440).

PR #1297, opened 2021-01-14
Only test on supported Python versions

lukebakken:
Modify Travis CI to only test on supported Python versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,17 +65,11 @@ after_success:
 jobs:
   include:
     - python: pypy3
-    - python: pypy
-    - python: 2.7
-    - python: 3.5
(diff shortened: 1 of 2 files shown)


yucer:
Hello, I have some questions here:

- How do we know that the other versions are not supported ? Do we need somewhere links to the issues ? Just for reference ?

- Don't you need also to change the python versions in the [classifiers section](https://github.com/pika/pika/blob/90d04a06222d195af0dc9f1daebca5f27eac8f6a/setup.py#L39) of `setup.py` ?

phillipuniverse:
Also wondering this, does 1.2 no longer support Python < 3.7?

PR #1302, opened 2021-01-26
Log basic.return events at debug level

michaelklishin:
Supersedes #1281.

This warning has been around for about 8 years so I think it was intentional. However, it can result in sensitive information from messages logged, and generally no other client produces this warning or logs messages except at `debug` level (e.g. as part of logging all outgoing and inbound frames).

@gmr @lukebakken should we change the warning to a debug log entry?


lukebakken (approved):
It makes sense to me to follow what other client libraries do.

PR #1320, opened 2021-06-01
Fix for function path

tkchk:
AttributeError: module 'urllib' has no attribute 'urlencode'

## Proposed Changes

The documentation has incorrect example. Fixed it

## Types of Changes

Just a string in docs


michaelklishin (approved):

michaelklishin:
This seems to be a Python 2 era leftover. Thank you!

PR #1345, opened 2022-03-07
#1265 fix

### LINKED ISSUE #1265: pika can't encode unsigned long long header values that don't fit in long long
I try to send a huge number in an application property header (11171330917854890490 which is between 2^63 and 2^64) but it fails with and exception:

(using Python 3.6.7 | packaged by conda-forge | (default, Nov 21 2018, 03:09:43) )

Stacktace in context is:

```
    self._channel.basic_publish(self._exchange, self._routing_key, self._message, properties)
  File "/.../lib/python3.6/site-packages/pika/channel.py", line 428, in basic_publish
    mandatory=mandatory), (properties, body))
  File "/.../lib/python3.6/site-packages/pika/channel.py", line 1401, in _send_method
    self.connection._send_method(self.channel_number, method, content)
  File "/.../lib/python3.6/site-packages/pika/connection.py", line 2232, in _send_method
    self._send_message(channel_number, method, content)
  File "/.../lib/python3.6/site-packages/pika/connection.py", line 2253, in _send_message
    marshaled_body_frames.append(frame_header.marshal())
  File "/.../lib/python3.6/site-packages/pika/frame.py", line 103, in marshal
    pieces = self.properties.encode()
  File "/.../lib/python3.6/site-packages/pika/spec.py", line 2179, in encode
    data.encode_table(pieces, self.headers)
  File "/.../lib/python3.6/site-packages/pika/data.py", line 93, in encode_table
    tablesize += encode_value(pieces, value)
  File "/.../lib/python3.6/site-packages/pika/data.py", line 133, in encode_value
    pieces.append(struct.pack('>cq', b'l', value))
struct.error: int too large to convert
```

To reproduce, I isolated this short code in the python interpreter:

imports:
```
>>> import pika.compat
>>> import pika.data
```

verify int size
```
>>> 11171330917854890490 < 2**64
True
>>> 11171330917854890490 < 2**63
False
```

try converting a small long: it works
```
>>> a=[];pika.data.encode_value(a, pika.compat.long(10));a
9
[b'l\x00\x00\x00\x00\x00\x00\x00\n']
```

try converting a big long: it fails
```
>>> a=[];pika.data.encode_value(a, pika.compat.long(11171330917854890490));a
  File "/home/.../lib/python3.6/site-packages/pika/data.py", line 132, in encode_value
    pieces.append(struct.pack('>cq', b'l', value))
struct.error: int too large to convert
```

AFAICT (pdf page 31) the amqp 0-9-1 format specifier `b'l'` is for long-long-uint, which is correct to fit this number greater than 2^63.

But the struct.pack [format string](https://docs.python.org/3.8/library/struct.html#format-characters) is incorrect: it should be **'>cQ'** (upper-case Q).

Success with **'>cQ'** specifier
```
>>> import struct
>>> struct.pack('>cQ', b'l', 11171330917854890490)
b'l\x9b\x08\x8aH\x13\x14I\xfa'
```

Failure with **'>cq'** specifier
```
>>> struct.pack('>cq', b'l', 11171330917854890490)
struct.error: int too large to convert
```

tesla1856:
#1265 fix
Docs:
https://docs.python.org/3.8/library/struct.html#format-characters
https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf (page 31)


lukebakken (approved):
Thank you!

PR #1347, opened 2022-03-23
use channel.connection instead of passing in the connection for simplification in the threaded consumer example

jackxxu:
## Proposed Changes

i find in the `examples/basic_consumer_threaded.py` passing in the connection might be unnecessary, as the channel object holds an instance of the connection already. Removing connecton argument makes the logic a little easier to follow, at least to me.

## Types of Changes

- [x] Cosmetics (whitespace, appearance)

## Checklist

## Further Comments

this is a rather simple change in an example app.

### DIFF
--- a/examples/basic_consumer_threaded.py
+++ b/examples/basic_consumer_threaded.py
@@ -27,20 +27,20 @@ def ack_message(ch, delivery_tag):
         pass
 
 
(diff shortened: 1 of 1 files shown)


lukebakken:
Thank you

PR #1364, opened 2022-05-28
Add semver note to changelog

### LINKED ISSUE #1343: Versioning policy is undocumented (appears to be semver)
There is no documentation in this repo or on the mailing list which indicates that `pika` uses semver. There are no past issues in which this is discussed, nor are there commit messages or the like. Searching the listhost history for "semver" and "semantic versioning" returns a single hit for a thread which _suggests_ that the library is using semantic versioning.

If the maintainers approve and will confirm that semver is in use, I'd like to put in a PR to add a single sentence to the top of the changelog stating
> `pika` uses [Semantic Versioning](https://semver.org)

This confirms that pinning as in `pika>=1.1,<2` will have the expected effect, and that `pika` does not use a different policy (examples of other, non-calver, policies in python: cpython, `cryptography`).

sirosen:
resolves #1343

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,8 @@
 Version History
 ===============
 
+`pika` uses [Semantic Versioning](https://semver.org)
+
 1.2.1 2022-04-27
 ----------------
 


lukebakken:
Thanks!

PR #1386, opened 2022-08-22
Update connection module to use TLS v1.2 as fallback

### LINKED ISSUE #1385: Update fallback SSL/TLS protocol version
In the connection module, the method `URLParameters._set_url_ssl_options()` tries to set the secure protocol version to `ssl.PROTOCOL_TLS` and falls back to `ssl.PROTOCOL_TLSv1` if the first is not available (since it was introduced in 3.6 and this client supports 3.4+).

https://github.com/pika/pika/blob/c34c411874f049f2048aa5133ea20f87c95fa223/pika/connection.py#L905-L912

This has the following issues:
- The usage of `ssl.PROTOCOL_TLS` is [deprecated in 3.10](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS) in favour of ` ssl.PROTOCOL_TLS_CLIENT`
- The fallback sets an unsecure protocol

I propose the following changes:
1. Try to set [ssl.PROTOCOL_TLS_CLIENT](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_CLIENT)
2. Fall back to [ssl.PROTOCOL_TLSv1_2](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLSv1_2)

Those changes would address the usage of deprecated `ssl.PROTOCOL_TLS`, and would fall back to a secure protocol by default. The constant `ssl.PROTOCOL_TLSv1_2` was introduced in Python 3.4, therefore it is safe to assume this constant will be there for supported Python versions.

Zerpet:
## Proposed Changes

URLParameters class parsing SSL options will now default/fallback to a secure
TLS version. The constant `ssl.PROTOCOL_TLS` is deprecated starting in
3.10. Constant `ssl.PROTOCOL_TLS_CLIENT` was introduced in 3.6,
 therefore this change should be backwards compatible.

For those clients out there running 3.4-3.5, there's a constant to fallback to use
TLS v1.2.

Fixes #1385

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #NNNN)

## Checklist

ask on the

## Further Comments

N/A

### DIFF
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -903,9 +903,9 @@ class URLParameters(Parameters):
             #
             # SSLContext.load_verify_locations(cafile=None, capath=None, cadata=None)
             try:
-                opt_protocol = ssl.PROTOCOL_TLS
+                opt_protocol = ssl.PROTOCOL_TLS_CLIENT
             except AttributeError:
-                opt_protocol = ssl.PROTOCOL_TLSv1
+                opt_protocol = ssl.PROTOCOL_TLSv1_2
             if 'protocol' in opts:
                 opt_protocol = opts['protocol']
 
--- a/tests/unit/connection_parameters_tests.py
+++ b/tests/unit/connection_parameters_tests.py
@@ -12,18 +12,16 @@ import pika
 from pika.compat import urlencode, url_quote, dict_iteritems
 from pika import channel, connection, credentials, spec
 
-
 # disable missing-docstring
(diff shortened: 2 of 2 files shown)


lukebakken (approved):

lukebakken:
Thank you @Zerpet

Issue #1409, opened 2022-12-27
ExchangeType should be from str type


liortct:
I am trying to use `ExhangeType` but PEP8 give me a warning.
PEP8 give me the following warning `Expected type 'str', got 'ExchangeType' instead`. 
It's because `ExhangeType` enum is inheritance only from enum and not from str as needed.

Versions:
pika=1.3.1
Python=3.8.10

Example code:
```
import pika
from pika.exchange_type import ExchangeType

connection = pika.BlockingConnection(pika.ConnectionParameters('localhost'))
channel = connection.channel()

channel.exchange_declare(exchange='logs', exchange_type=ExchangeType.topic)
```

lukebakken:
How did you get the PEP8 warning? Did you run a program like pylint?

liortct:
I got the warning simply via the PyCharm.
I will check about pylint and mypy.

lukebakken:
It is just a warning and can be ignored. If you have a fix a pull request would be appreciated.

My guess is that `ExchangeType` should be a subclass of [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum)

lukebakken:
Fixed by #1410

PR #1410, opened 2022-12-27
Fix exchange type enum type

liortct:
Addressing  #1409 .
Fix `ExchangeType` enum type to str.

### DIFF
--- a/pika/exchange_type.py
+++ b/pika/exchange_type.py
@@ -1,7 +1,7 @@
 from enum import Enum
 
 
-class ExchangeType(Enum) :
+class ExchangeType(str, Enum):
     direct = 'direct'
     fanout = 'fanout'
     headers = 'headers'
--- a/pika/exchange_type.py
+++ b/pika/exchange_type.py
@@ -1,8 +1,18 @@
-from enum import Enum
+try:
+    from enum import StrEnum  # available from Python 3.11
 
 
-class ExchangeType(str, Enum):
-    direct = 'direct'
-    fanout = 'fanout'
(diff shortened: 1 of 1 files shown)


lukebakken (changes requested):
Please see this comment, thanks - https://github.com/pika/pika/issues/1409#issuecomment-1366156190

liortct:
> Please see this comment, thanks - [#1409 (comment)](https://github.com/pika/pika/issues/1409#issuecomment-1366156190)

You are right but `StrEnum` is new in Python 3.11.
I want this will work not only for Python 3.11 so I don't use `StrEnum` and simply did like [here](https://docs.python.org/3.10/library/enum.html#others) and inheritance from `str`.

lukebakken:
Argh that's annoying.

Could you try to do an import, catch the `ImportError`, and fix things up? Otherwise at some point I'm going to forget that `StrEnum` is available and never update the code.

I'd rather do that than check the python version.

Thanks!

liortct:
> Argh that's annoying.
> 
> Could you try to do an import, catch the `ImportError`, and fix things up? Otherwise at some point I'm going to forget that `StrEnum` is available and never update the code.
> 
> I'd rather do that than check the python version.
> 
> Thanks!

Fixed :)

lukebakken (approved):
Thanks!

liortct:
> Thanks!

Can you merge this PR?

lukebakken:
> Can you merge this PR?

Yes, I was waiting for the checks to finish.

PR #1430, opened 2023-05-16
Authorization with OAuth2

binarin:
This snippet shows a few possible approaches for regularly updating secrets for SelectConnection's using OAuth2. It's not a finished work, but a point to start a discussion about this feature, and how/whether it'd be possible to support it properly in API, instead of every user hand-rolling their own implementation.  (For reference, Java client already has it - https://www.rabbitmq.com/api-guide.html#oauth2-support).

Turns out implementing OAuth2-flow is not completely trivial. While fetching a new secret is easy (it's implemented here using requests_oauthlib , but to reduce dependencies this can be done in pure http), doing this regularly is a bit more complex task. I've tried the following 2 approaches:

- Doing a blocking http call in connection thread (see BlockingConnectionSecretRefresher). This is simple, but an OAuth2 http call takes ~0.3 second in my tests, which can be too much blocking.
- Running a separate thread (see ThreadedConnectionSecretRefresher). In this case it has no negative effect on processing speed, but introduces a lot of moving parts.

There is also a question of what to do with all other connection adapters. It looks like methods similar `call_later`/`add_callback_threadsafe` are universally available, but e.g. for aio/twisted adapters it should be possible to do the same using non-blocking http calls.

### DIFF
--- a/examples/oauth2.py
+++ b/examples/oauth2.py
@@ -1,0 +1,202 @@
+from pprint import pp
+import logging
+import traceback
+import pika
+from pika.credentials import PlainCredentials
+import json
+import os
+import os.path
+import threading
+import time
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
+import abc
+
+logger = logging.getLogger(__package__)
+
+def mk_token_fetcher(token_url, client_id, client_secret, params):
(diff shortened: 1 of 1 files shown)


binarin:
One way I envision is to sublclass `PlainCredentials` as `OAuth2Credentials`, and then do a special-case in every adapter, where a new thread for updates will be started automatically.

PR #1467, opened 2024-05-24
Use sys.platform instead of platform.system()

cvijdea-bd:
`platform.system()` is notoriously slow on Windows: https://github.com/python/cpython/issues/95531

This change shaves >100ms off the runtime for `python -c "import pika"` on my machine.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -15,20 +15,20 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


lukebakken (approved):
@cvijdea-bd if you have a second, please double-check the changes I made. Thank you for your contribution. PS it's nice to run into someone else who uses Windows 👍

cvijdea-bd:
Seems good!

PR #1484, opened 2024-08-19
Convert utcnow() and utcfromtimestamp()

mh21:
Both methods are deprecated, and now result in warnings like

    DeprecationWarning: datetime.datetime.utcfromtimestamp() is
    deprecated and scheduled for removal in a future version. Use
    timezone-aware objects to represent datetimes in UTC:
    datetime.datetime.fromtimestamp(timestamp, datetime.UTC).

and

    DeprecationWarning: datetime.datetime.utcnow() is deprecated and
    scheduled for removal in a future version. Use timezone-aware
    objects to represent datetimes in UTC:
    datetime.datetime.now(datetime.UTC).

An example can be seen in https://gitlab.com/cki-project/cki-lib/-/jobs/7614326286#L166.

While most use of these methods is in test code, there is one instance of datetime.utcfromtimestamp() used during deserialization in data.py that might affect user code.

## Proposed Changes

- `datetime.utcnow() -> datetime.now(timezone.utc)`
- `datetime.utcfromtimestamp(...) -> datetime.fromtimestamp(..., timezone.utc)`

## Types of Changes

- [x] Bugfix (non-breaking change which resolves DeprecationWarnings)

## Checklist

### DIFF
--- a/pika/data.py
+++ b/pika/data.py
@@ -3,7 +3,7 @@ import struct
 import decimal
 import calendar
 
-from datetime import datetime
+from datetime import datetime, timezone
 
 from pika import exceptions
 from pika.compat import long, as_bytes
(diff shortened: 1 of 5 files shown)


lukebakken (approved):
Thank you

PR #1485, opened 2024-08-21
specify your own consumer_tag when using consume() method of BlockingChannel

ivan-eguidazu:
When using consume() method of BlockingChannel class, was not possible to define your own consumer_tag
This optional parameter has been added to consume() method


lukebakken:
Thank you!

PR #1505, opened 2024-12-23
Add time_limit optional argument to be able to inject in start_consuming function.

AzgTamir:
Add time_limit optional argument to be able to inject time_limit in BlockingConnection.start_consuming function

This pull request is necessary because it enhances the "start_consuming" method by adding an optional time_limit parameter, making it more flexible and aligning its functionality with the documented description. The original implementation lacked the ability to specify a time limit, limiting its usability in scenarios requiring controlled operation duration. By introducing this parameter and passing it to "_process_data_events", the method now supports time-bound consumption, which broadens its applicability while maintaining backward compatibility, as the default behavior remains unchanged. This change not only ensures consistency between the code and its documentation but also improves clarity and usability, empowering developers to integrate the method into applications with stricter timing requirements without any disruption to existing workflows.

## Proposed Changes
A simple change to "start_consuming" function to be able to inject time_limit.

## Types of Changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist

- [X] I have read the `CONTRIBUTING.md` document
- [X] All tests pass locally with my changes
- [<No additional tests are required>] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)

## Further Comments
No further comment.

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -1853,7 +1853,7 @@ class BlockingChannel:
 
         return unprocessed_messages
 
-    def start_consuming(self):
+    def start_consuming(self, time_limit=None):
         """Processes I/O events and dispatches timers and `basic_consume`
         callbacks until all consumers are cancelled.
 
@@ -1877,7 +1877,7 @@ class BlockingChannel:
         # Process events as long as consumers exist on this channel
         while self._consumer_infos:
             # This will raise ChannelClosed if channel is closed by broker
-            self._process_data_events(time_limit=None)
+            self._process_data_events(time_limit=time_limit)
 
     def stop_consuming(self, consumer_tag=None):
         """ Cancels all consumers, signalling the `start_consuming` loop to
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -1861,6 +1861,8 @@ class BlockingChannel:
         pika callback, because dispatching `basic_consume` callbacks from this
         context would constitute recursion.
 
+        :param int time_limit: time limit of the consume operation
+
         :raises pika.exceptions.ReentrancyError: if called from the scope of a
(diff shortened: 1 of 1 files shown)


lukebakken:
Thanks for using Pika.

If people need a time limit, the usual solution is to use the generator: [docs](https://pika.readthedocs.io/en/stable/examples/blocking_consumer_generator.html). This method allows setting a time limit.

AzgTamir:
Hi,
The BlockingChannel.consume and BlockingChannel.start_consuming are
entirely different functions.

Please reconsider my approach.

Thanks,
Tamir.

lukebakken:
> The BlockingChannel.consume and BlockingChannel.start_consuming are entirely different functions

As I said, you can achieve what you want using the existing `BlockingChannel.consume` function.

Other reasons this PR was closed:
* No tests.
* There has not yet been a request for this feature in the long history of this project. Presumably because nobody needs it.

If you can provide tests for this feature, as well as a _concrete_ example, using code I can run, that demonstrates how the existing `BlockingChannel.consume` is insufficient for your needs, I will reconsider this change.

shinji-s:
The BlockingChannel.consume and BlockingChannel.start_consuming are different styles for achieving the same goal, Providing 'inactivity_timeout' parameter only for the former violates symmetry. IMO, provided that proper unit tests are given, the parameter  had better be allowed for both. (In the past I resorted to setting a timeout handler which cancels channels to work around the lack of the parameter in start_consuming().)

lukebakken:
The thing is, if a timeout is added to `start_consuming`, people will use it, then do other work on the same thread that Pika's I/O loop is using, then will complain about Pika not working as intended.

One nice thing about `start_consuming`'s behavior is that it _forces_ users to think about threading.

For better or worse, Pika is not thread-safe.

PR #1544, opened 2026-02-21
Fix asyncio.get_event_loop() deprecation in _AsyncioIOServicesAdapter

### LINKED ISSUE #1479: There is no current event loop in thread
https://github.com/pika/pika/blob/96a92379346285a8f53dd0b3f76cb26f45baf350/examples/asyncio_consumer_example.py#L64

May I suggest an addition here. If connect is run on a thread, this will result in a RuntimeError exception with the message "There is no current event loop in thread".

The fix in this case is to call `asyncio.new_event_loop` instead of`asyncio.get_event_loop`. Here is a, possibly inelegant, solution:

```python
        try:
            connection = AsyncioConnection(
                parameters=pika.URLParameters(self._url),
                on_open_callback=self.on_connection_open,
                on_open_error_callback=self.on_connection_open_error,
                on_close_callback=self.on_connection_closed)
        except RuntimeError as e:
            if 'loop' in str(e):

                # Assume that this exception is related to no current event loop on this thread, ie:
                # "There is no current event loop in thread"...
                # This happens when the connection is on a thread

                connection = AsyncioConnection(
                    parameters=pika.URLParameters(self._url),
                    on_open_callback=self.on_connection_open,
                    on_open_error_callback=self.on_connection_open_error,
                    on_close_callback=self.on_connection_closed,
                    custom_ioloop=asyncio.new_event_loop())
            else:
                raise e
```

geomarceau:
## Summary

`asyncio.get_event_loop()` was deprecated in Python 3.10 when no running
loop exists, and raises `RuntimeError` in Python 3.14. This caused
`AsyncioConnection` to fail when instantiated from a non-async thread or
from a plain synchronous context where no event loop has been started.

## Changes

- Replace `asyncio.get_event_loop()` in `_AsyncioIOServicesAdapter.__init__`
  with `asyncio.get_running_loop()`, falling back to `asyncio.new_event_loop()`
  when no loop is running. This correctly reuses an existing running loop (e.g.
  when called from within async code) and safely creates a new one otherwise.
- Update the docstring in `AsyncioConnection.__init__` to reflect the new
  behaviour.
- Remove the `try/except get_event_loop()` workaround in
  `io_services_test_stubs_test.py` that was masking the same issue in the
  test suite.
- Add `tests/unit/asyncio_connection_tests.py` covering all three code paths:
  explicit loop, running loop reuse, and new loop creation when none is running
  (both from a thread and from a plain synchronous context).

## Related

Closes #1479

## Types of Changes

What types of changes does your code introduce to this project? Put an x in the boxes that apply

- [x] Bugfix (non-breaking change which fixes issue #1479)

## Checklist

Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask on the **pika-python** mailing list. We're here to help! This is simply a reminder of what we are going to look for before merging your code.

- [x] I have read the CONTRIBUTING.md document


lukebakken (dismissed):
Thank you! @suchitd is new to this project, so I'm letting him give the final 👍 on PRs

lukebakken (approved):

PR #1545, opened 2026-02-26
fix: replace 3 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.

### DIFF
--- a/pika/callback.py
+++ b/pika/callback.py
@@ -231,7 +231,7 @@ class CallbackManager:
             LOGGER.debug('Calling %s for "%s:%s"', callback, prefix, key)
             try:
                 callback(*args, **keywords)
-            except:
+            except Exception:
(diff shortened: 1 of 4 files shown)


lukebakken (commented):
Hello, thanks for taking the time to contribute!

Would you mind explaining how you found these? Also, is there any sort of linter we could run to catch practices like this?

alonfaraj:
@lukebakken `flake8` for example, although I think those days [`ruff`](https://docs.astral.sh/ruff/) is probably the most popular and super fast.
It can catch (and auto-fix safe issues) like unused import, bad formatting etc.

BTW @haosenwang1018 - the linter caugth another bare except if you want to fix it as well
https://github.com/pika/pika/blob/f950404a752fe663db53fe9e9fd13de5984b504d/tests/wrappers/threaded_test_wrapper.py#L146

lukebakken (approved):

Issue #1547, opened 2026-04-27
GitHub actions workflows and test code need updates for RabbitMQ 4.3


lukebakken:
https://github.com/pika/pika/actions/runs/24999124998/job/73204056511

RabbitMQ 4.3 does not support certain queue arguments / properties that are used by the test suites, so updates need to happen.

lukebakken:
Commit https://github.com/pika/pika/commit/f950404a752fe663db53fe9e9fd13de5984b504d works around this issue for now.

lukebakken:
Fixed by #1548

PR #1548, opened 2026-04-28
Implement tests compatibility patch for RabbitMQ 4.3 

alonfaraj:
## Proposed Changes

RabbitMQ 4.3 deprecated `transient_nonexcl_queues` so queues declared with `durable=False` and `exclusive=False` are now rejected.

Fixed by adding `exclusive=True` to every relevant `queue_declare` in the tests. 
Exclusive queues are scoped to the declaring connection and auto-delete on disconnect.
Removed stale cleanup lambdas that opened new connections to delete queues since they're unnecessary with exclusive queues and would fail with `RESOURCE_LOCKED`.

One test (`TestCloseConnectionFromBasicConsumeCallback`) uses `durable=True` instead because it reconnects after closing and verifies messages survive.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #1547)

## Checklist

## Further Comments
Also changed `.github/workflows/main.yaml` push branch filter from `"*"` to `"**"` so CI triggers on branches with `/` in the name (e.g. `fix/...`, `feature/...`).


lukebakken (changes requested):
I appreciate your efforts here, however, I would rather see the tests themselves fixed. If this means touching every test that declares a queue, so be it. Honestly there should be a base class or helper class to consolidate common operations like queue declarations.

I would like @suchitd to chime in on this as well.

alonfaraj:
@lukebakken Sounds like a better approach; done

tests/acceptance/blocking_adapter_test.py
```diff
@@ -373,8 +373,7 @@ def test(self):
         body1 = 'a' * 1024
 
         # Declare a new queue
-        ch.queue_declare(q_name, auto_delete=True)
-        self.addCleanup(lambda: self._connect().channel().queue_delete(q_name))
+        ch.queue_declare(q_name, auto_delete=True, exclusive=True)
```
suchitd: `auto_delete` is redundant when we use `exclusive=True`.

alonfaraj: True.
removed all `auto_delete[True/False]` when `exclusive=True`.

lukebakken (dismissed):

lukebakken:
Thank you @alonfaraj

suchitd (commented):
The acceptance tests changes LGTM!

suchitd (approved):

PR #1441, opened 2023-08-03
Shrink the compatibility module and remove Python <3.7 support

alfawal:
## Proposed Changes

Reducing the usage of the `pika.compat` usage. It's really depressing for a first-time contributor to see Python 2 being used nowadays.

I believe/hope that these changes will encourage more people to contribute!

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation (correction or otherwise)
- [x] Cosmetics (whitespace, appearance)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after
creating the PR. If you're unsure about any of them, don't hesitate to
ask on the
[`pika-python`](https://groups.google.com/forum/#!forum/pika-python)
mailing list. We're here to help! This is simply a reminder of what we
are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] All tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further Comments

On the consistency side. I believe that this repository should:
- Add pre-commit config and inform the contributors to use it. The pre-commit config:
    - Use `black` as its formatter.
    - Use `isort` as its imports reordering handler.
    - (Optional) Use items from `ruff`.
    - (Optional) Use `pydocstyle` for docstrings with the Google style.
    - etc.
- Do an initial run with the pre-commit on all files.

PR #1551, opened 2026-04-29
Add `ruff` lint check to CI (GH-1371)

### LINKED ISSUE #1371: pylint
https://pylint.pycqa.org/en/latest/user_guide/usage/run.html

Add a GHA that runs pylint, and fix all of the new warnings

lukebakken:
## Proposed Changes

Add [ruff](https://docs.astral.sh/ruff/) as a CI lint check and fix all findings it reports with the default rule set.

Closes #1371.

Ruff has become the standard fast Python linter since #1371 was filed, covering the practical value originally sought from pylint (unused imports, lambda style, comparison bugs, import ordering) in a single fast binary that integrates cleanly with GitHub Actions.

What this PR does, per commit:

1. 69ccb52 adds a `[tool.ruff]` section to `pyproject.toml` with `target-version = "py37"` (matching `requires-python`) and `line-length = 100` (matching the old `pylintrc`). Generated `pika/spec.py` is excluded from the start.
2. 611accf applies the 9 auto-fixable findings from `ruff check --fix`: 6 unused imports (F401) and 3 `not X in Y` rewrites to `X not in Y` (E713).
3. 3e18993 excludes `utils/codegen.py` from ruff. The file asserts `sys.version_info.major == 2` on startup and uses the Python 2 `unicode` builtin; linting it against a Python 3 `target-version` produces false positives (F821 and E402).
4. 98747a8 fixes the 2 F841 unused-variable findings. In `pika/frame.py` the `out = properties.decode(...)` binding is dropped to match the adjacent `FRAME_METHOD` branch (`.decode()` returns `self` and is called for its side effect). In `examples/async_as_callback.py` `file_content` is renamed to `_file_content` to signal intentional non-use in the tutorial code.
5. e032b95 replaces `type(body) == bytes` with `isinstance(body, bytes)` in the `async_as_callback.py` example (E721).
6. a0367dc rewrites 3 `name = lambda: ...` assignments as nested `def` definitions (E731). Two in `pika/adapters/blocking_connection.py` (`is_done`, `common_terminator`); one in `tests/unit/select_connection_timer_tests.py` (`cb`), where the test asserts identity and needs a named callable.
7. 1637378 adds a file-scope `# ruff: noqa: E402` directive to `pika/__init__.py` next to the pre-existing `# pylint: disable=C0413`. The 11 E402 findings are all intentional: `__version__` is assigned on line 1 so submodules can do `from pika import __version__` at import time (`pika/adapters/utils/connection_workflow.py:17` does), and `logging.getLogger(__name__).addHandler(logging.NullHandler())` runs before submodule imports so any import-time logging has a handler installed.
8. 969726d strips 6 pre-existing trailing-whitespace instances across 3 files. Not introduced by this branch; surfaced while working in the area.
9. 01ce2f9 adds `.github/workflows/lint.yaml`, which runs `ruff check pika tests examples utils` on every push and on PRs targeting `main`, matching the trigger configuration of `main.yaml`. Uses `astral-sh/ruff-action@v3` with no version pin; ruff drift is accepted as a legitimate source of future CI fixes rather than flaky failures.

## Out of scope

- Removing pylint (its config, `# pylint:` comments, and `.vscode/settings.json`). Ruff does not implement every pylint rule, and running both in parallel is a legitimate option; the decision is deferred to a separate issue.
- `ruff format` (would replace yapf; separate change).
- Pre-commit integration (tracked in #1441).

## Types of Changes

None of the template categories fit cleanly: this is a CI tooling addition plus small code cleanups. The cosmetics box is arguably closest for the code changes but undersells the CI workflow.

## Checklist

No new tests: the CI lint job itself is the regression guard. No doc changes: `CONTRIBUTING.md` currently points at yapf for formatting and does not document linting; a follow-up doc change can mention `ruff check` once the approach is merged.


suchitd (approved):
Thanks @lukebakken for the detailed description in the PR.

PR #1556, opened 2026-04-30
Align Python versions metadata and docs 

alonfaraj:
## Proposed Changes

PyPI metadata and a few docs/code paths were out of sync with what the project actually supports and tests after recent chnages.

**Package metadata:** 

- Update `requires-python` to `>=3.10`. 
- Update Trove classifiers versions in `pyproject.toml`. The PyPI “Python versions” badge and install UI derive from that metadata.

**README:** Update to matches metadata: 3.10+, and a short note that **1.1.0** was the last release for **2.7**.

**CONTRIBUTING.md:** Removed wording that implied a Python 2 vs 3 split (“which on Python 3 might look like this”);

**Code cleanup:** `gevent_connection` dropped the dead `Queue` / Python 2 import branch.

---

## Types of Changes

- [x] Documentation (correction or otherwise)
- [x] Cosmetics (whitespace, appearance)

---

## Checklist


lukebakken (changes requested):
Just so you know, even though we're not testing on pre-3.10 versions, I do request that 3.7 - 3.9 are still listed as supported. Pika is installed on out-of-date systems _all of the time_, for better or worse. Thanks.

alonfaraj:
Ah sorry about that!
According to PEP 345, [Required-Python](https://peps.python.org/pep-0345/#requires-python) should specify 
> Python version(s) that the distribution is guaranteed to be compatible with

Which is a bit risky without testing (especially for future development) so thought to remove those versions. 

I will revert it.

lukebakken (approved):
Thanks!

PR #1574, opened 2026-05-05
Support Python 3.7+ in CI and fix `typing_extensions` import

lukebakken:
## Summary

- Add Python 3.7, 3.8, and 3.9 to the CI test matrix for both the Linux and Windows jobs, so that what we test matches what `pyproject.toml` advertises via `requires-python = ">=3.7"` and the version classifiers.
- Move `from typing_extensions import TypedDict` inside the `TYPE_CHECKING` block in `select_connection.py`. `TypedDict` is only used in the `POLLER_PARAMS` type alias, which is itself defined under `TYPE_CHECKING`, so the import was running unnecessarily at module load time and created an undeclared runtime dependency on `typing_extensions`.

## Notes

A scan of the source found no other Python 3.7 incompatibilities: no walrus operator, no positional-only parameters, no dict merge operators, no 3.8+ `typing` imports, and `from __future__ import annotations` is already present on every source file.

### DIFF
--- a/.github/workflows/legacy-python.yaml
+++ b/.github/workflows/legacy-python.yaml
@@ -1,4 +1,4 @@
-name: pika
+name: pika (legacy Python)
 
 on:
   push:
@@ -14,7 +14,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: [ '3.10', '3.11', '3.12', '3.13', '3.14' ]
+        python-version: [ '3.7', '3.8', '3.9' ]
         test-tls: [true, false]
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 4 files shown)


suchitd (commented):
Based on the failing checks, I think we need to handle CI python support for '3.7', '3.8' & '3.9' through a separate GHa workflow(s).
1. Use Ubuntu 22.04 
2. mypy 1.20 requires python>=3.10, I checked "--python-version" flag locally but it doesn't work for 3.7 and 3.8.

lukebakken:
Good catches, thanks. I've pushed a follow-up commit that addresses both points:

- Moved 3.7/3.8/3.9 into a new `legacy-python.yaml` workflow that runs on `ubuntu-22.04` and `windows-latest`, keeping them separate from the modern workflow.
- Added a `python_version >= "3.10"` environment marker to the `mypy` line in `test-requirements.txt` so the install does not fail on older interpreters.

Issue #3, opened 2009-11-09
Implement 0-9-1's Field Array type, 'A', and all the rest of the 0-9-1 type extensions


tonyg:
Implement 0-9-1's table-type extensions, even in the 0-8 branch. They're jolly useful.

Issue #2, opened 2009-11-08
codegen/amqp_codege.py in Makefile


domas:
Hi, there is problem with Makefile with master brunch.
There is : $(AMQP_CODEGEN_DIR)/amqp_codegen.py but when you don't have ../rabbitmq-codegen/ there is no codegen/amqp_codegen.py in your folder

Issue #1426, opened 2023-03-10
Expand docs and examples with regard to thread-safety


lukebakken:
Inspired by this discussion:
https://groups.google.com/g/rabbitmq-users/c/fSrIb6df-yc

See the following code:
https://github.com/lukebakken/rabbitmq-users-pika-fSrIb6df-yc/blob/main/connection.py

We _could_ expand this into a thread-safe Client class. Worth investigating.

jpaluska:
Hey Luke, I'd be interested in working on this. I was thinking about adding a thread-safe publisher class example to the examples dir. It would be based off your code that you referenced. Is that cool?

lukebakken:
It's _always_ OK to open a PR here, so, yes. Thanks.

Dheeraj4103:
Hi @lukebakken 👋

I'd like to take a shot at implementing the thread-safe client class mentioned here.

My proposed approach:
- A `ThreadSafeConnection` wrapper that uses `threading.Lock` to serialize access to the connection
- Internally route all publish/channel operations through `add_callback_threadsafe` so the I/O loop thread is the only one touching `_tx_buffers`
- Include a multi-threaded test that reproduces the `IndexError: pop from an empty deque` without the fix and passes with it

This addresses the root cause seen in issues like #1144 and #511 where concurrent threads race on `_tx_buffers.popleft()`.

Before I start coding — a few quick questions:
1. Would you prefer the wrapper live in `pika/` as a new module, or in `examples/` as a reference implementation?
2. Any constraints on the public API shape I should know about?
3. Should I also update the FAQ/docs to reference this class?

Happy to open a draft PR once I have a working implementation. Let me know if this direction looks good!

lukebakken:
1 - it should be a new class, in a similar manner to `BlockingConnection`, `SelectConnection`, etc
2 - It should be API compatible.
3 - Yes, of course. At least one example would be helpful as well.

Issue #511, opened 2014-10-30
IndexError: pop from an empty deque


chubakur:
```
Traceback (most recent call last):
  File "./tracer.py", line 223, in <module>
    proc.run()
  File "./tracer.py", line 133, in run
    self._connection.ioloop.start()
  File "/usr/lib/python2.6/site-packages/pika/adapters/select_connection.py", line 138, in start
    self.poller.start()
  File "/usr/lib/python2.6/site-packages/pika/adapters/select_connection.py", line 433, in start
    self.poll()
  File "/usr/lib/python2.6/site-packages/pika/adapters/select_connection.py", line 494, in poll
    self._handler(fileno, event, write_only=write_only)
  File "/usr/lib/python2.6/site-packages/pika/adapters/base_connection.py", line 318, in _handle_events
    self._handle_write()
  File "/usr/lib/python2.6/site-packages/pika/adapters/base_connection.py", line 358, in _handle_write
    frame = self.outbound_buffer.popleft()
IndexError: pop from an empty deque
```

Sometimes it arises from additional thread
```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib64/python2.6/threading.py", line 532, in __bootstrap_inner
    self.run()
  File "/usr/lib64/python2.6/threading.py", line 484, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/lib/python2.6/site-packages/unirest/__init__.py", line 102, in __request
    callback(_unirestResponse)
  File "./tracer.py", line 176, in put_jobs_in_queue
    self.publish(ip)
  File "./tracer.py", line 98, in publish
    self._w_channel.basic_publish(exchange='', routing_key=self.W_QUEUE_NAME, body=message)
  File "/usr/lib/python2.6/site-packages/pika/channel.py", line 300, in basic_publish
    (properties, body))
  File "/usr/lib/python2.6/site-packages/pika/channel.py", line 1057, in _send_method
    self.connection._send_method(self.channel_number, method_frame, content)
  File "/usr/lib/python2.6/site-packages/pika/connection.py", line 1503, in _send_method
    self._send_frame(frame.Method(channel_number, method_frame))
  File "/usr/lib/python2.6/site-packages/pika/connection.py", line 1490, in _send_frame
    self._flush_outbound()
  File "/usr/lib/python2.6/site-packages/pika/adapters/select_connection.py", line 77, in _flush_outbound
    self.ioloop.poller.poll(write_only=True)
[...]

PR #1582, opened 2026-05-08
Add ThreadSafeConnection: fix _tx_buffers race and thread-safety bugs

### LINKED ISSUE #1144: IndexError: pop from an empty deque
here is the bug report

```
QUANTAXIS>> _AsyncBaseTransport._produce() failed, aborting connection: error=IndexError('pop from an empty deque',); sock=<socket.socket fd=2044, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6, laddr=('::1', 3520, 0, 0), raddr=('::1', 5672, 0, 0)>; Caller's stack:
  File "e:\pika\pika\adapters\utils\io_services_utils.py", line 1092, in _on_socket_writable
    self._produce()
  File "e:\pika\pika\adapters\utils\io_services_utils.py", line 817, in _produce
    chunk = self._tx_buffers.popleft()
IndexError: pop from an empty deque
  File "e:\pika\pika\adapters\utils\io_services_utils.py", line 1092, in _on_socket_writable
    self._produce()
  File "e:\pika\pika\adapters\utils\io_services_utils.py", line 817, in _produce
    chunk = self._tx_buffers.popleft()
IndexError: pop from an empty deque
QUANTAXIS>> _AsyncTransportBase._initate_abort(): Initiating abrupt asynchronous transport shutdown: state=1; error=IndexError('pop from an empty deque',); <socket.socket fd=2044, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6, laddr=('::1', 3520, 0, 0), raddr=('::1', 5672, 0, 0)>
QUANTAXIS>> Deactivating transport: state=1; <socket.socket fd=2044, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6, laddr=('::1', 3520, 0, 0), raddr=('::1', 5672, 0, 0)>
QUANTAXIS>> connection_lost: StreamLostError: ("Stream connection lost: IndexError('pop from an empty deque',)",)
QUANTAXIS>> AMQP stack terminated, failed to connect, or aborted: error-arg=StreamLostError: ("Stream connection lost: IndexError('pop from an empty deque',)",); pending-error=None
QUANTAXIS>> Stack terminated due to StreamLostError: ("Stream connection lost: IndexError('pop from an empty deque',)",)
QUANTAXIS>> Closing transport socket and unlinking: state=2; <socket.socket fd=2044, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6, laddr=('::1', 3520, 0, 0), raddr=('::1', 5672, 0, 0)>
QUANTAXIS>> Unexpected connection close detected: StreamLostError: ("Stream connection lost: IndexError('pop from an empty deque',)",)
Stream connection lost: IndexError('pop from an empty deque',)
QUANTAXIS>> BlockingConnection.close(200, 'Normal shutdown') called on closed connection.
QUANTAXIS>> [QUANTAXIS REALTIME] connecting to ('::1', 5672, 0, 0)


```

Dheeraj4103:
## Summary

Introduces `ThreadSafeConnection` (wrapping `SelectConnection`) and `ThreadSafeChannel` that route all writes through `add_callback_threadsafe`, making `_tx_buffers` exclusively owned by the IOLoop thread. Fixes the `IndexError: pop from an empty deque` race seen when multiple threads share a raw connection.

Consumer delivery callbacks are dispatched on a per-channel worker thread, never the IOLoop thread, matching the architecture of the RabbitMQ Java, .NET, and Bunny clients. The single-worker-per-channel design guarantees FIFO delivery while letting blocking work (database writes, `queue_declare`, etc.) run safely inside callbacks.

Closes #1144, #511, #1426

## API surface

`ThreadSafeChannel` wraps `pika.channel.Channel`:

- Fire-and-forget: `basic_publish`, `basic_ack`, `basic_nack`, `basic_reject`
- Blocking RPCs (10s default timeout, `None` to wait indefinitely): `basic_qos`, `basic_consume`, `basic_cancel`, `basic_get`, `confirm_delivery`, `queue_declare`, `queue_delete`, `queue_purge`, `queue_bind`, `queue_unbind`, `exchange_declare`, `exchange_delete`, `exchange_bind`, `exchange_unbind`, `close`
- `abort` (close-and-swallow)
- Listeners (dispatched on the per-channel worker pool): `add_on_return_callback`, `add_on_cancel_callback`
- Properties: `channel_number`, `is_open`, `is_closed`

`ThreadSafeConnection` wraps `SelectConnection`:

- Constructor / `channel()` / `close()` all block with a 10s default timeout
- `abort` (close-and-swallow)
- Listeners (dispatched on a connection-level worker pool): `add_on_connection_blocked_callback`, `add_on_connection_unblocked_callback`
- `add_callback_threadsafe(callback)` for scheduling work on the IOLoop directly
- Context manager support

`ThreadSafeConnection` exported from `pika/adapters/__init__.py` and `pika/__init__.py`. `examples/basic_publisher_threaded.py` updated.

## Threading model

- One IOLoop thread per connection (`pika-ioloop-N`); all socket I/O happens here.
- Per-channel single-worker pool (`pika-consumer`) dispatches deliveries, publisher confirms, return-listener and cancel-listener notifications. FIFO and serial.
- Per-connection single-worker pool (`pika-conn-N`) dispatches `Connection.Blocked`/`Unblocked` so a slow listener can't stall heartbeats.
- All user callbacks run through `_safe_dispatch`, which logs unhandled exceptions via `LOGGER.exception` so they aren't silently lost on unobserved `Future` objects.

## Tests

- 103 unit tests in `tests/unit/thread_safe_connection_tests.py` covering all public methods, error paths, timeouts and waiter cleanup, dispatch-pool lifecycle, and exception logging.
- 7 acceptance tests in `tests/acceptance/thread_safe_connection_test.py` against a real broker, including the original `_tx_buffers` race regression (10 threads behind a `Barrier`), broker drops while blocked in `channel()` / `queue_declare()`, and concurrent `close()` from 5 threads.

## Follow-ups (deferred to 2.0.0)

- #1600 — Bound work queues. The executors currently use unbounded queues; Java caps at 1000 with optional offer-timeout. Deferred because `ThreadPoolExecutor` doesn't support `maxsize` and the fix is structural.

## Types of Changes

- [x] Bugfix (closes #1144, #511, #1426)
- [x] New feature
- [ ] Breaking change
- [x] Documentation

## Checklist


Dheeraj4103:
@lukebakken can you help me to figure out why yapf is failing

alonfaraj:
> @lukebakken can you help me to figure out why yapf is failing

Simply install hatch (`pip install hatch`) and run `hatch run fmt`.

Fixed it for you 79371294a793aab40616000b4291b68bee765665

lukebakken:
@Dheeraj4103 - I will have several changes to push to this branch, so I'm marking this PR as "draft" for now.

lukebakken:
@Dheeraj4103 as you can see, creating `ThreadSafe*` classes is a complex task. I do intend to finish this work as it will hopefully fix the most common issues encountered with this library.

lukebakken (commented):
## Fresh-eyes re-review

Independent re-read of the threading model after time away from the code, plus a re-audit of the RabbitMQ Java client and a fresh audit of the .NET client. Findings recorded here for posterity. None are blockers; severity noted per item.

### Correctness findings

**1. Late `basic_get` reply can be misrouted to a subsequent caller (medium-high)**

If `basic_get` times out and the user calls `basic_get` again, a late `Basic.GetOk` or `Basic.GetEmpty` for the *first* call fires the second call's `one_shot` callback. The second caller gets the first call's result (or empty), and when the actual second response arrives both callbacks are already removed (one-shot fired) so it is dropped. Caller gets wrong data.

Root cause is broader than `ThreadSafe*`: `pika.channel.Channel` has no reply-type correlation, no per-call request ID, and no way to deregister `one_shot` on timeout. The Java client uses `BlockingRpcContinuation.canHandleReply()` plus `_checkRpcResponseType` to discard mismatched late replies (`AMQChannel.java:561-614`). Worth tracking as a separate issue.

**2. `ThreadSafeConnection.channel()` race against connection close (medium)**

In `channel()` (around line 1107), if the connection closes (`Connection.Close` from broker, or socket death) between `_on_open` firing and the user thread waking, the sequence is:

1. `_on_open` runs on the IOLoop thread, sets `result[0] = ch`, sets `ready`.
2. `_on_connection_closed` then runs (also on the IOLoop thread). It walks `_channels` to shut down their consumer pools — but the new channel has not been appended yet.
3. The user thread wakes, sees `error[0]` is `None`, constructs `ThreadSafeChannel(result[0], self)`, and appends to `_channels` (line 1146).

Resulting state: the user holds a `ThreadSafeChannel` whose underlying `Channel` is dead, and whose `_consumer_work_pool` will never be reached by `_shutdown_all_consumer_pools()` (which already ran). The pool's daemon worker thread leaks until process exit. Subsequent calls on the channel correctly raise `_closed_reason` via `_register_waiter` so the channel is unusable rather than hanging — but the thread leak is real.

Fix: re-check `_closed_reason` after `ready.wait()` returns, before constructing the channel; raise it if set. Alternatively, append-to-`_channels` and `_closed_reason` check should happen under the same lock acquisition.

**3. Pool leak on `__init__` failure (low-medium)**

`_connection_work_pool` is created at line 994 *before* the `SelectConnection` constructor at line 1000. If `SelectConnection.__init__` raises (bad parameters, immediate getaddrinfo failure, etc.), or if the connection-open wait times out and `__init__` raises `TimeoutError` (line 1043), the executor pool's worker thread leaks. `_run_ioloop` does call `_shutdown_connection_pool()` after `ioloop.start()` returns, but in the `SelectConnection`-construction-failure path the IOLoop thread was never started, so cleanup never runs.

Accumulates if a caller retries connection construction in a tight loop after failures. Fix: wrap the construction + wait in a `try/except` that calls `_shutdown_connection_pool()` and joins the IOLoop thread on failure.

**4. `_on_chan_close` and `_on_get_empty` callback retention (low)**

Each `_blocking_rpc` call appends an `_on_chan_close` closure to `pika.channel.Channel`'s callback table. `pika.channel.Channel` has no removal API, so closures retain `ready` / `error` / `result` references until the channel closes. The guard `if not ready.is_set()` (line 156) makes them no-ops at fire time, so this is a memory leak rather than a correctness bug.

Same shape applies to `basic_get`'s `_on_get_empty` registered as `one_shot=True` — survives until either it fires or the channel closes. For typical setup-time usage (a few topology calls) the retention is invisible. For pathological usage (10K+ blocking RPCs on a long-lived channel) it grows unbounded.

Worth a docstring note on `_blocking_rpc`. The deeper fix belongs in `pika.channel.Channel` (add a removal API).

### Design findings

**5. Confirms / returns / blocked listeners run on the work pool, not the I/O thread (intentional divergence)**

Re-audited both reference clients to confirm: Java's `ChannelN.callConfirmListeners`, `callReturnListeners`, and `processControlCommand`'s `Connection.Blocked` handler all run inline on the MainLoop thread (`ChannelN.java:362, 377-383, 503-525`). .NET's `Channel.HandleBasicAckAsync` / `HandleBasicReturnAsync` / `Connection.HandleConnectionBlockedAsync` invoke listener wrappers from the frame-processing loop on the main loop thread (`Channel.cs:674-724`, `Connection.Commands.cs:57-64`).

pika dispatches all five (return, cancel, confirm, blocked, unblocked) through a work pool. Safer (a slow listener cannot stall heartbeats) but slower. Documented in `confirm_delivery`'s docstring. Worth adding a similar note to the `add_on_*` docstrings for consistency.

**6. `confirm_delivery` does not guard against being called twice (low)**

`pika.channel.Channel.confirm_delivery` appends rather than replaces the ack-nack callback (`channel.py:643-646`). Calling `ThreadSafeChannel.confirm_delivery(cb1)` followed by `confirm_delivery(cb2)` results in both callbacks firing for every ack/nack. Pika-core foible inherited by the wrapper. Could be guarded in the wrapper with a "confirm already enabled" check.

### Comparison to .NET — additional observations

**7. .NET supports configurable consumer dispatch concurrency**

`ConsumerDispatcherChannelBase` defaults to single-worker concurrency (matching pika) but exposes a per-connection knob to opt into parallel delivery (giving up FIFO per channel). Worth considering for #1600 alongside the bounded-queue work, for users who want shared/concurrent dispatch.

**8. RPC timeout defaults**

Confirmed: Java 10min (`ConnectionFactory.DEFAULT_CHANNEL_RPC_TIMEOUT`), .NET 20s (`Constants.DefaultContinuationTimeout`), pika 10s. pika's default is the most defensive of the three.

### Recommendation

None of these are blockers for #1582. Suggest:

- **Fix #2 (`channel()` race) in this PR** — small, local fix; prevents a real leak.
- **Fix #3 (`__init__` cleanup) in this PR** — small, optional belt-and-suspenders.
- **File follow-up issue** for #1 (basic_get late reply) — structural, deserves discussion.
- **Document #4, #5, #6** in docstrings; file follow-up issue for the missing callback-removal API in `pika.channel.Channel` if interested.

lukebakken:
## 4-hour soak: passed

Ran a long-form soak against this branch from a separate harness ([lukebakken/pika-1582](https://github.com/lukebakken/pika-1582)) to put the cross-thread serialization through sustained pressure. Two `ThreadSafeConnection` instances (one publisher, one consumer), 8 publisher threads sharing a single `ThreadSafeChannel`, channel recycles every 5 minutes on both sides.

| Metric | Value |
|---|---|
| Duration | 14,400s (4h) |
| Published | 11,520,024 |
| Consumed | 11,520,024 |
| Diff | 0 |
| Errors (pub/con) | 0 / 0 |
| Channel recycles | 94 |
| Sustained rate | 800 msg/s |
| RSS drift | 27.5 MB → 28.2 MB (+2.5%) |
| Heartbeat | 30s, no broker-driven close |

The exact-equality publish/consume count is the headline: across ~23M AMQP frame round trips (publish + deliver + ack) with continuous channel churn, every message produced was acked, no `_tx_buffers` IndexError, no `ChannelWrongStateError` from late acks racing channel close, and no IOLoop stall under cross-thread `add_callback_threadsafe` load.

Harness, soak script, and failure-detection logic are at lukebakken/pika-1582@67fd2e8 if anyone wants to reproduce or extend it (publisher confirms is the obvious next thing to exercise; tracked in `FUTURE.md`).

lukebakken:
I converted to draft since there is more work to be done with publisher confirmations.

lukebakken:
## 4-hour publisher confirms run: passed

Ran a publisher-confirms harness ([lukebakken/pika-1582, confirms.py](https://github.com/lukebakken/pika-1582)) against this branch for 4 hours. Eight publisher threads share a single `ThreadSafeChannel` with `confirm_delivery` enabled; per-message confirm latency is tracked from the IOLoop-thread publish callback to the per-channel worker-thread ack/nack callback. A separate manual-ack consumer drains end-to-end.

| Metric | Value |
|---|---|
| Duration | 14,400s (4h) |
| Published | 11,520,336 |
| Acked | 11,520,336 |
| Nacked | 0 |
| Consumed | 11,520,336 |
| Pending (at every sample) | 0 |
| Errors (pub/con) | 0 / 0 |
| Sustained rate | 800 msg/s |
| Confirm latency (mean) | 0.4ms |
| Confirm latency (p99) | 0.9ms |
| Confirm latency (max) | 1.6ms |
| RSS drift | 28.0 MB to 30.6 MB (+9%) |
| Heartbeat | 30s, no broker-driven close |

The exact triple-match (published = acked = consumed) across 11.5M messages confirms that the threading model around `confirm_delivery` works correctly: the ack/nack callback runs on the per-channel worker thread without stalling the IOLoop, confirmations arrive in lockstep with publishes under sustained cross-thread load, and no confirmations are lost during the 4-hour run.

Combined with the earlier basic soak (11,520,024 messages, 94 channel recycles, 0 errors), this covers the two primary production paths: fire-and-forget multi-threaded publish and confirmed multi-threaded publish.

lukebakken:
@alonfaraj @suchitd - anything to add here? I'm thinking that, in Pika 2, the new `ThreadSafe*` classes will be the de-facto `Connection` / `Channel` classes to use.

alonfaraj:
That's great! LGTM :rocket: 
Agree with the future release to make it the default classes

Issue #1580, opened 2026-05-06
Update example code to work with RabbitMQ 4.3


lukebakken:
I just tried `asynchronous_publisher_example.py` and it didn't work with RabbitMQ 4.3

PR #1585, opened 2026-05-09
fix: examples with rabbitmq 4.3 need durable queues

dadodimauro:
## Proposed Changes

Starting from version 4.3 of RabbitMQ, Transient (non-durable) non-exclusive classic queues are [deprecated](https://www.rabbitmq.com/release-information/deprecated-features-list) and cannot be declared by default (https://www.rabbitmq.com/docs/queues#durability)

This fixes the various examples to work with version 4.3, setting the `durable=True` when creating the queues

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #1580)

## Checklist

## Further Comments

@lukebakken I don't think changing the default value in the library code is a good idea, do you think we should document this? Any other action we should take?

[Open Source Saturday](https://lu.ma/open-source-saturday-torino)

### DIFF
--- a/examples/asynchronous_consumer_example.py
+++ b/examples/asynchronous_consumer_example.py
@@ -204,7 +204,7 @@ class ExampleConsumer(object):
         """
         LOGGER.info('Declaring queue %s', queue_name)
         cb = functools.partial(self.on_queue_declareok, userdata=queue_name)
-        self._channel.queue_declare(queue=queue_name, callback=cb)
+        self._channel.queue_declare(queue=queue_name, durable=True, callback=cb)
 
     def on_queue_declareok(self, _unused_frame, userdata):
         """Method invoked by pika when the Queue.Declare RPC call made in
--- a/examples/asynchronous_consumer_example.py
+++ b/examples/asynchronous_consumer_example.py
@@ -204,7 +204,7 @@ class ExampleConsumer(object):
         """
         LOGGER.info('Declaring queue %s', queue_name)
         cb = functools.partial(self.on_queue_declareok, userdata=queue_name)
-        self._channel.queue_declare(queue=queue_name, exclusive=True, callback=cb)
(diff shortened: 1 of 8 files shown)


examples/asynchronous_consumer_example.py
```diff
@@ -204,7 +204,7 @@ def setup_queue(self, queue_name):
         """
         LOGGER.info('Declaring queue %s', queue_name)
         cb = functools.partial(self.on_queue_declareok, userdata=queue_name)
-        self._channel.queue_declare(queue=queue_name, callback=cb)
+        self._channel.queue_declare(queue=queue_name, durable=True, callback=cb)
```
suchitd, line 207: Hi,
Thanks for the fix! 
Please use `exclusive=True` instead of durable queues. 
Please refer: https://github.com/pika/pika/pull/1548

dadodimauro: Hey! stupid question, but why did you reverted the changes to use durable queue instead of exclusive?

suchitd: Hi,

I reverted for some examples that requires non-exclusive access to the queue. If you notice, the `text` queue is common across [asynchronous_consumer_example.py](https://github.com/pika/pika/blob/main/examples/asynchronous_consumer_example.py) , [asynchronous_publisher_example.py](https://github.com/pika/pika/blob/main/examples/asynchronous_publisher_example.py) & [asyncio_consumer_example.py](https://github.com/pika/pika/blob/main/examples/asyncio_consumer_example.py). Since, it will be accessed  by different connection, we can't declare it as exclusive (Other connections will fail with `RESOURCE_LOCKED` when `text` is accessed).

suchitd (approved):

lukebakken:
Thank you @dadodimauro

PR #1586, opened 2026-05-09
fix: handle channel closure in BlockingConnection.process_data_events

dadodimauro:
## Proposed Changes

When using `BlockingConnection.process_data_events()` in long-running publisher applications (per `examples/long_running_publisher.py`), if RabbitMQ closes a channel due to consumer timeout or other broker-initiated reasons, the exception was silently swallowed. The connection remained open but the application had no way of knowing the channel was closed

This PR implement the existing behavior in https://github.com/pika/pika/blob/305fbe60088790a2648580078b4808f138e325d7/pika/adapters/blocking_connection.py#L2128 in `BlockingConnection.process_data_events`

The error was caused by `BlockingConnection.process_data_events()` not checking after dispatching channel events whether any channel had been closed by the broker. The channel closure detection logic only existed in the channel-level method. The fix is to, after dispatching channel events, to iterate over all channels and raise `ChannelClosedByBroker` if a channel is closed with that reason.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #1472 )

## Checklist

[Open Source Saturday](https://lu.ma/open-source-saturday-torino)

### DIFF
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -371,6 +371,9 @@ class BlockingConnection:
         # Receives on_close_callback args from Connection
         self._closed_result = _CallbackResult(self._OnClosedArgs)
 
+        # Store exceptions for server-initiated channel closures
+        self._server_channel_closures: deque[Exception] = deque()
+
         # Perform connection workflow
         self._impl: select_connection.SelectConnection = None  # type: ignore  # so that attribute is created in case below raises
         self._impl = self._create_connection(parameters, _impl_class)
@@ -899,13 +902,10 @@ class BlockingConnection:
         if self._channels_pending_dispatch:
             self._dispatch_channel_events()
 
-        for impl_channel in self._impl._channels.values():
-            channel = impl_channel._get_cookie()
-            if channel.is_closed and isinstance(
-                    channel._closing_reason, exceptions.ChannelClosedByBroker):
-                LOGGER.debug('Channel %d closed by broker: %r',
-                             channel.channel_number, channel._closing_reason)
-                raise channel._closing_reason
+        if self._server_channel_closures:
+            exc = self._server_channel_closures.popleft()
+            LOGGER.debug('Channel closed by broker: %r', exc)
+            raise exc
 
     def sleep(self, duration: float) -> None:
         """A safer way to sleep than calling time.sleep() directly that would
@@ -1492,6 +1492,7 @@ class BlockingChannel:
         self._closing_reason = reason
 
         if isinstance(reason, exceptions.ChannelClosedByBroker):
+            self.connection._server_channel_closures.append(reason)
             self._cleanup()
 
             # Request urgent termination of `process_data_events()`, in case
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -899,6 +899,14 @@ class BlockingConnection:
         if self._channels_pending_dispatch:
             self._dispatch_channel_events()
 
+        for impl_channel in self._impl._channels.values():
+            channel = impl_channel._get_cookie()
+            if channel.is_closed and isinstance(
+                    channel._closing_reason, exceptions.ChannelClosedByBroker):
+                LOGGER.debug('Channel %d closed by broker: %r',
+                             channel.channel_number, channel._closing_reason)
+                raise channel._closing_reason
+
     def sleep(self, duration: float) -> None:
         """A safer way to sleep than calling time.sleep() directly that would
         keep the adapter from ignoring frames sent from the broker. The
--- a/tests/unit/blocking_connection_tests.py
+++ b/tests/unit/blocking_connection_tests.py
@@ -358,7 +358,7 @@ class BlockingConnectionTests(unittest.TestCase):
         self.assertIs(unblocked_buffer[0], frame)
 
     @patch.object(blocking_connection.select_connection,
-                  'SelectConnection',
+                  "SelectConnection",
                   spec_set=SelectConnectionTemplate)
     def test_process_data_events_raises_on_channel_closed_by_broker(
             self, select_connection_class_mock):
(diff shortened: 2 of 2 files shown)


alonfaraj (commented):
I think scanning `self._impl._channels` here is not reliable after a broker-initiated channel close. In the real close path, `Channel._transition_to_closed()` dispatches the close callbacks and then `_cleanup()` runs; cleanup clears the channel cookie and also triggers the connection’s channel cleanup path. That means by the time `process_data_events()` reaches this new loop, the impl channel may no longer be in `_impl._channels`, or `_get_cookie()` may return None.

Maybe we could instead capture the `ChannelClosedByBroker` reason in `BlockingChannel._on_channel_closed()` and hand it to `BlockingConnection`, e.g. a pending channel-close exception queue/result on the connection? Then `BlockingConnection.process_data_events()` can raise from that stored reason after `_flush_output()`/dispatch without relying on closed impl channels still being present.

The current unit test should also model the real cleanup path, or at least cover `_get_cookie() is None` / channel removed from `_impl._channels`; otherwise it can pass while the production path still misses the closure.

For example, as far as I understand, this should pass using the suggested fix but it fails:

`pytest tests/unit/temp_test.py -k cleanup -vv`

```
from unittest import mock
from unittest.mock import patch

import pytest
import pika
from pika.adapters import blocking_connection


def test_process_data_events_should_raise_even_after_channel_cleanup():
    impl_mock = mock.Mock()
    impl_mock.is_closed = False
    impl_mock._channels = {}

    with mock.patch.object(
        blocking_connection.BlockingConnection,
        '_create_connection',
        return_value=impl_mock,
    ):
        connection = blocking_connection.BlockingConnection('params')

    exc = pika.exceptions.ChannelClosedByBroker(406, 'consumer_timeout')

    channel_mock = mock.Mock(
        is_closed=True,
        _closing_reason=exc,
        channel_number=1,
    )

    impl_channel_mock = mock.Mock()
    impl_channel_mock._get_cookie.return_value = channel_mock

    impl_mock._channels = {1: impl_channel_mock}
    connection._channels_pending_dispatch.add(1)

    def dispatch_and_cleanup():
        impl_mock._channels.clear()

    with mock.patch.object(connection, '_flush_output'), \
         mock.patch.object(connection, '_dispatch_channel_events',
                           side_effect=dispatch_and_cleanup):
        with pytest.raises(pika.exceptions.ChannelClosedByBroker):
            connection.process_data_events(time_limit=0)
```

@dadodimauro LMKWYT

dadodimauro:
@alonfaraj sorry for the late reply. Nice catch and thanks for the careful review! I have a fix ready, I'll push the changes later today once I finish testing it

tests/unit/blocking_connection_tests.py
```diff
@@ -357,3 +357,67 @@ def test_connection_unblocked_evt(self):
...
+
+        exc = pika.exceptions.ChannelClosedByBroker(406, "consumer_timeout")
+
+        impl_channel_mock = mock.Mock()
+        impl_channel_mock.channel_number = 1
+        channel = blocking_connection.BlockingChannel(impl_channel_mock,
+                                                      connection)
+
+        with mock.patch.object(channel, "_cleanup"):
+            channel._on_channel_closed(impl_channel_mock, exc)
```
dadodimauro, line 412: we store the exception in `_on_channel_closed`

tests/unit/blocking_connection_tests.py
```diff
@@ -357,3 +357,67 @@ def test_connection_unblocked_evt(self):
...
+
+        impl_mock._channels.pop(1, None)
+        impl_channel_mock._cookie = None
```
dadodimauro, line 415: simulate what `Channel._cleanup()` does after the callback, remove channel from connection's channel dict and clear cookie

tests/unit/blocking_connection_tests.py
```diff
@@ -357,3 +357,67 @@ def test_connection_unblocked_evt(self):
...
+
+        with mock.patch.object(connection, "_flush_output"):
+            with self.assertRaises(pika.exceptions.ChannelClosedByBroker) as cm:
```
dadodimauro, line 418: we check that `process_data_events` raise despite the channel being gone

pika/adapters/blocking_connection.py
```diff
@@ -905,6 +908,11 @@ def common_terminator():
         if self._channels_pending_dispatch:
             self._dispatch_channel_events()
 
+        if self._server_channel_closures:
+            exc = self._server_channel_closures.popleft()
+            LOGGER.debug('Channel closed by broker: %r', exc)
+            raise exc
```
dadodimauro, line 908: if we collected exceptions then raise

pika/adapters/blocking_connection.py
```diff
@@ -1492,6 +1500,7 @@ def _on_channel_closed(self, _channel: pika.channel.Channel,
         self._closing_reason = reason
 
         if isinstance(reason, exceptions.ChannelClosedByBroker):
+            self.connection._server_channel_closures.append(reason)
```
dadodimauro, line 1495: as suggested we store the exception in the `_on_channel_closed` callback, before cleanup

lukebakken (approved):
## Summary

The fix is correct and directly addresses @alonfaraj concern about the original approach.

## Original problem alonfaraj identified

The first version of this PR scanned `_impl._channels` after dispatching events to find closed channels. alonfaraj correctly noted the race: by the time `process_data_events()` reached that loop, `Channel._transition_to_closed()` -> `_cleanup()` had already run, removing the channel from `_impl._channels` and clearing its cookie. The check would find nothing and the exception would be lost.

## The updated fix

The approach in the current diff eliminates that race. In `BlockingChannel._on_channel_closed`, when the reason is `ChannelClosedByBroker`, the exception is stored on the connection **before** `_cleanup()` is called:

```python
self.connection._server_channel_closures.append(reason)
self._cleanup()
self.connection._request_channel_dispatch(-self.channel_number)
```

The wakeup call (`_request_channel_dispatch(-channel_number)`) was already present in the original code -- it causes `common_terminator()` to return True and terminate `_flush_output`. The new code adds the deque storage so `process_data_events()` has something to raise after waking up.

## Sequence under a broker-initiated close

1. `Channel.Close` frame arrives -> `_on_channel_closed()` fires
2. Exception stored in `_server_channel_closures` before cleanup
3. `_request_channel_dispatch(-N)` wakes up the ioloop waiter
4. `_flush_output()` terminates (negative channel number satisfies `common_terminator`)
5. `_dispatch_channel_events()` runs -- skips the negative channel number
6. `_server_channel_closures` is non-empty -> exception raised

## Tests

Both new tests are sound. `test_process_data_events_raises_after_full_channel_cleanup` explicitly clears `_impl._channels` and nulls `_cookie` before calling `process_data_events`, which is exactly the scenario alonfaraj's test was modelling. It passes with the new fix because the exception was stored in the deque before cleanup.

Note: alonfaraj's test in the review comment was written against the **old** approach (scanning channels after dispatch). It is not applicable to the new design and does not need to be added.

## Minor nit

- The `process_data_events` docstring could mention that `ChannelClosedByBroker` may be raised, but this is optional.

PR #1594, opened 2026-05-18
fix: codecov coverage upload

alonfaraj:
## Proposed Changes

codecov upload didn't work correctly for a few days.
Errors caused by multiple flags and not found gcov file error.

image

- Use explicit output path (xml:coverage.xml) so the file lands in the workspace root where the codecov action can find it
 - Combine per-job flags into a single hyphenated string to satisfy Codecov's one-flag-per-upload requirement
 
## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

## Fixes

- Fixes #

### DIFF
--- a/.github/workflows/legacy-python.yaml
+++ b/.github/workflows/legacy-python.yaml
@@ -36,7 +36,7 @@ jobs:
           python -m pip install --upgrade pip wheel
           python -m pip install hatch "virtualenv<21.0.0"
       - name: Test with pytest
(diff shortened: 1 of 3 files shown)


lukebakken (approved):
Thanks for noticing! I've gotta admit, I don't really "get" codecov 😉

alonfaraj:
> Thanks for noticing! I've gotta admit, I don't really "get" codecov 😉 

You don't like they break the CI every few weeks? 😆

PR #1595, opened 2026-05-19
tests: fix invisible tests and improve coverage

alonfaraj:
## Proposed Changes

This PR recovers tests that were silently ignored by `pytest` due to a naming convention mismatch.

`pytest` config in `project.toml` requires test methods to be prefixed with `test_`; two test files didn't follow that rule, causing their tests to be collected as classes but none of their methods to be executed.
Instead of extend `python_functions` to collect `*_test`, it would be better to stick with one convention.

#### `tests/unit/frame_tests.py`
- All 24 test methods used the `*_test` suffix pattern (e.g. `frame_marshal_not_implemented_test`).
- As a result, zero tests were running from this file under `pytest`.
- Renamed all methods to the `test_*` convention.
- Fixed one test bug where `DeliveryMode.Persistent` (an enum) was passed where an integer is required by using `.value`.

#### `tests/unit/select_connection_ioloop_tests.py`

- 8 test classes used `start_test` as their entry point but had no `test_*` method, making them invisible to `pytest`.
- Renamed `start_test` → `test_start` in those 8 classes.
- The following classes were intentionally left unchanged because `start_test` is used as a helper invoked by `test_normal`:
  - `IOLoopTimerTestSelect`
  - `IOLoopSleepTimerTestSelect`

#### `tests/unit/data_tests.py`

Added 11 tests covering previously untested paths in `data.py`, increasing coverage from **88% → 100%**:

- `test_encode_short_string_too_long` - Verifies `ShortStringTooLong` is raised for strings >255 bytes
- `test_decode_short_string_invalid_utf8` - Ensures non-UTF-8 short strings are returned as `bytes`
- `test_decode_value_short_short_int` - Covers type tag `b` (signed byte)
- `test_decode_value_short_short_uint` - Covers type tag `B` (unsigned byte)
- `test_decode_value_short_int` - Covers type tag `U` (signed short)
- `test_decode_value_short_uint` - Covers type tag `u` (unsigned short)
- `test_decode_value_long_uint` - Covers type tag `i` (unsigned 32-bit int)
- `test_decode_value_long_long_int_uppercase` - Covers type tag `L` (signed 64-bit int)
- `test_decode_value_float` - Covers type tag `f` (32-bit float)
- `test_decode_value_double` - Covers type tag `d` (64-bit double)
- `test_decode_value_long_string_invalid_utf8` - Verifies type tag `S` with non-UTF-8 content remains `bytes`

## Net Result

- Tests collected and passing: **634 → 670**
- `frame.py` coverage: **62% → 100%**
- `data.py` coverage: **88% → 100%**

## Types of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

## Checklist

## Fixes

Fixes #

## Further Comments

### DIFF
--- a/pika/heartbeat.py
+++ b/pika/heartbeat.py
@@ -74,6 +74,7 @@ class HeartbeatChecker:
         self._heartbeat_frames_received = 0
         self._heartbeat_frames_sent = 0
         self._idle_byte_intervals = 0
+        self._idle_heartbeat_intervals = 0
 
         self._send_timer = None
         self._check_timer = None
--- a/tests/unit/data_tests.py
+++ b/tests/unit/data_tests.py
@@ -4,6 +4,7 @@ pika.data tests
 """
 import datetime
 import decimal
+import struct
 import unittest
 from collections import OrderedDict
 from typing import ClassVar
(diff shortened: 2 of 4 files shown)


lukebakken (approved):
This is great, thanks yet again!

PR #1598, opened 2026-05-24
feat(CI): improved version release flow

alonfaraj:
## Proposed Changes

Replace the manual release process with a fully automated GitHub Actions workflow. Previously, releasing pika required manually updating the version in two files, running a Ruby changelog generator tool, creating GPG-signed tags, and publishing via the CLI. This was error-prone and undocumented tribal knowledge.

The new flow is a single command: `gh workflow run release.yaml -f bump=patch`. It handles version bumping, changelog generation, tagging, CI validation, PyPI publishing, and GitHub Release creation automatically.

### Key changes:

- **New `release.yaml` workflow** - `workflow_dispatch` triggered, bumps version in `pyproject.toml` and `pika/__init__.py`, auto-generates `CHANGELOG.md` via `requarks/changelog-action`, commits, tags, and pushes
- **Rewritten `publish.yaml` workflow** - triggers on version tag push instead of GitHub Release event, runs a CI gate (lint, format, typecheck, unit tests) before publishing, uses OIDC trusted publishing (no API tokens needed), creates a GitHub Release with auto-generated notes
- **New `.github/release.yml`** - configures PR label categories for auto-generated release notes
- **Removed `publish-pypi.yaml`** - replaced by the unified `publish.yaml`
- **Updated `RELEASE.md`** - documents the new automated flow
- Added `contents: read` to publish workflow permissions to fix `actions/checkout`
- Pre-release support routes to TestPyPI automatically based on PEP 440 suffixes
- Input validation prevents shell injection and empty pre-release tags

## Types of Changes

- [x] New feature
- [ ] Bugfix
- [ ] Breaking change
- [x] Documentation
- [ ] Cosmetics

## Checklist

### 1. One-time setup

- Create GitHub environments `pypi` and `testpypi` in repo Settings → Environments
- Configure trusted publishing on test.pypi.org: repo `pika/pika`, workflow `publish.yaml`, environment `testpypi`
- (Skip pypi.org setup until ready for a real release)

### 2. Dry-run with a pre-release

```bash
# This publishes to TestPyPI only - safe to run
gh workflow run release.yaml -f bump=patch -f prerelease=true -f prerelease_tag=rc1
```

### 3. Watch the run

```bash
gh run watch
```

### 4. Verify

- [ ] `release.yaml` computed version `1.4.1rc1`
- [ ] Version bumped in `pyproject.toml` and `pika/__init__.py`
- [ ] `CHANGELOG.md` updated with new entry
- [ ] Tag `1.4.1rc1` created and pushed
- [ ] `publish.yaml` triggered by the tag push
- [ ] CI gate passed (lint, format, typecheck, tests)
- [ ] Package published to TestPyPI
- [ ] GitHub Release created and marked as pre-release
- [ ] Install and test:
  ```bash
  pip install -i https://test.pypi.org/simple/ pika==1.4.1rc1
  python -c "import pika; print(pika.__version__)"
  ```

## Further Comments

**One-time setup required:** PyPI trusted publishing (OIDC) must be configured on pypi.org and test.pypi.org for the `pypi` and `testpypi` GitHub environments respectively. See RELEASE.md for details.

This replaces the token-based authentication (`PYPI_API_TOKEN` / `TEST_PYPI_API_TOKEN` secrets) with OIDC trusted publishing, which is the modern standard used by most major Python projects (requests, Flask, Pydantic, etc.).


lukebakken:
Hi @alonfaraj! Should this PR still be "draft"?

lukebakken (changes requested):
Reviewing the change overall: direction is right (workflow_dispatch trigger, OIDC trusted publishing, automatic GitHub Release creation). Grouped by the 🧞's confidence in each finding.

## High confidence

1. **GITHUB_TOKEN-pushed tag will not trigger `publish.yaml`.** Per GitHub's docs, events triggered by `GITHUB_TOKEN` do not create new workflow runs (with documented exceptions for `workflow_dispatch` and `repository_dispatch`, not `push`). So `release.yaml` pushes the tag, but the `on: push: tags:` trigger in `publish.yaml` does not fire. To chain workflows you need a PAT or a GitHub App token. Worth verifying with a pre-release before merge.

2. **Tag glob is too loose.** `'[0-9]+.[0-9]+.[0-9]+*'` accepts `1.2.3foo`, `1.2.3-broken`, etc. Tighten to explicit alternatives that match what `release.yaml` produces (`1.2.3` and `1.2.3rc1`-style).

3. **Two changelog generators producing two different changelogs.** `release.yaml` commits `CHANGELOG.md` via `requarks/changelog-action`; `publish.yaml` then calls `gh release create --generate-notes`, which generates different notes from the same data. Pick one. I would drop the committed `CHANGELOG.md` generation and rely on `--generate-notes` plus `.github/release.yml`. That also removes the third-party action.

4. **`actionlint` SC2001 on release.yaml:56.** `IFS='.' read -r major minor patch <<< "$(echo "$current" | sed 's/[a-zA-Z].*//')"` can be `<<< "${current%%[a-zA-Z]*}"`.

5. **`grep -P` is GNU-only.** Line 57 uses `grep -oP ... \K`. Works on `ubuntu-latest` but is a footgun on portability. `sed -nE 's/^version = "([^"]+)".*/\1/p' pyproject.toml` is equivalent.

## Strong recommendations

6. **Pin `requarks/changelog-action@v1` to a SHA.** Third-party action with `contents: write`; a moving tag is a supply-chain risk. Or remove it entirely if you adopt point 3.

7. **Add `concurrency:` to `release.yaml`.** Two near-simultaneous dispatches will race on the version bump. `concurrency: { group: release, cancel-in-progress: false }`.

8. **Reconsider the `ci` gate in `publish.yaml`.** It runs `lint-check`/`fmt-check`/`typecheck`/`unit` on one OS and one Python (3.14). That is a strict subset of the matrix that gates every PR pre-merge. Either drop it entirely (releases do not change code, only the version) or call the reusable test workflow that #1606 introduces. The current shape adds runtime without adding coverage.

## Suggestions

9. Move `id-token: write` from workflow scope to the two publish jobs that need it.
10. Pass `inputs.bump` via `env:` like the other inputs already do, rather than inline in the `case`.
11. Consider a `dry_run` input so `release.yaml` can be exercised without committing, tagging, or publishing.
12. Confirm branch protection on `main` allows `github-actions[bot]` to push directly, or this all stalls at the push step.

alonfaraj:
The headline change: **1 forced an architectural rethink**, and verifying the fix surfaced a PyPI constraint that reshaped the whole thing (details under 1). Net result - `publish.yaml` is gone and everything now lives in a single top-level `release.yaml`.

## High confidence

> 1. **GITHUB_TOKEN-pushed tag will not trigger `publish.yaml`.** ... To chain workflows you need a PAT or a GitHub App token.

Fixed by removing the cross-workflow event entirely. I first tried converting `publish.yaml` to a reusable `workflow_call`, but PyPI trusted publishing **rejects OIDC tokens minted inside a reusable workflow** (it matches the `job_workflow_ref` claim against the configured filename - [pypi/warehouse#11096](https://github.com/pypi/warehouse/issues/11096), [gh-action-pypi-publish#166](https://github.com/pypa/gh-action-pypi-publish/issues/166)). So the publish job *must* be top-level. Given that, I merged the publish jobs into `release.yaml` as `needs:` dependencies - no event to suppress, no PAT/App token needed. Note: trusted-publisher config must now point at workflow `release.yaml`.

> 2. **Tag glob is too loose.** `'[0-9]+.[0-9]+.[0-9]+*'` accepts `1.2.3foo` ...

Removed. There's no `on: push: tags:` trigger anymore.

> 3. **Two changelog generators producing two different changelogs.** ... I would drop the committed `CHANGELOG.md` generation and rely on `--generate-notes`.

Done. Dropped `requarks/changelog-action` and the committed `CHANGELOG.md`; release notes come solely from `gh release create --generate-notes` + `.github/release.yml`.

> 4. **`actionlint` SC2001 on release.yaml:56.** ... can be `<<< "${current%%[a-zA-Z]*}"`.

Fixed - using `${current%%[a-zA-Z]*}`.

> 5. **`grep -P` is GNU-only.** ... `sed -nE 's/^version = "([^"]+)".*/\1/p' pyproject.toml` is equivalent.

Fixed - switched to exactly that `sed -nE`.

## Strong recommendations

> 6. **Pin `requarks/changelog-action@v1` to a SHA.** ... Or remove it entirely if you adopt point 3.

Removed entirely (per 3).

> 7. **Add `concurrency:` to `release.yaml`.** ... `concurrency: { group: release, cancel-in-progress: false }`.

Added verbatim.

> 8. **Reconsider the `ci` gate in `publish.yaml`.** ... Either drop it entirely ... or call the reusable test workflow that #1606 introduces.

Dropped the gate. I also merged latest `main` so the branch carries #1606's reusable `_test.yaml`, and considered wiring it in as a pre-publish gate - but since the release commit only changes version strings (already matrix-gated pre-merge), it added runtime without real signal, so I left it gateless.

## Suggestions

> 9. Move `id-token: write` from workflow scope to the two publish jobs that need it.

Done - workflow default is `contents: read`; `id-token: write` is scoped to the two publish jobs only.

> 10. Pass `inputs.bump` via `env:` like the other inputs already do.

Done - `BUMP` env var.

> 11. Consider a `dry_run` input ...

Added - computes/bumps the version and prints the diff, skipping commit/tag/publish.

> 12. Confirm branch protection on `main` allows `github-actions[bot]` to push directly ...

Documented the requirement in `RELEASE.md`; will confirm the actual `main` protection setting before the first real release.

lukebakken (approved):
Thank you @alonfaraj

PR #1599, opened 2026-05-27
chore: upgrade RabbitMQ to 4.3.1 and pin Linux CI Docker image

alonfaraj:
## Proposed Changes

Upgrade RabbitMQ from `4.3.0` to `4.3.1` and ensure the Linux CI job uses the pinned version from `testdata/versions.json`.

Previously, the `build-linux` job in both `main.yaml` and `legacy-python.yaml` used the untagged `rabbitmq` Docker image (i.e. `latest`), while macOS and Windows were already reading the pinned version from `testdata/versions.json`. This change brings Linux into parity by introducing a `get-versions` job that reads the version from `versions.json` and passes it to the `services.image` tag.

## Types of Changes

- [ ] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Cosmetics

## Checklist

## Fixes

- Fixes #

## Further Comments

The `services:` block in GitHub Actions doesn't support dynamic expressions from steps, so a lightweight `get-versions` job is used to read the version and expose it as a job output. The `build-linux` job then consumes it via `needs.get-versions.outputs.rabbitmq` in the service image tag - keeping the `services:` structure intact.


lukebakken (approved):
Thanks!

PR #1609, opened 2026-06-02
feat(CI): Add TLS to Linux and macOS

### LINKED ISSUE #1605: Add TLS test runs to Linux and macOS CI
Currently only the Windows CI matrix runs the test suite with TLS enabled (`test-tls: [true, false]` in `.github/workflows/main.yaml` and `legacy-python.yaml`). Linux and macOS only run plaintext.

This means TLS-related code paths in pika are tested on a single OS, which has caused TLS regressions to slip through Linux/macOS CI in the past.

## Proposal

Extend the TLS matrix to Linux and macOS:

- **Linux**: configure the `services: rabbitmq:` container with a custom config and the project's certs (mounted via `volumes:` or built into a derived image).
- **macOS**: extend `.ci/macos/gha-setup.sh` (after #1582-era CI reorg lands) to drop a `rabbitmq.conf` that points at `tests/certs/`, similar to what `.ci/windows/gha-setup.ps1` does today.

## Notes

- This work was deliberately deferred from the CI reorganization PR (\`feat/ci-gold-plating\`) to keep that change focused.
- The existing certs in \`tests/certs/\` are already used by the Windows-on-TLS jobs and are checked in.

alonfaraj:
## Proposed Changes

Extend the CI test matrix to run pika's TLS acceptance tests on Linux and macOS, not just Windows. The reusable `_test.yaml` workflow now matrixes `test-tls: [true, false]` for all three OSes, passing `--use-tls` to pytest on the TLS leg.

On Linux this required moving RabbitMQ from a GH Actions `services:` container to a `docker run` launched by `.ci/linux/gha-setup.sh`, so we control the config (cert mount, TLS listener) instead of accepting the platform's stock one. macOS already started rabbitmq natively; it now also writes a `rabbitmq.conf` derived from the shared template via `RABBITMQ_CONFIG_FILE`.

The shared config template `.ci/rabbitmq.conf.in` (previously Windows-only at `.ci/windows/rabbitmq.conf.in`) is now the single source of truth, consumed by all three platforms.

Matrix job names on all three OSes are normalized to `build/test on <runner> py<X.Y> tls=<true|false>` so the TLS and python  is visible in the Actions UI checks list and make it clear.

### Why the Linux Docker setup needed extra work

Two CI-specific issues surfaced once `rabbitmq:latest` ran with our config under Docker on GitHub-hosted Linux runners:

1. **`loopback_users.guest = true` (template default) blocks AMQP.** Inside the container, traffic arriving on the published port comes from the Docker bridge IP, which RabbitMQ does not classify as loopback — so the `guest` user is rejected. The Linux script sed-overrides this to `false` for Docker only; the native macOS and Windows paths keep the loopback default.
2. **`log.console = false` hides the only available log surface.** Without console logging, `docker logs rabbitmq` is empty when startup fails. The Linux script sed-overrides to `true` so diagnostics are visible.

The Linux script also:
- Rewrites `PIKA_DIR/tests/certs` → `/etc/rabbitmq/certs` to match the cert bind mount.
- Polls `rabbitmqctl await_startup` as the `rabbitmq` user (not root, the `docker exec` default) and dumps `ls -la /var/lib/rabbitmq` on timeout so cookie/ownership problems are visible from the Actions log.

## Types of Changes

- [x] New feature
- [ ] Bugfix
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

## Checklist

- [x] I have added tests that prove my fix is effective

## Fixes

- Fixes #1605

## Further Comments

Things I considered and didn't do, in case reviewers want to push back:

- **Per-OS templates instead of sed overrides.** The current sed expressions in `.ci/linux/gha-setup.sh` are positional and would silently no-op if `.ci/rabbitmq.conf.in` is reformatted (e.g. `loopback_users.guest=true` without spaces). A parametric template (`@LOOPBACK_GUEST@`) or one template per OS would be more robust, at the cost of more files.


lukebakken (approved):
🔥 thanks!

PR #1610, opened 2026-06-02
Run tests in parallel via pytest-xdist by default

alonfaraj:
## Proposed Changes

Run unit tests in parallel via `pytest-xdist`. `hatch run unit` and `hatch run test` now default to `-n auto --dist=loadscope`, cutting the "Test with pytest" step by ~2–3× across all platforms.

`loadscope` (instead of xdist's default `load`) is required: it groups tests by class/module onto a single worker, which suites like `tests/unit/io_services_test_stubs_test.py` rely on — their `tearDownClass` asserts that every dynamically generated test method ran in the class. Under `load`, those methods get scattered across workers and the assertion fails.

Changes:
- `pyproject.toml`: add `pytest-xdist` to hatch test deps; bake `-n auto --dist=loadscope` into the `unit` and `test` scripts.
- `CONTRIBUTING.md`, `AGENTS.md`: document the parallel default, why `loadscope` is required, and the `-n 0` escape hatch for single-worker debugging.

Benchmark — "Test with pytest" step, `test-modern` matrix, Python 3.11, TLS off:

| OS             | Before (1 worker) | After (`-n auto`) | Workers | Speedup |
| -------------- | ----------------- | ----------------- | ------- | ------- |
| ubuntu-latest  | 41.19s            | 22.25s            | 4       | ~1.85×  |
| macos-latest   | 44.77s            | 15.05s            | 3       | ~2.97×  |
| windows-latest | 87.52s            | 29.56s            | 4       | ~2.96×  |

Sources: before — pika/pika main run [26768621604](https://github.com/pika/pika/actions/runs/26768621604); after — branch run [26819794216](https://github.com/alonfaraj/pika/actions/runs/26819794216).

## Types of Changes

- [ ] Bugfix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Cosmetics
- [x] Tooling / CI

## Checklist

## Fixes

- Fixes #

## Further Comments

`-n auto` lets pytest-xdist pick worker count from the host CPU count: 4 on GitHub's Linux/Windows runners, 3 on macOS. Local runs scale to whatever the dev machine has. Pass `-n 0` to disable parallelism entirely.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -94,6 +94,12 @@ hatch run test
 hatch run rabbitmq
 ```
 
+Tests run in parallel by default via `pytest-xdist` (`-n auto
+--dist=loadscope`). `loadscope` is required so classes that generate test
+methods dynamically (e.g. `tests/unit/io_services_test_stubs_test.py`) stay on
+a single worker — their `tearDownClass` asserts every generated method ran.
+Pass `-n 0` to disable parallelism.
+
 ## PR conventions
 
 - Base branch is `main`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,12 @@ To run unit tests only (no RabbitMQ required), use
 
     hatch run unit
 
+Tests run in parallel by default via `pytest-xdist` (`-n auto --dist=loadscope`).
+`loadscope` keeps tests from the same class on the same worker, which some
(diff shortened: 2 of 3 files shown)


lukebakken (approved):

lukebakken:
You're on 🔥 @alonfaraj !

PR #1576, opened 2026-05-05
Migrate Docs to MkDocs

alonfaraj:
## Proposed Changes

Migrate Docs to MkDocs.
Fixes #1571

**Preview**
image

- **Docs stack migrated from Sphinx/RST to MkDocs/Material**
  - Converts docs from `.rst` to `.md`
  - Adds `mkdocs.yml` and Material theme configuration
  - Removes old Sphinx/ReadTheDocs artifacts (`docs/conf.py`, `docs/Makefile`, `.readthedocs.yaml`)

- **Docs structure and navigation overhauled**
  - Reorganizes nav (including FAQ/Contributors placement)
  - Improves FAQ/content layout and module pages
  - Adds/updates `docs/changelog.md`, `docs/history.md`, `docs/contributing.md`, etc.

- **Examples/docs synchronization improved**
  - Moves inline code snippets from docs into real files under `examples/`
  - Uses MkDocs snippets/includes so docs reference executable example sources
  - Adds/updates several example scripts (TLS, blocking, tornado, URL params, etc.)

- **README and contributor docs refreshed**
  - README links/badges/visibility improvements
  - Better contributor guidance, including local docs preview and `mike` multi-version preview notes

- **CI and tooling updates for docs/linting**
  - Updates docs GitHub Actions workflow (newer action versions, trigger/permission changes) to build on each push for validation
  - Adds docs lint dependencies (`ruff`), includes formatting/lint cleanup commits
  - Removes legacy `.style.yapf` config in favor of `pyproject.toml` setup

- **Misc**
 - Create dynamic contributors graph instead of a static list
 - Add hatch correspond commands, update CI and docs to use them
 
## Types of Changes

- [ ] Bugfix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation (correction or otherwise)
- [x] Cosmetics (whitespace, appearance)

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] All tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

## Further Comments

Worth mention this warning on mkdocs build from the Material team
```
 │  ⚠  Warning from the Material for MkDocs team
 │
 │  MkDocs 2.0, the underlying framework of Material for MkDocs,
 │  will introduce backward-incompatible changes, including:
 │
[...]

PR #1614, opened 2026-06-08
Deploy versioned docs to GitHub Pages via mike

### LINKED ISSUE #1612: Deploy MkDocs documentation to GitHub Pages
## Summary

Deploy the MkDocs documentation to GitHub Pages and configure versioned docs via mike.

This was originally part of #1571 but split out since PR #1576 covers the docs build/conversion and CI gating, while deployment is a separate concern.

## Tasks

- [ ] Add a GitHub Actions workflow to build and deploy docs to GitHub Pages (triggered on push to `main` or release tags, using `mike deploy`)
- [ ] Configure GitHub Pages in repo settings (source: `gh-pages` branch or GitHub Actions artifact)

## Context

- `mike` is already a dependency in the docs hatch environment
- The docs build is validated in CI via `hatch run docs:build --strict` (gated by `tests-passed`)
- The deployment workflow should use `mike deploy` for versioned documentation, writing to the `gh-pages` branch

alonfaraj:
> [!NOTE]
> This description was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before it was posted. Most of the current diff is also Claude's work, replacing the dispatch-based design this pull request originally proposed. Claims under "What was verified" are commands actually run against this branch; anything not verified is marked as such.

## Proposed Changes

Publish the MkDocs documentation as a versioned site on GitHub Pages using [`mike`](https://github.com/jimporter/mike), which keeps every version in its own subdirectory of the `gh-pages` branch. Completes the deployment half of the docs work (follow-up to #1576).

The work lives in a reusable `.github/workflows/_deploy-docs.yaml`, which nothing triggers directly. Three callers decide what to publish:

| caller | when | publishes |
|--------|------|-----------|
| `main.yaml`, job `deploy-dev-docs` | push to `main`, after `tests-passed` | `dev` |
| `release.yaml`, job `deploy-docs` | end of a release or pre-release | `MAJOR.MINOR` for a release, the full version for a pre-release |
| `deploy-docs.yaml` | manual dispatch only | whatever is asked for |

`main.yaml` also runs `validate-docs-deploy` on every pull request: a full deploy with `push: false`, so a change that breaks only the publish path fails on the pull request rather than mid-release. It is in `tests-passed`, so it can block a merge.

`RELEASE.md` carries the design: the guards, the `mkdocs.yml` settings, the permissions rule and how to recover a skipped deploy.

### Why `uses:` rather than a dispatch

The tag push in `release.yaml` authenticates with `GITHUB_TOKEN`, and GitHub documents that "events triggered by the `GITHUB_TOKEN` will not create a new workflow run", so a tag trigger would be dead for every automated release while still firing unguarded for a hand-pushed tag. That trigger is therefore gone.

`release.yaml` calls the deploy with `uses:` rather than `gh workflow run` so that the deploy's conclusion is the release's conclusion. A dispatch returns as soon as the API accepts it, which would report a release as successful whether or not its documentation ever published, with no status check and nobody notified. The job runs last, after the PyPI publish, the smoke test and the GitHub Release, so a release that fails partway through never publishes docs for a version nobody can install.

### Guards

Three failure modes here are silent rather than loud, so each is checked:

- **Only a stable release may hold `latest`.** It backs the site root and every `latest/` URL compiled into a released wheel. Eligibility is decided by parsing with `packaging` in `.ci/docs_site.py`, not by pattern matching, so a pre-release, a post-release, a non-canonical spelling such as `1.05` and a typo such as `1.5.0rcl` are all refused. `dev` is the one exception, and only until a release exists.
- **An alias is never moved backwards.** A deploy on an older tag would otherwise take `latest` and the site-root redirect with it, rolling the whole site back for every reader. A stable release can still reclaim the alias from its own release candidate, which is a forward move.
- **The deploy is verified against the remote.** `mike` puts its push inside the block that downgrades an empty commit to a warning, so a no-op deploy exits 0 having pushed nothing. Checking that a version is merely listed is not enough either, because a patch release publishes into the same `MAJOR.MINOR` directory. The workflow records the `gh-pages` tip before deploying and requires it to move, then reads `versions.json` back and checks the site-root redirect points at the alias it set.

The alias policy and its tests live in `.ci/docs_site.py` and `tests/unit/docs_site_tests.py` rather than in shell, because a shell version of it was both untestable and wrong.

## Maintainer steps after merge

Ordered, and none of it can be done before the merge because `gh-pages` does not exist yet.

1. **Merge.** The push to `main` runs `deploy-dev-docs`, which publishes `dev`, takes the `latest` alias, writes the site-root redirect and creates `gh-pages`. The alias bootstrap is automatic; no manual `mike` commands are needed.
2. **Settings -> Pages -> Source:** `Deploy from a branch`, branch `gh-pages`, folder `/ (root)`. The branch has to exist first, so this comes after step 1.

Verified in advance rather than left to discover: @lukebakken has `admin` and `maintain` on pika/pika, the token carries `repo` scope, the `pika` organization permits Pages, no ruleset or branch protection covers `gh-pages`, and the repository's default workflow permissions are `write`.

## What was verified

- `hatch run docs:build` and a full `mike` deploy both succeed; `validate-docs-deploy` exercises the publish path on this pull request and passes.
- Building with `MIKE_DOCS_VERSION` set confirms `canonical_version: latest` takes effect, and removing the setting confirms it is load-bearing.
- The alias policy has unit tests covering dev-versus-release both ways, pre-release-versus-stable, numeric ordering, typos, non-canonical spellings, a populated site and an unreadable payload. Two mutations that previously passed the suite now fail it.
- Pre-release suffix collisions were measured with `packaging`: `c1`, `pre1` and `preview1` all normalize to `rc1`, which is why `prerelease_tag` accepts only `a`, `b` and `rc`.
- Tag formats, action SHA pins and the workflow JSON schema all check out.

Not verified: no release has run end to end with this in place, so the release path is checked by the job graph, by the pull-request rehearsal and by documentation rather than by observation.

## Fixes

- Closes #1612
- Follow-ups: #1712 (sitemaps under `canonical_version`)

### DIFF
--- a/.github/workflows/deploy-docs.yaml
+++ b/.github/workflows/deploy-docs.yaml
@@ -8,6 +8,11 @@ name: Deploy Docs
 #                              alias (and the default redirect) to it
 #   - pre-release tag       -> publishes under the full tag, leaves `latest`
 #                              pointing at the last stable release
+#
+# NOTE: tags pushed by release.yaml use GITHUB_TOKEN and therefore do NOT
+# fire the on.push.tags trigger below; release.yaml dispatches this workflow
+# explicitly on the tag ref instead. The tag trigger remains for manually
+# pushed tags.
 on:
   push:
     branches:
--- a/.github/workflows/deploy-docs.yaml
+++ b/.github/workflows/deploy-docs.yaml
@@ -1,0 +1,78 @@
+name: Deploy Docs
+
+# Build and publish the versioned MkDocs site to the `gh-pages` branch using
+# `mike`. GitHub Pages must be configured to serve from the `gh-pages` branch.
+#
+#   - push to `main`        -> publishes the `dev` version
+#   - stable release tag    -> publishes `MAJOR.MINOR` and moves the `latest`
+#                              alias (and the default redirect) to it
+#   - pre-release tag       -> publishes under the full tag, leaves `latest`
+#                              pointing at the last stable release
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+"
+      - "[0-9]+.[0-9]+.[0-9]+[a-zA-Z]*"
+  workflow_dispatch:
+
+# `mike` commits to and pushes the `gh-pages` branch.
+permissions:
+  contents: write
+
+# Serialize deploys so two runs cannot race on pushing `gh-pages`. Do not
+# cancel in progress: a half-finished `mike` push would leave the branch in an
+# inconsistent state.
+concurrency:
+  group: deploy-docs
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          # `mike` needs the full history of `gh-pages` to amend it, and tags
+          # for version detection.
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+          cache: pip
+
+      - name: Install documentation dependencies
+        run: python -m pip install hatch
+
+      # Fail fast on broken docs before touching `gh-pages`. `docs:build` runs
+      # `mkdocs build --strict`.
+      - name: Validate documentation build
+        run: hatch run docs:build
+
+      - name: Configure git
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Deploy development docs
+        if: github.ref == 'refs/heads/main'
+        run: hatch run docs:mike deploy --push --update-aliases dev
+
+      - name: Deploy release docs
+        if: startsWith(github.ref, 'refs/tags/')
+        env:
+          TAG: ${{ github.ref_name }}
+        run: |
+          if echo "$TAG" | grep -qE '[a-zA-Z]'; then
+            # Pre-release: publish under its own tag and leave `latest` alone.
+            hatch run docs:mike deploy --push "$TAG"
+          else
+            # Stable release: publish MAJOR.MINOR so patch releases update the
+            # same version in place, and move `latest` to it.
+            docs_version="${TAG%.*}"
+            hatch run docs:mike deploy --push --update-aliases "$docs_version" latest
+            hatch run docs:mike set-default --push latest
+          fi
--- a/.github/workflows/deploy-docs.yaml
+++ b/.github/workflows/deploy-docs.yaml
@@ -33,13 +33,13 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           # `mike` needs the full history of `gh-pages` to amend it, and tags
           # for version detection.
           fetch-depth: 0
 
-      - uses: actions/setup-python@v6
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: "3.12"
           cache: pip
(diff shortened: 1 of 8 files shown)


lukebakken (commented):
@alonfaraj - is there a way to update https://pika.readthedocs.io/en/stable/ and other, past versions, to point to the new docs site? All of the internet search engines will have been indexing that location for years.

alonfaraj:
> @alonfaraj - is there a way to update https://pika.readthedocs.io/en/stable/ and other, past versions, to point to the new docs site? All of the internet search engines will have been indexing that location for years.

I think the easiest and cleanest approach is to use the RTD builtin redirect which issues 301s and supports external destinations (RTD recommends this when moving off-platform) [here](https://docs.readthedocs.com/platform/latest/user-defined-redirects.html#migrating-your-documentation-to-another-domain).
You can set it in the RTD admin panel.
Not sure how the admin panel looks like those days but you can map specific url (e.g. for specific version) or just map everything to `/latest` github pages.

lukebakken:
@alonfaraj I do plan on getting to this soon! I'll need a couple hours to ensure that the readthedocs redirects are correct as well.

.github/workflows/deploy-docs.yaml
```diff
@@ -0,0 +1,78 @@
...
+concurrency:
+  group: deploy-docs
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
```
suchitd: We should use actions/checkout@v7

lukebakken:
## ReadTheDocs redirect plan

The old Sphinx docs at `pika.readthedocs.io` are indexed by every search engine. Since the new site uses MkDocs Material (directory URLs, no `.html`), a single wildcard redirect won't work - the URL shapes differ. Here's the page-redirect table that covers the indexed pages. All rules should be **forced** so they apply even when old pages still exist.

### Page redirects (apply across all RTD versions)

These are "Page Redirect" type in the RTD admin panel (they automatically cover `/en/stable/`, `/en/latest/`, `/en/1.3.2/`, etc.).

| From (RTD page path) | To (new site) |
|---|---|
| `/index.html` | `https://pika.github.io/pika/latest/` |
| `/intro.html` | `https://pika.github.io/pika/latest/intro/` |
| `/examples.html` | `https://pika.github.io/pika/latest/examples/` |
| `/faq.html` | `https://pika.github.io/pika/latest/faq/` |
| `/contributors.html` | `https://pika.github.io/pika/latest/contributors/` |
| `/modules/index.html` | `https://pika.github.io/pika/latest/modules/` |
| `/modules/channel.html` | `https://pika.github.io/pika/latest/modules/channel/` |
| `/modules/connection.html` | `https://pika.github.io/pika/latest/modules/connection/` |
| `/modules/parameters.html` | `https://pika.github.io/pika/latest/modules/parameters/` |
| `/modules/credentials.html` | `https://pika.github.io/pika/latest/modules/credentials/` |
| `/modules/exceptions.html` | `https://pika.github.io/pika/latest/modules/exceptions/` |
| `/modules/adapters/index.html` | `https://pika.github.io/pika/latest/modules/adapters/` |

### Catch-all (ordered last in the panel)

RTD evaluates rules in order - first match wins. Put this one at the bottom so named redirects above take priority:

- Type: **Exact Redirect**
- From: `/en/*`
- To: `https://pika.github.io/pika/latest/`
- Force: **Yes**

Anything not covered by the specific rules above (e.g. old API subpages, genindex, search) lands on the new site's home page - not a 404.

### Maintenance notes

- Do **not** delete the RTD project after configuring redirects - that kills all the redirect rules.
- Disable builds on the RTD project to avoid stale content being served alongside the redirects.
- RTD free-tier limit is 100 redirect rules. This plan uses ~13 (well under).
- If you later add more top-level docs pages, add a matching redirect rule.

Reviewed with the assistance of Claude Fable 5.

lukebakken:
@alonfaraj if you could take a pass at this PR yourself, I would appreciate it.

alonfaraj:
LGTM :+1: 
Pushed 3 small commits.

Issue #1618, opened 2026-06-08
Remove remaining 21 `type: ignore` comments


lukebakken:
## Summary

PR for #1565 reduced `type: ignore` comments from 167 to 26. This issue tracks removing the rest.

The inventory below was re-measured on 88df7ca2. There are now 21 suppressions, not 26: the `AbstractBase` category (10 of the original 26) was already eliminated by 33bddac7, which replaced `abc.ABCMeta('AbstractBase', (object,), {})` with `abc.ABC`. That was item 1 of the original plan, so it is done.

14 of the remaining 21 can be fixed in 1.5.0. The other 7 depend on adapter removal and are tracked separately in #1669.

**Current state.** All three PRs have merged. `main` at 295ad9e5 carries the 7 suppressions tracked in #1669 and nothing else, so the 1.5.0 scope of this issue is complete. See the verification section at the bottom. The inventory table below is the original 88df7ca2 baseline and is kept for reference.

## Measured inventory

Stripping all 21 suppressions and running mypy produces exactly 21 errors:

| File | Errors | Error codes |
|------|--------|-------------|
| `adapters/twisted_connection.py` | 9 | `override`, `func-returns-value` (2), `arg-type` (2), `misc`, `call-arg`, `assignment` (2) |
| `adapters/gevent_connection.py` | 4 | `import-untyped` |
| `adapters/base_connection.py` | 4 | `assignment` |
| `adapters/tornado_connection.py` | 2 | `arg-type` |
| `adapters/blocking_connection.py` | 1 | `assignment` |
| `adapters/select_connection.py` | 1 | `attr-defined, assignment, unused-ignore` |

## Fixable in 1.5.0 (14)

Verified in an isolated tree: mypy clean on all 14 with no suppressions, clean under both the default platform and `--platform darwin`, 903 unit tests pass, 558 acceptance tests pass against a live broker.

1. **`_StreamingProtocolShim` (4, `[assignment]`)**. The shim assigns bound methods to abstract method slots on the instance, and declares those slots as `None` at class level to make the class instantiable. Replace with four real `@override` methods that delegate to `conn._proto_*`. This does not depend on the `_core/` restructure in #1617.

2. **Tornado `IOLoop.instance()` (2, `[arg-type]`)**. `AbstractSelectorIOLoop` is a nominal base class, so tornado's `IOLoop` never satisfies it even though it implements every member. Converting `AbstractSelectorIOLoop` to a `typing.Protocol` fixes both.

[...]

PR #1619, opened 2026-06-08
Remove type: ignore comments from the codebase

### LINKED ISSUE #1565: Remove all type: ignore comments from the codebase
## Summary

The type annotations added in PR #1529 include 176 `type: ignore` comments across the pika source. These suppress real type errors or work around patterns that mypy cannot express cleanly. The goal is to eliminate all of them by fixing the underlying type issues.

## Current state

A quick breakdown of the categories:

- `type: ignore[arg-type]` -- argument type mismatches, often from `None` being passed where a concrete type is expected
- `type: ignore[assignment]` -- assignments that mypy cannot narrow (e.g., `None` placeholders overwritten later)
- `type: ignore[name-match]` -- `namedtuple` class names not matching the variable name
- `type: ignore[import-untyped]` -- third-party libraries without type stubs (gevent)
- `type: ignore[union-attr]` -- attribute access on union types where one branch lacks the attribute
- `type: ignore[misc]` -- various other suppressions

## Approach

- Audit each `type: ignore` comment and determine whether the fix is a code change, a type narrowing (`assert isinstance`), or a genuine limitation requiring suppression
- For `import-untyped` on gevent, consider contributing type stubs or using a `py.typed` marker
- For `namedtuple` name-match issues, consider switching to `typing.NamedTuple` subclasses
- Remove every `type: ignore` that can be replaced with correct types

## Related

- See #1564 for evaluating pyright as a type checker (some `type: ignore` comments may also affect pyright diagnostics)

lukebakken:
## Summary

- Reduce `type: ignore` comments from 167 to 26 by fixing the underlying type issues
- All 26 remaining suppressions are categorized with specific mypy error codes and represent genuine type system limitations (tracked in #1618 for 2.0)
- Zero functional changes to runtime behavior; all fixes are type annotations, assertions, variable renames, and `cast()` calls

## Approach

The dominant pattern (80+ occurrences) was fields initialized as `self._foo: SomeType = None` with a `# type: ignore[assignment]`, where the field is actually `SomeType | None`. The fix widens the type annotation and adds `assert self._foo is not None` at use sites where the code path guarantees non-None.

Other fixes include:
- Converting class-attribute `namedtuple` definitions to `NamedTuple` subclasses (eliminates name-match errors)
- Using `cast()` for mixin self-as-argument patterns and third-party API mismatches
- Introducing separate variables to avoid same-name type-changing reassignments
- Using `isinstance` checks for type narrowing instead of bare comparisons

## Remaining suppressions (26)

| Category | Count | Reason |
|----------|-------|--------|
| `[misc]` | 9 | Dynamic `AbstractBase` metaclass |
| `[import-untyped]` | 4 | gevent has no type stubs |
| `[method-assign]` | 4 | Instance-method override pattern in shim |
| `[arg-type]` | 4 | Third-party API signature mismatches |
| `[func-returns-value]` | 2 | Twisted API quirks |
| Other | 3 | Platform-specific, Twisted override/union |

## Test plan

- [ ] All existing unit tests pass (813 passed, 20 skipped for missing optional deps)
- [ ] mypy passes on all modified files with `--ignore-missing-imports`
- [ ] No trailing whitespace introduced
- [ ] No runtime behavior changes (assertions use the same invariants the code already relied on)

## Fixes

- Fixes #1565

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -3,7 +3,7 @@ name: lint
 on:
   push:
     branches:
-      - "**"
+      - main
   pull_request:
     branches:
       - main
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -3,7 +3,7 @@ name: pika
 on:
   push:
     branches:
-      - "**"
+      - main
   pull_request:
     branches:
       - main
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -3,7 +3,7 @@ name: mypy
 on:
   push:
     branches:
-      - "**"
+      - main
   pull_request:
     branches:
       - main
--- a/.github/workflows/yapf.yaml
+++ b/.github/workflows/yapf.yaml
@@ -3,7 +3,7 @@ name: yapf
 on:
   push:
     branches:
-      - "**"
+      - main
   pull_request:
     branches:
       - main
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -510,16 +510,23 @@ class _StreamingProtocolShim(nbio_interface.AbstractStreamProtocol):
 
     """
 
+    # Override abstract methods at class level to enable instantiation;
+    # actual implementations are assigned on the instance in __init__.
+    connection_made = None  # type: ignore[assignment]
+    connection_lost = None  # type: ignore[assignment]
+    eof_received = None  # type: ignore[assignment]
+    data_received = None  # type: ignore[assignment]
+
     def __init__(self, conn: BaseConnection) -> None:
         """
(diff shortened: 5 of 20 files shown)


alonfaraj:
Sorry for the mess 😅 - was trying to get `test_eintr` to actually run under xdist instead of being skipped. Since `signal.signal()` needs the main thread, I tried `pytest-forked` to fork each test into its own process. Worked locally, but broke CI on Windows (no `os.fork`) and Python 3.7 (forked test left the xdist worker's teardown state inconsistent). Reverted it all back to the main-thread skip🙏

alonfaraj (approved):

lukebakken:
Thanks for taking an additional look at that @alonfaraj, and for the review!

PR #1621, opened 2026-06-10
Fix: Wide typehints and docstrings

### LINKED ISSUE #1569: Review and update all inline code documentation and comments
## Summary

Review and update all inline code documentation (docstrings, comments, and type descriptions in `:param`/`:rtype` annotations) across the pika source for correctness and consistency.

## Scope

- All docstrings in `pika/` source files
- All inline comments
- `:param`, `:type`, `:rtype`, and `:returns` annotations in docstrings
- Module-level docstrings

## Known issues to address

- Docstring `:rtype` annotations that say `str` where the actual return type is `bytes` (several in `pika/frame.py`)
- Docstring parameter types that reference `str|unicode` (Python 2 artifact)
- Comments referencing removed features (e.g., backpressure detection)
- Stale `pylint: disable` comments for rules that no longer trigger
- Docstrings that contradict the type annotations added in PR #1529
- `# pylint: disable=C0111` and similar suppressions in files that now have docstrings

## Approach

- Audit each `pika/` source file
- Verify docstrings match actual function signatures and type annotations
- Remove or update stale comments and pylint suppressions
- Ensure consistency between docstring type descriptions and type annotations

alonfaraj:
## Proposed Changes

Add missing type annotations to `pika/*.py` and adapter modules using pyrefly and MonkeyType runtime tracing combined with static inference. Also augment Sphinx docstrings with missing `:param:` and `:rtype:` entries across the codebase.

All annotations use `from __future__ import annotations` (PEP 563) so the syntax is compatible at runtime with Python 3.7+. `Self` and `Literal` are imported from `typing_extensions` rather than `typing` for the same reason.

## Types of Changes

- [ ] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [x] Documentation 
- [x] Cosmetics 

## Checklist

## Fixes

- Fixes #1569 (partially?)

## Further Comments

**On `:param TYPE VAR_NAME:` in docstrings**

Many existing docstrings in this codebase use the Sphinx inline form `:param TYPE name: desc`. Our augmented docstrings follow the same convention for simple types (e.g. `:param int value:`), but for complex union/generic types (e.g. `Callable[[...], None]` or `X | Y`) we stripped the type from the `:param` directive entirely, leaving just `:param name: desc`.

This is worth discussing: it might be cleaner to drop the inline type from **all** `:param TYPE name:` directives going forward and rely solely on the function signature annotations instead. Reasons:

1. The inline type in `:param` duplicates what the annotation already says — two places to keep in sync.
2. Complex types with `|` or `[...]` break the docs build under `--strict` (griffe cannot parse them as field directives) and must remove spaces inbetween to fix.
3. Modern Sphinx and griffe can render parameter types directly from PEP 484 annotations, making the inline type redundant.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - chore/types-and-docstrings
   pull_request:
     branches:
       - main
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - chore/types-and-docstrings
   pull_request:
     branches:
       - main
--- a/pika/_utils.py
+++ b/pika/_utils.py
@@ -41,7 +41,7 @@ def time_now() -> float:
 def as_bytes(value: str | bytes) -> bytes:
     """Returns value as bytes.
 
-    :param value: string or bytes value to convert
+    :param value: value to convert
     :rtype: bytes
     """
     if not isinstance(value, bytes):
--- a/pika/_utils.py
+++ b/pika/_utils.py
@@ -86,9 +86,9 @@ def nonblocking_socketpair(
     Returns a pair of sockets in the manner of socketpair with the additional
     feature that they will be non-blocking.
 
-    :param int family: socket family (default AF_INET)
-    :param int socket_type: socket type (default SOCK_STREAM)
-    :param int proto: socket protocol (default 0)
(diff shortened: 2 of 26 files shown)


lukebakken (commented):
> This is worth discussing: it might be cleaner to drop the inline type from all :param TYPE name: directives going forward and rely solely on the function signature annotations instead

@alonfaraj this sounds good to me if you want to include it in this PR

lukebakken (approved):

PR #1632, opened 2026-06-24
Include py.typed marker in distribution

### LINKED ISSUE #1631: Include `py.typed` file in distribution
Hi, I see the recent releases add type hints in the library itself (#1529)

This is great, thanks! However I think a `py.typed` file is required for mypy to use these types when checking code that uses pika.

E.g. when I type check my code that uses pika, I get the following error from mypy:

```text
error: Skipping analyzing "pika": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information

For example, see the redis client library: https://github.com/redis/redis-py/blob/master/redis/py.typed

alonfaraj:
## Proposed Changes

Add a PEP 561 py.typed marker so type checkers recognize pika's inline type hints. Without it, mypy reports import-untyped when checking code that uses pika.

- Add empty `pika/py.typed` marker file
- Package it via setuptools package-data

## Types of Changes

- [ ] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

## Fixes

- Fixes #1631

## Further Comments

Verified locally: built the wheel and confirmed `pika/py.typed` is included in the distribution.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,6 +100,7 @@ include-package-data = false
 packages = ["pika", "pika.adapters", "pika.adapters.utils"]
 
 [tool.setuptools.package-data]
+"pika" = ["py.typed"]
 "*" = ["LICENSE", "README.md"]
 
 [tool.docformatter]


lukebakken (approved):
Huh, just an empty file. Thanks for doing this!

PR #1634, opened 2026-06-25
Extract hostname and port for error reporting: blocking_connection._c…

cumulus13:
## Proposed Changes

Extract hostname and port for error reporting: blocking_connection._c…

## Types of Changes

- [ ] Bugfix 
- [ ] New feature 
- [ x] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

- [x ] I have read the `CONTRIBUTING.md` document
- [ x] All tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Fixes

```python
def _create_connection(
        self,
        configs: None | (pika.connection.Parameters |
                         Sequence[pika.connection.Parameters]) = None,
        impl_class: select_connection.SelectConnection | None = None
    ) -> select_connection.SelectConnection:
        """
        Run connection workflow, blocking until it completes.

        :param configs: Connection
            parameters instance or non-empty sequence of them.
            Type: `None | pika.connection.Parameters | sequence`.
        :param impl_class: for tests/debugging only; implementation class.


        :raises Exception: on failure
        """
        if configs is None:
            configs = (pika.connection.Parameters(),)

        if isinstance(configs, pika.connection.Parameters):
            configs = (configs,)

        if not configs:
            raise ValueError('Expected a non-empty sequence of connection '
                             f'parameters, but got {configs!r}.')

        # Extract hostname and port for error reporting
        hostname = 'unknown'
        port = 'unknown'
        
        try:
            last_config = configs[-1] if isinstance(configs, (list, tuple)) else configs
            if hasattr(last_config, 'host'):
                hostname = last_config.host
            if hasattr(last_config, 'port'):
                port = last_config.port
        except (IndexError, AttributeError):
            pass

        # Connection workflow completion args
        #   `result` may be an instance of connection on success or exception on
        #   failure.
        on_cw_done_result = _CallbackResult(
            namedtuple('BlockingConnection_OnConnectionWorkflowDoneArgs',
                       'result'))

        resolved_impl_class = impl_class or select_connection.SelectConnection

        ioloop = select_connection.IOLoop()

        ioloop.activate_poller()
        try:
            impl_class.create_connection(
                configs,
                on_done=on_cw_done_result.set_value_once,
                custom_ioloop=ioloop)

            while not on_cw_done_result.ready:
                ioloop.poll()
                ioloop.process_timeouts()

[...]

PR #1635, opened 2026-06-25
Include hostname and port in connection error reporting

lukebakken:
## Summary

- Add `host` and `port` attributes to all connection-related exceptions so users can programmatically identify which broker failed
- `AMQPConnectorPhaseErrorBase` (and subclasses `AMQPConnectorSocketConnectError`, `AMQPConnectorTransportSetupError`, `AMQPConnectorAMQPHandshakeError`) now carry `host`/`port` from the connection parameters at the point of failure
- `AMQPConnectionError` gains optional `host`/`port` keyword args, propagated through to `__repr__`
- `ConnectionClosed` (and `ConnectionClosedByBroker`, `ConnectionClosedByClient`) accepts optional `host`/`port` kwargs
- All instantiation sites within `Connection` and `BaseConnection` pass `self.params.host`/`self.params.port`: `StreamLostError`, `ConnectionOpenAborted`, `IncompatibleProtocolError`, `ProbableAuthenticationError`, `ProbableAccessDeniedError`, `ConnectionBlockedTimeout`, `AMQPHeartbeatTimeout`, `ConnectionClosedByBroker`, `ConnectionClosedByClient`
- All adapter paths (asyncio, select, gevent, tornado, twisted, blocking, thread-safe) get structured host/port in errors
- Backward compatible: new parameters are keyword-only with `None` defaults
- Builds on the idea from #1634; supersedes it

## Test plan

- [x] Unit tests verify `host`/`port` attributes and repr output for every `AMQPConnectionError` subclass
- [x] Unit tests verify `host`/`port` on all `AMQPConnectorPhaseErrorBase` subclasses
- [x] Unit tests verify `None` defaults when host/port not passed
- [x] All existing tests pass (862 passed)
- [ ] Integration test: connect to unreachable host, verify exception carries host/port

### DIFF
--- a/pika/exceptions.py
+++ b/pika/exceptions.py
@@ -28,14 +28,19 @@ class AMQPConnectionError(AMQPError):
         self.host = host
         self.port = port
 
+    def _host_port_suffix(self) -> str:
+        """Return a ` (host=..., port=...)` repr suffix, or '' when unset."""
+        if self.host is not None:
+            return f' (host={self.host!r}, port={self.port})'
+        return ''
+
+    @override
     def __repr__(self) -> str:
         if len(self.args) == 2:
             base = f'{self.__class__.__name__}: ({self.args[0]}) {self.args[1]}'
         else:
             base = f'{self.__class__.__name__}: {self.args}'
-        if self.host is not None:
-            return f'{base} (host={self.host!r}, port={self.port})'
-        return base
+        return f'{base}{self._host_port_suffix()}'
 
 
 class ConnectionOpenAborted(AMQPConnectionError):
@@ -52,7 +57,7 @@ class IncompatibleProtocolError(AMQPConnectionError):
     def __repr__(self) -> str:
         return (
             f'{self.__class__.__name__}: The protocol returned by the server is not supported: {self.args}'
-        )
+            f'{self._host_port_suffix()}')
 
 
 class AuthenticationError(AMQPConnectionError):
@@ -61,7 +66,7 @@ class AuthenticationError(AMQPConnectionError):
     def __repr__(self) -> str:
         return (
             f'{self.__class__.__name__}: Server and client could not negotiate use of the {self.args[0]} '
-            'authentication mechanism')
+            f'authentication mechanism{self._host_port_suffix()}')
 
 
 class ProbableAuthenticationError(AMQPConnectionError):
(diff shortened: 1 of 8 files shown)


pika/exceptions.py
```diff
@@ -16,10 +16,22 @@ def __repr__(self) -> str:
 
 class AMQPConnectionError(AMQPError):
 
+    def __init__(self,
+                 *args,
+                 host: str | None = None,
+                 port: int | None = None) -> None:
+        super().__init__(*args)
+        self.host = host
+        self.port = port
```
suchitd, line 29: Although `AMQPConnectionError` accepts `host` and `port` argument, but it seems the subclasses (like `ConnectionOpenAborted`, `StreamLostError`, etc.) aren't updated where they are actually instantiated/raised in the codebase to pass these new keyword arguments (host=..., port=...).
So, essentially this change is not getting honoured.

lukebakken: Good observation. The subclasses you mention (`IncompatibleProtocolError`, `ProbableAuthenticationError`, `ProbableAccessDeniedError`, `AMQPHeartbeatTimeout`, `ConnectionBlockedTimeout`) are raised from within a `Connection` instance where `self.params.host`/`self.params.port` are available. I'll add a follow-up commit to pass those through.

Note that the primary goal of this PR was connection *establishment* errors (the `AMQPConnectorPhaseErrorBase` path), where previously there was no structured way to know which broker failed. That path is now covered. The subclass sites you identified are errors that occur after the connection is already up (or during the AMQP handshake phase inside `Connection`), so they're a secondary improvement - but a valid one.

alonfaraj (commented):
Following on from @suchitd's note about the raise sites (looks like the follow-up commit handles passing the kwargs through 👍) - there's a related gap one layer down on the display side. Now that host/port actually reach these exceptions, they still don't show up in `repr()` for any subclass that overrides `__repr__`:

```python
>>> repr(exceptions.ConnectionClosedByBroker(320, 'CONNECTION_FORCED', host='rabbit.local', port=5672))
"ConnectionClosedByBroker: (320) 'CONNECTION_FORCED'"   # host/port stored but not printed
>>> repr(exceptions.StreamLostError('eof', host='rabbit.local', port=5672))
"StreamLostError: ('eof',) (host='rabbit.local', port=5672)"   # shown
```

Same for `IncompatibleProtocolError`, `AuthenticationError`, `ProbableAuthenticationError`, `ProbableAccessDeniedError`, `NoFreeChannels`. A few of those are exactly the errors people log, so the broker info won't appear in their logs even though it's set. A small shared `_host_port_suffix()` helper on the base class that each custom `__repr__` appends would keep it consistent.

alonfaraj (approved):

lukebakken:
Thank you @alonfaraj!

PR #1637, opened 2026-06-29
Add ThreadSafe examples for publisher confirms and long-running publisher

### LINKED ISSUE #1607: Add ThreadSafe* examples for publisher confirms and long-running publisher
PR #1582 introduced `ThreadSafeConnection` / `ThreadSafeChannel` and ships two examples that exercise the new API:

- `examples/basic_publisher_threaded.py`
- `examples/basic_consumer_threaded.py`

A few of the existing `BlockingConnection`-based examples cover patterns where `ThreadSafe*` is a meaningful improvement and would be useful complements. Two candidates:

### 1. Publisher confirms (`examples/confirmation.py`)

A `ThreadSafe*` variant would show how to enable publisher confirms via `ThreadSafeChannel.confirm_delivery(ack_nack_callback)` and demonstrate that the ack/nack callback runs on the per-channel worker thread (not the IOLoop), so listener work cannot stall heartbeats.

### 2. Long-running publisher (`examples/long_running_publisher.py`)

The existing example exists to show the heartbeat-vs-work tension on a single-threaded `BlockingConnection`. With `ThreadSafe*` that tension goes away: the IOLoop runs on its own thread, so application work never blocks heartbeats. A `ThreadSafe*` variant would be a useful side-by-side comparison.

### Notes

- These are net-new examples, not modifications of the existing ones.
- Out of scope for #1582 itself, which is already large.
- Each example should have a docstring explaining how it pairs with (or differs from) its `BlockingConnection` counterpart.

alonfaraj:
## Proposed Changes

Adds the two `ThreadSafe*` examples requested in #1607, covering the scenarios where `ThreadSafeConnection` provides a meaningful improvement over `BlockingConnection`:

- **`examples/publisher_confirms_threaded.py`** - enables publisher confirms via `ThreadSafeChannel.confirm_delivery(ack_nack_callback)`. It tracks outstanding delivery tags (using the `on_publish` callback and the `next_publish_seq_no` counter) and blocks until the broker confirms all of them. The key teaching point is that the ack/nack callback is dispatched on the **per-channel worker thread**, not the IOLoop thread, so a slow confirm handler can never stall heartbeats. It also handles `multiple=True` acks correctly.

- **`examples/long_running_publisher_threaded.py`** - the `ThreadSafeConnection` counterpart to the existing `BlockingConnection`-based `examples/long_running_publisher.py`. It shows that the dedicated IOLoop thread keeps heartbeats flowing while the main thread sleeps far longer than the heartbeat interval, eliminating the `process_data_events()` / `add_callback_threadsafe` "heartbeat vs. work" juggling that the blocking design requires.

**Docs:** added a page for each example, wired them into the MkDocs nav and the `docs/examples.md` index (which was also missing the pre-existing Thread-Safe Consumer/Publisher entries, now filled in), and updated the `examples/README.md` table.

## Types of Changes

- [ ] Bugfix
- [ ] New feature
- [x] Documentation
- [ ] Breaking change
- [ ] Cosmetics

## Checklist

## Fixes

- Closes #1607

## Further Comments


lukebakken (approved):
Thanks!

PR #1638, opened 2026-07-01
Chore: general typing fixes

alonfaraj:
## Proposed Changes

Modernize pika's type-safety and remove Python 2-era legacy, delivered as a series of self-contained, individually-reviewable commits.

**1. Replace the Python 2/3 metaclass idiom with `abc.ABC`.** `pika/_utils.py` defined `AbstractBase = abc.ABCMeta('AbstractBase', (object,), {})` - a portability shim from the Python 2/3 era. pika is Python 3.7+ now, so it is obsolete, and it forced a `# type: ignore[valid-type, misc]` on every subclass, which **silently disabled type-checking across pika's entire abstract-class hierarchy**. Inheriting `abc.ABC` directly removed those suppressions and unmasked **48 latent mypy errors**, all fixed here. The largest was a genuine type bug: the socket/stream connectors were typed `AbstractIOServices | AbstractFileDescriptorServices`, but call methods from *both* - and a union exposes only members common to both (here, none). A new `AbstractFileDescriptorIOServices` combined interface models what the connectors actually require.

**2. Add `@override` (PEP 698) and enforce it via mypy.** 152 methods now declare override intent, with `explicit-override` enabled. **No new runtime dependency** - `override` is re-exported from `pika._utils`: stdlib on 3.12+, `typing_extensions` for the type-checker only, no-op fallback on 3.7–3.11.

**3. Remove redundant boilerplate** now covered by the above (`raise NotImplementedError` in abstract bodies; `Implement :py:meth:` pointer docstrings).

**4. Fix an interface gap** - `AbstractAMQPConnectionWorkflow.start`/`abort` weren't abstract, so the "interface" was instantiable.

**5. Tests & misc** - `@override` on the `Connection` test doubles, removed dead `call_later`/`remove_timeout` stubs, `setblocking(False)`, param-name fix.

### Why
The Python 2 shim turned off type-checking for pika's core interfaces, letting 48 errors accumulate unseen. Removing it + `@override` gives mypy real coverage going forward, with no runtime dependency and no drop in 3.7 support.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [x] Documentation 
- [ ] Cosmetics 

## Checklist

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -3,7 +3,7 @@ name: lint
 on:
   push:
     branches:
-      - '**'
+      - main
   pull_request:
     branches:
       - main
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -3,7 +3,7 @@ name: lint
 on:
   push:
     branches:
-      - main
+      - '**'
   pull_request:
     branches:
       - main
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -3,7 +3,7 @@ name: pika
 on:
   push:
     branches:
-      - '**'
+      - main
   pull_request:
     branches:
       - main
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -3,7 +3,7 @@ name: pika
 on:
   push:
     branches:
-      - main
+      - '**'
   pull_request:
     branches:
       - main
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -3,7 +3,7 @@ name: mypy
 on:
   push:
     branches:
-      - main
+      - '**'
   pull_request:
     branches:
       - main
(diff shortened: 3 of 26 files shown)


lukebakken (approved):
Solid PR — well-structured commits, each self-contained and clearly messaged.

**What this does well:**
- Removes the Py2/3 `AbstractBase` metaclass shim that was silently suppressing type checking across the entire abstract-class hierarchy
- `AbstractFileDescriptorIOServices` correctly models the combined interface the connectors/transports actually require (the old union type was unsound)
- `@override` + `explicit-override` mypy rule catches future signature drift with zero runtime cost
- `cast()` for the `_StreamingProtocolShim` is the right fix — it duck-types as `Connection` via `__getattr__` but isn't an instance

**One change made on this end:**
- Reverted the `.github/workflows/*.yaml` push-branch broadening (`**` → `main`) per the commit message's own "TEMPORARY ... Revert before merging upstream" note.

All local CI checks pass: fmt, lint, typecheck, and 846 unit tests.

-- reviewed with AI assistance (lukebakken)

PR #1641, opened 2026-07-06
Stop mutating global asyncio event loop policy on import

### LINKED ISSUE #1639: Importing pika.adapters can break asyncio subprocesses on Windows
On Windows, importing `pika.adapters`, which indirectly [imports](https://github.com/pika/pika/blob/41695b16f2bbad84651a4bbef4b08fc00ac7ef2a/pika/adapters/__init__.py#L23) `asyncio_connection`, [changes](https://github.com/pika/pika/blob/41695b16f2bbad84651a4bbef4b08fc00ac7ef2a/pika/adapters/asyncio_connection.py#L19) the process-wide asyncio event loop policy.

That can break unrelated asyncio code running in the same process.

Here is a minimal reproducer:

```python
import sys
import asyncio

import pika.adapters

async def main():
    await asyncio.create_subprocess_exec(
        sys.executable,
        "-c",
        "print('ok')",
    )


asyncio.run(main())
```

On Windows, this fails with `NotImplementedError` from `asyncio.base_events._make_subprocess_transport`.

We ran into this through Playwright. `async_playwright()` starts its driver process via `asyncio.create_subprocess_exec()`, so it fails if `pika.adapters` has already been imported.

Would it be possible to avoid changing the global asyncio policy at import time? For example, Pika could document the Windows loop requirement, or enforce it when `AsyncioConnection` is constructed instead of when the module is imported.

lukebakken:
## Summary

- Remove module-level `asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())` that fired on every `import pika.adapters` on Windows
- Instead, construct `asyncio.SelectorEventLoop()` directly in `_AsyncioIOServicesAdapter.__init__` when no loop is provided on Windows
- This gives pika the selector loop it needs for `add_reader`/`add_writer` without polluting the process-wide event loop policy

Fixes #1639

## Test plan

- [ ] Verify on Windows: `import pika.adapters` no longer changes `asyncio.get_event_loop_policy()`
- [ ] Verify on Windows: `AsyncioConnection` still works (gets a `SelectorEventLoop`)
- [ ] Verify on Windows: `asyncio.create_subprocess_exec()` works after importing pika (the reporter's reproducer)
- [ ] Verify on Linux/macOS: no behavior change (still uses `asyncio.new_event_loop()`)
- [ ] All existing CI checks pass

### DIFF
--- a/pika/adapters/asyncio_connection.py
+++ b/pika/adapters/asyncio_connection.py
@@ -16,9 +16,6 @@ if TYPE_CHECKING:
 
 LOGGER = logging.getLogger(__name__)
 
-if sys.platform == 'win32':
-    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
-
 
 class AsyncioConnection(base_connection.BaseConnection):
     """The AsyncioConnection runs on the Asyncio EventLoop."""
@@ -140,7 +137,10 @@ class _AsyncioIOServicesAdapter(io_services_utils.SocketConnectionMixin,
             try:
                 loop = asyncio.get_running_loop()
             except RuntimeError:
-                loop = asyncio.new_event_loop()
+                if sys.platform == 'win32':
+                    loop = asyncio.SelectorEventLoop()
+                else:
+                    loop = asyncio.new_event_loop()
         self._loop = loop
 
     @override
--- a/pika/adapters/asyncio_connection.py
+++ b/pika/adapters/asyncio_connection.py
@@ -141,6 +141,14 @@ class _AsyncioIOServicesAdapter(io_services_utils.SocketConnectionMixin,
                     loop = asyncio.SelectorEventLoop()
                 else:
                     loop = asyncio.new_event_loop()
+        if sys.platform == 'win32' and not isinstance(
+                loop, asyncio.SelectorEventLoop):
+            raise TypeError(
+                'On Windows, Pika requires an asyncio.SelectorEventLoop '
+                'because it uses add_reader/add_writer. The running loop '
+                f'is {type(loop).__name__}. Either pass a SelectorEventLoop '
+                'via custom_ioloop, or run with '
+                'asyncio.run(main(), loop_factory=asyncio.SelectorEventLoop).')
         self._loop = loop
 
     @override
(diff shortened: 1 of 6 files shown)


alonfaraj:
I can't run this here (no Windows), so I verified the call path by code, not execution. @lukebakken can you test to verify this theory?

---

Since Python 3.8 the default Windows loop is `ProactorEventLoop`. After this PR, `import pika.adapters` no longer forces the selector policy, so `asyncio.run(...)` gives a Proactor loop. If a user then builds an `AsyncioConnection` without passing a loop, `__init__` takes the `get_running_loop()` branch (not the new `SelectorEventLoop()` branch) and reuses the Proactor loop and pika's `set_writer`/`set_reader` call `loop.add_writer`/`add_reader`, which raise `NotImplementedError` on Proactor. This happens on the very first `set_writer` during the non-blocking connect, so **no reachable broker is required to reproduce**.

Minimal reproducer (Windows, Python 3.8+, **with this PR's branch checked out**):

```python
import asyncio
import pika
from pika.adapters.asyncio_connection import AsyncioConnection

def on_open_error(_conn, err):
    print(f"open error: {err!r}")
    asyncio.get_running_loop().stop()

async def main():
    loop = asyncio.get_running_loop()
    print("running loop:", type(loop).__name__)   # -> ProactorEventLoop on Windows

    # No loop passed -> _AsyncioIOServicesAdapter reuses the running Proactor loop.
    # 127.0.0.1:1 doesn't need to be a broker; the failure is at the first add_writer.
    AsyncioConnection(
        parameters=pika.ConnectionParameters(host="127.0.0.1", port=1),
        on_open_error_callback=on_open_error,
    )
    await asyncio.sleep(3)

asyncio.run(main())   # Windows default -> ProactorEventLoop
```

Expected on this branch (Windows): a `NotImplementedError` out of `add_writer`. On current `main` it "works" only because the import-time policy hack forces `asyncio.run` to build a selector loop.

Suggestion: in `_AsyncioIOServicesAdapter.__init__`, when `sys.platform == 'win32'` and the running loop isn't a `SelectorEventLoop`, raise a clear error (or `warnings.warn`) so this fails loudly instead of deep inside asyncio.

lukebakken:
Merging since @mike386 checked these changes in his env.

PR #1645, opened 2026-07-09
fix: connection_is_idle closes connection after 1 idle interval instead of 2

maxtaran2010:
## What is wrong

`HeartbeatChecker.connection_is_idle` (in `pika/heartbeat.py`) returns
`True` as soon as `_idle_byte_intervals > 0`.  Because `_idle_byte_intervals`
starts at `0` and is incremented by `1` on every check cycle that sees no new
bytes, this means the connection is torn down after **a single idle check
interval** — not two, as the AMQP spec and the RabbitMQ documentation require.

The relevant comment already in the file acknowledges the spec:

> There is a certain amount of confusion around how client developers
> interpret the spec.  The spec talks about **2 missed heartbeats** as a
> *timeout*...

Yet the property implements only 1 missed interval as the threshold.

The property docstring also says "in **enough** intervals", but `> 0` means
any positive value — i.e., exactly one interval — is already "enough".

## Impact

Any transient condition that causes the check timer to fire during a
brief gap in network activity — GC pause, OS scheduler jitter, transient
load spike — will raise `AMQPHeartbeatTimeout` and drop a perfectly healthy
connection.  This is especially visible with small heartbeat timeouts (the
minimum allowed value is 1 second).

## Fix

Change the threshold from `> 0` to `> 1` so the connection is only
considered idle after **two** consecutive idle check intervals, matching
the AMQP spec and the existing comment.

```python
# before
return self._idle_byte_intervals > 0

# after
return self._idle_byte_intervals > 1
```

Two regression tests are added to pin the boundary:
- `test_connection_is_idle_false_after_one_interval` — asserts that a
  single missed interval does **not** trigger idle.
- `test_connection_is_idle_true_after_two_intervals` — asserts that two
  missed intervals **do** trigger idle.

All 33 existing heartbeat unit tests continue to pass.

### DIFF
--- a/pika/heartbeat.py
+++ b/pika/heartbeat.py
@@ -89,8 +89,13 @@ class HeartbeatChecker:
     def connection_is_idle(self) -> bool:
         """Returns true if the byte count hasn't changed in enough intervals to trip the max idle
         threshold.
+
+        Per the AMQP spec and the RabbitMQ documentation, a connection is considered idle only
+        after *two* consecutive check intervals with no activity.  Using ``> 0`` would close the
+        connection after a single missed interval, which is far too aggressive and causes spurious
+        disconnects under transient load or GC pauses.
         """
-        return self._idle_byte_intervals > 0
+        return self._idle_byte_intervals > 1
 
     def received(self) -> None:
         """Called when a heartbeat is received."""
--- a/tests/unit/heartbeat_tests.py
+++ b/tests/unit/heartbeat_tests.py
@@ -102,8 +102,20 @@ class HeartbeatTests(unittest.TestCase):
     def test_connection_is_idle_false(self):
         self.assertFalse(self.obj.connection_is_idle)
 
+    def test_connection_is_idle_false_after_one_interval(self):
+        # A single missed check interval must NOT be considered idle.
+        # The AMQP spec requires two consecutive missed heartbeats before
+        # declaring the connection dead.
+        self.obj._idle_byte_intervals = 1
+        self.assertFalse(self.obj.connection_is_idle)
+
     def test_connection_is_idle_true(self):
         self.obj._idle_byte_intervals = self.INTERVAL
+        self.assertTrue(self.obj.connection_is_idle)
+
+    def test_connection_is_idle_true_after_two_intervals(self):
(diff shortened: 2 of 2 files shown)


lukebakken:
Thank you for the PR, but the `> 0` threshold is intentional and this change would be a regression, so I am closing it.

The premise of the PR is that "2 missed heartbeats" must be implemented as a counter reaching 2. In Pika it is implemented in the *width of the check window* instead. The key is that `_check_interval` is not the heartbeat send interval - it is `timeout + 5` seconds (see the comment block in `HeartbeatChecker.__init__` and #1072, specifically 4649f87f89317bf8893a6320ecdd1e2c68c86aa0). RabbitMQ sends a heartbeat frame every `timeout / 2` seconds when its side of the connection is otherwise idle (`start_heartbeat_sender` in [rabbit_heartbeat.erl](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit_common/src/rabbit_heartbeat.erl), which uses `TimeoutSec * 1000 div 2`). So a single Pika check window with **zero** bytes received already spans two full broker heartbeat send intervals, plus 5 seconds of slack. `_idle_byte_intervals > 0` therefore already encodes "two missed heartbeats" - the `__init__` comment you quoted is describing that design, not an unfulfilled TODO.

With this change, teardown would require two consecutive idle windows, i.e. `2 * (timeout + 5)` seconds of total silence - four or more missed broker heartbeats. With the default 60 second timeout, Pika would sit on a dead socket for up to 130 seconds before noticing. Detecting bad connections quickly was an explicit goal of the current design (again, see #1072).

On the stated impact: a GC pause or scheduler jitter cannot trigger a spurious disconnect here. The idle condition requires zero bytes received across the entire `timeout + 5` window, during which the broker sends at least two heartbeats. If nothing at all arrived in that window, the connection is genuinely dead. A client-side pause also stalls the check timer itself (same event loop), so it cannot manufacture false idleness. And with the minimum timeout of 1 second the check window is 6 seconds, covering roughly twelve broker heartbeat intervals - even less fragile, not more.

The one fair point is that the docstring's "in enough intervals" wording invites exactly this misreading. I will clarify that documentation separately.

PR #1633, opened 2026-06-24
Chore: expand ruff rules

alonfaraj:
## Proposed Changes

Enable three additional Ruff lint rule families to guard against common issues going forward:

- `G` (flake8-logging-format) - enforces lazy `%`-style args in log calls
- `ISC` (implicit-string-concatenation) - catches accidental string concatenation from missing commas
- `PGH` (pygrep-hooks) - flags blanket `# type: ignore` comments

The codebase is already (almost) clean against all three, so this adds no churn - it simply locks in good practices for future contributions.

## Types of Changes

- [ ] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [x] Cosmetics 

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] All tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works 
- [ ] I have added necessary documentation (if appropriate) 

## Further Comments

PR #1646, opened 2026-07-12
Chore: guard type-only imports and tighten mypy

alonfaraj:
## Proposed Changes

This PR continues the ongoing lint/type-checking hardening effort (following #1633, #1638, #1643) with two focused, low-risk improvements plus some incidental typo fixes.

**1. Enable the `TC` (flake8-type-checking) Ruff rule.** This is the natural sequel to the `from __future__ import annotations` work in #1643: it moves imports used *only* in type annotations behind `if TYPE_CHECKING:`, trimming runtime import cost and making the runtime-vs-typing intent of each import explicit. The fixes applied:

- `pika/channel.py` — `collections.abc.Sequence` (annotation-only) → `TYPE_CHECKING`
- `pika/adapters/twisted_connection.py` — `Deferred` (annotation-only) → `TYPE_CHECKING`
- `pika/adapters/blocking_connection.py` — `spec.Basic` → `TYPE_CHECKING`; and `import pika.frame` moved *out* to the runtime imports (TC004: the import also binds the `pika` root name, which is used at runtime)

`TC006` (which quotes the type expression in `cast()` calls) is added to the `ignore` list. Applying it turned cast types into forward-reference strings that then triggered a `UP007`/`UP045` PEP-604 rewrite cascade — a lot of churn for a micro-optimization, and quoted casts break editor "go to definition". The valuable part of the family (guarding imports) is kept; the one debatable rule is disabled with a comment explaining why.

**2. Enable zero-churn strict mypy flags.** After the recent typing pushes (#1638, #1619) the codebase already satisfies several stricter mypy checks, so this locks them in to prevent regressions — no code changes required:

- `warn_redundant_casts` — flags unnecessary `cast()` calls
- `strict_equality` — flags non-overlapping `==`/`!=` comparisons (a real bug class)
- `extra_checks` — additional type-consistency checks

Higher-churn flags (`disallow_incomplete_defs`, `warn_return_any`, etc.) each carry 40–70 existing errors and are intentionally left for a dedicated typing pass.

**3. Minor typo fixes** in comments and docstrings (`dependency`, `receive`, `content`, `heartbeat`, `encoded`).

## Types of Changes

- [ ] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Cosmetics

## Checklist

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,6 +7,9 @@ check_untyped_defs = True
 no_implicit_optional = True
 show_error_codes = True
 warn_unused_ignores = True
+warn_redundant_casts = True
+strict_equality = True
+extra_checks = True
 enable_error_code = explicit-override
 
 [mypy-pika.spec]
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -44,7 +44,7 @@ import pika.spec
 import pika.validators as validators
 from pika._utils import override
 
-# NOTE: import SelectConnection after others to avoid circular depenency
+# NOTE: import SelectConnection after others to avoid circular dependency
 from pika.adapters import select_connection
 from pika.adapters.utils import connection_workflow
 from pika.exchange_type import ExchangeType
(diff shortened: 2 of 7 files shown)


lukebakken (approved):
Verified the moved imports have no runtime uses (including `cast()` sites), and that the new mypy flags, full test suite, and strict docs build all pass locally. The codecov/patch failure is an artifact of `TYPE_CHECKING` blocks never executing. Thank you!

Reviewed with the assistance of Claude Fable 5.

PR #1647, opened 2026-07-12
Remove dead capabilities-stripping code from Connection._set_server_information

alonfaraj:
## Proposed Changes

`Connection._set_server_information` contained dead code intended to remove the `capabilities` entry from `server_properties` after copying it into `server_capabilities`:

```python
if hasattr(self.server_properties, 'capabilities'):
    del self.server_properties['capabilities']
```

`server_properties` is a `dict`, so `hasattr(some_dict, 'capabilities')` checks for an *attribute* named `capabilities`, not a *key* - which is always `False`. The `del` therefore never ran, and since 7cd60baef213f9b11fcd0e2f3c85cb773378f049 (2011) the public `server_properties` attribute has always retained the `capabilities` key.

Because removing the key never actually worked, stripping it now would silently change 15-year-old observable behavior for any user code reading `connection.server_properties['capabilities']`. So this PR resolves the inconsistency by declaring the de facto behavior intentional (see [this review comment](https://github.com/pika/pika/pull/1647#pullrequestreview-4680406909)):

- `server_properties` is left exactly as the broker sent it, `capabilities` included.
- `server_capabilities` is populated with `.get('capabilities', {})` as a convenience view of the same dict.
- The dead `hasattr`/`del` code is removed, and a comment documents why the key deliberately stays.
- The regression test pins the kept-key behavior and asserts `server_capabilities` is the same object as `server_properties['capabilities']`.

Also included: removal of a stale commented-out `get_socket_repr` block and its dangling `# TODO` in `BaseConnection.__repr__`, plus comment/typo cleanups in the affected test.

## Types of Changes

- [ ] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Cosmetics 

## Checklist

## Further Comments

There are 5 more `# TODO` comments in the code base which are not handled in this PR.

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -93,28 +93,6 @@ class BaseConnection(connection.Connection):
 
     @override
     def __repr__(self) -> str:
-
-        # def get_socket_repr(sock):
-        #     """Return socket info suitable for use in repr"""
-        #     if sock is None:
-        #         return None
-        #
-        #     sockname = None
-        #     peername = None
-        #     try:
-        #         sockname = sock.getsockname()
-        #     except pika._utils.SOCKET_ERROR:
-        #         # closed?
-        #         pass
-        #     else:
-        #         try:
-        #             peername = sock.getpeername()
-        #         except pika._utils.SOCKET_ERROR:
-        #             # not connected?
-        #             pass
-        #
-        #     return '%s->%s' % (sockname, peername)
-        # TODO need helpful __repr__ in transports
         return (
             f'<{self.__class__.__name__} {self._STATE_NAMES[self.connection_state]} transport={self._transport} params={self.params}>'
         )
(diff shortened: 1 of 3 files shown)


lukebakken (commented):
Thank you - the diagnosis is exactly right: `hasattr()` on a dict never sees keys, so the `del` has been dead code since 7cd60baef213f9b11fcd0e2f3c85cb773378f049 in 2011.

However, after discussing it, we are going to resolve the bug in the opposite direction. Because the strip never actually worked, `server_properties` (a public attribute) has included `'capabilities'` for 15 years, and user code may legitimately read it from there. Stripping the key now would be a silent behavioral break in a minor release, for no functional gain. So instead we will declare the de facto behavior intentional: leave `server_properties` exactly as the broker sent it, populate `server_capabilities` with `.get()`, and delete the dead `hasattr`/`del` code. Same cleanup, zero user-visible change.

I will push the adjustment (plus a matching test assertion) directly to this branch. The `get_socket_repr` dead-code removal and comment cleanups are unaffected and appreciated.

PR #1649, opened 2026-07-15
Fix: address pyright issues

alonfaraj:
## Proposed Changes

This PR addresses pyright type-checker errors across the `pika` package and test suite. The goal was to make the codebase type-check cleanly without changing runtime behavior — most fixes are annotations, import hygiene, and tightening test mocks so the checker can see them correctly. Where pyright surfaced genuinely dead or incorrect test code, that code was removed or corrected rather than silenced.

**Error count (excluding the auto-generated `pika/spec.py`, which is unchanged at 133 errors):**

|                     | Errors  |
| ------------------- | ------- |
| Before (`main`)     | **878** |
| After (this branch) | **635** |
| **Resolved**        | **243** |

Changes are grouped by the class of error they resolve:

**1. Module / submodule attribute access (`reportAttributeAccessIssue`)**
- Re-export `frame` and `spec` submodules from the `pika` package so `pika.frame` / `pika.spec` resolve statically (they previously worked only via transitive import).
- Reference submodules through their imported names instead of package paths: `unittest.mock.ANY` → imported `ANY`; `pika.exceptions` / `pika.adapters.select_connection` → the bound local names.

**2. `timeout` parameter types**
- Widen `timeout: int` → `timeout: float | None` in `thread_safe_connection`. The value flows to `Thread.join()` / `Event.wait()`, which accept fractional seconds and `None`; several docstrings already documented passing `None`.

**3. Mock false positives (`reportFunctionMemberAccess` / `reportAttributeAccessIssue` on `MethodType`/`FunctionType`)**
- Assert on the actual mock handle rather than through a statically real-typed attribute — via `with ... as m`, decorator-injected mock params, a `setUp` handle (e.g. `self.callbacks`, `self.channel_impl_mock`, `self.conn_impl_mock`), or a local `cast`. Covers `connection_tests`, `channel_tests`, `blocking_channel_tests`, `blocking_connection_tests`, `select_connection_ioloop_tests`, and `twisted_adapter_tests`.

**4. Type narrowing / incorrect test values**
- Narrow `decode_frame(...)` results with `assert isinstance(...)` before accessing subclass-only attributes (`method`/`properties`/`fragment`).
- Instantiate `spec.BasicProperties()` where an instance (not the class) is required by `frame.Header`.

**5. Incompatible overrides (`reportIncompatibleMethodOverride`)**
- Annotate must-override test stubs (`begin`, `connected`, `verify_message`) as `-> None`. Their `self.fail(...)` bodies made pyright infer `NoReturn`, flagging every subclass override — one base annotation cleared 39 errors.
- Trim `SelectConnectionTemplate` so it no longer redeclares inherited properties (`is_closed`/`is_open`/`ioloop`/…) as `None`.

**6. Dead / no-op code removal**
- Remove a dead `# pylint: disable=...` comment (project uses ruff), two `# pyright: ignore` comments that suppressed nothing, and dead mock assignments to nonexistent `Connection` attributes (`_adapter_connect`, `_flush_outbound`).

**7. Test coverage**
- Add property-delegation tests for `BlockingConnection` (filled a `TODO: test properties`).

## Types of Changes

- [x] Bugfix 
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Cosmetics 

## Checklist

## Fixes

- Fixes #

## Further Comments

- `pika/spec.py` is intentionally untouched (auto-generated by `utils/codegen.py`); its 133 errors are excluded from the counts above.

### DIFF
--- a/pika/__init__.py
+++ b/pika/__init__.py
@@ -7,7 +7,7 @@ logging.getLogger(__name__).addHandler(logging.NullHandler())
 
 # ruff: noqa: E402
 
-from pika import adapters
+from pika import adapters, frame, spec
 from pika.adapters import BaseConnection, BlockingConnection, SelectConnection, ThreadSafeConnection
 from pika.adapters.utils.connection_workflow import AMQPConnectionWorkflow
 from pika.connection import ConnectionParameters, SSLOptions, URLParameters
@@ -28,4 +28,6 @@ __all__ = [
     'ThreadSafeConnection',
     'URLParameters',
     'adapters',
+    'frame',
+    'spec',
 ]
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -145,7 +145,8 @@ class ThreadSafeChannel:
             except ValueError:
                 pass
 
-    def _blocking_rpc(self, method_name: str, channel_method, timeout: int,
+    def _blocking_rpc(self, method_name: str, channel_method,
+                      timeout: float | None,
                       *args, **kwargs) -> Any:
         """
         Execute a channel RPC and block until the broker responds.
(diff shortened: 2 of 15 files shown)


lukebakken (approved):
Awesome stuff!

PR #1651, opened 2026-07-17
Bound work queues in `ThreadSafe*` adapters.

### LINKED ISSUE #1600: Bound work queues in `ThreadSafe*` adapters
## Background

`ThreadSafeChannel` and `ThreadSafeConnection` (added in 1.5.0 via #1582) dispatch user callbacks on a `concurrent.futures.ThreadPoolExecutor(max_workers=1)`:

- `ThreadSafeChannel._consumer_work_pool` runs `basic_consume` deliveries, publisher confirms, return-listener notifications, and server-initiated cancel notifications.
- `ThreadSafeConnection._connection_work_pool` runs `Connection.Blocked` / `Connection.Unblocked` notifications.

`ThreadPoolExecutor` does not expose a `maxsize` parameter, so its internal queue is unbounded. Under sustained load with a slow consumer callback (e.g. high-rate auto-ack delivery, blocking I/O inside the callback), the queue grows without bound and the client process eventually OOMs.

## Reference: RabbitMQ Java client

The Java client uses a per-channel `VariableLinkedBlockingQueue<W>` with `MAX_QUEUE_LENGTH = 1000`. Enqueue behavior is configurable via `ConnectionFactory.setWorkPoolTimeout(ms)`:

- Default (`-1`): `queue.put(item)` — blocks forever, applying TCP back-pressure to the broker via the I/O thread.
- Positive value: `queue.offer(item, timeout, ms)`; on timeout, throws `WorkPoolFullException`.

See `WorkPool` and `ConsumerWorkService` in `rabbitmq-java-client/src/main/java/com/rabbitmq/client/impl/`.

## Proposal

Replace each `ThreadPoolExecutor(max_workers=1)` with a dedicated worker thread that consumes from a `queue.Queue(maxsize=N)`:

- Default `maxsize` of 1000 (matches Java).
- Configurable via a constructor parameter on `ThreadSafeConnection` (propagated to each channel's pool and to the connection-level pool).
- On full queue, default to `queue.put(item)` (block); allow an opt-in timeout that raises a new `WorkQueueFullError`.

Affected sites (all currently `submit(...)` against the pool):

- `basic_consume` deliveries
- `confirm_delivery` ack/nack callback
- `add_on_return_callback`
- `add_on_cancel_callback`
- `add_on_connection_blocked_callback` / `add_on_connection_unblocked_callback`

## Considerations

- The blocking enqueue runs on the IOLoop thread. Blocking the IOLoop is exactly the back-pressure mechanism we want — it stops reading frames from the socket, which propagates through TCP windowing back to the broker.
- Heartbeats also run on the IOLoop. While the queue is blocking the IOLoop, heartbeats stop. The broker's heartbeat timeout (default 60s) bounds how long the IOLoop can be blocked before the broker tears the connection down. This is acceptable: a wedged consumer should not be allowed to silently lose messages.
- Shutdown: the worker thread must be joined explicitly (no `executor.shutdown(wait=True)`). The existing `_pool_shutdown` flag pattern continues to work; sentinel item on the queue is the simplest way to signal exit.
- Error handling: `_safe_dispatch` already wraps callbacks; that doesn't change.

## Out of scope (separate issue)

Optional shared executor across channels (closer to Java's `ConsumerWorkService` model) — the cost of a worker thread per channel only matters at very high channel counts.

## Reference

Surfaced during the design audit for #1582 — see [threadsafe-vs-java-comparison.md](https://gist.github.com/lukebakken/1b0aded3214cc4b44bf5e1d63394ba30#file-threadsafe-vs-java-comparison-md) under "Most important reliability gap: unbounded ThreadPoolExecutor queue".

jholaj:
## Proposed Changes
The `ThreadSafe*` adapters dispatch user callbacks on `ThreadPoolExecutor(max_workers=1)` pools. The executor's internal queue is unbounded, so a slow consumer callback under sustained delivery load grows it until the process runs out of memory.

This replaces each executor with a single worker thread reading from a `queue.Queue` bounded to 1000 items (same as the Java client's `WorkPool`). When the queue fills up, enqueueing blocks the IOLoop, which stops reading from the socket and lets TCP push back on the broker. Setting `work_queue_put_timeout` raises the new `WorkQueueFullError` instead of blocking forever. Both knobs are on `ThreadSafeConnection` and propagate to the per-channel pools; `work_queue_maxsize=0` gives the old unbounded behavior.

Shutdown semantics and the `except RuntimeError` call sites are unchanged, and the worker thread still starts lazily, so publish-only channels don't pay for one.

## Types of Changes

- [ ] Bugfix
- [x] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Cosmetics

## Checklist

## Fixes

- Fixes #1600

## Further Comments

Blocking the IOLoop on a full queue is intentional, that's the back-pressure mechanism the issue asks for. It also means heartbeats stop while blocked, so the broker's heartbeat timeout caps how long a wedged consumer can hold the connection. `WorkQueueFullError` deliberately isn't a `RuntimeError` subclass, otherwise the dispatch sites would swallow it as the pool-shutdown signal and silently drop the event.

### DIFF
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -14,7 +14,6 @@ multiple threads call basic_publish() on a shared connection directly
 
 from __future__ import annotations
 
-import concurrent.futures
 import itertools
 import logging
 import queue
@@ -160,12 +159,17 @@ class ThreadSafeChannel:
     blocking channel methods (which would deadlock).
     """
 
-    def __init__(self, channel, wrapper) -> None:
+    def __init__(self,
+                 channel,
+                 wrapper,
+                 work_queue_maxsize: int = DEFAULT_WORK_QUEUE_MAXSIZE,
+                 work_queue_put_timeout: float | None = None) -> None:
         self._channel = channel
         self._wrapper = wrapper
-        self._consumer_work_pool = concurrent.futures.ThreadPoolExecutor(
-            max_workers=1,
-            thread_name_prefix='pika-consumer',
+        self._consumer_work_pool = _BoundedWorkPool(
+            maxsize=work_queue_maxsize,
+            put_timeout=work_queue_put_timeout,
+            thread_name='pika-consumer',
         )
         self._pool_shutdown = False
         self._next_publish_seq_no: int | None = None
@@ -188,10 +192,9 @@ class ThreadSafeChannel:
         Execute *callback* on the pool worker, logging any exception.
 
         Used to wrap user callbacks before submitting them to a
-        :class:`concurrent.futures.ThreadPoolExecutor` so that exceptions
-        are not silently lost in the unobserved ``Future`` and so that one
-        failing dispatch does not prevent subsequent dispatches from being
-        processed.
+        :class:`_BoundedWorkPool` so that exceptions are not silently lost
+        and so that one failing dispatch does not prevent subsequent
+        dispatches from being processed.
 
         :param label: Human-readable name of the callback for log lines.
         :param callback: The user callback.
@@ -1146,6 +1149,20 @@ class ThreadSafeConnection:
         Pass ``None`` to wait indefinitely (relying on the socket timeout
         in *parameters*).  On timeout the IOLoop is stopped and
         :class:`TimeoutError` is raised.
+    :param work_queue_maxsize: Maximum number of user-callback dispatches
+        (deliveries, publisher confirms, returned messages, cancel and
+        blocked/unblocked notifications) that may be pending on each
+        worker's queue.  Defaults to :data:`DEFAULT_WORK_QUEUE_MAXSIZE`
+        (1000, matching the RabbitMQ Java client).  While a queue is full
+        the IOLoop blocks, applying back-pressure to the broker via TCP.
+        Heartbeats are not sent while the IOLoop is blocked, so the
+        broker's heartbeat timeout bounds how long a wedged consumer can
+        stall the connection.  Pass ``0`` for an unbounded queue.
+    :param work_queue_put_timeout: Seconds to wait for space on a full
+        work queue.  Defaults to ``None`` (wait indefinitely).  On timeout
+        :class:`pika.exceptions.WorkQueueFullError` is raised on the
+        IOLoop thread, closing the connection rather than silently
+        dropping the event.
     :raises Exception: if the connection cannot be established.
     :raises TimeoutError: if *timeout* expires before the connection opens.
     """
@@ -1156,9 +1173,13 @@ class ThreadSafeConnection:
                  parameters,
                  on_open_error_callback=None,
                  on_close_callback=None,
-                 timeout: float | None = DEFAULT_RPC_TIMEOUT) -> None:
+                 timeout: float | None = DEFAULT_RPC_TIMEOUT,
+                 work_queue_maxsize: int = DEFAULT_WORK_QUEUE_MAXSIZE,
+                 work_queue_put_timeout: float | None = None) -> None:
         self._user_on_open_error_callback = on_open_error_callback
         self._user_on_close_callback = on_close_callback
+        self._work_queue_maxsize = work_queue_maxsize
+        self._work_queue_put_timeout = work_queue_put_timeout
 
         self._connect_error = None
         self._connected_event = threading.Event()
@@ -1174,9 +1195,10 @@ class ThreadSafeConnection:
         # Single-worker pool for connection-level event callbacks
         # (Connection.Blocked / Unblocked).  Keeps user code off the
         # IOLoop thread so a slow listener cannot stall heartbeats.
-        self._connection_work_pool = concurrent.futures.ThreadPoolExecutor(
-            max_workers=1,
-            thread_name_prefix=f'pika-conn-{self._instance_id}',
+        self._connection_work_pool = _BoundedWorkPool(
+            maxsize=work_queue_maxsize,
+            put_timeout=work_queue_put_timeout,
+            thread_name=f'pika-conn-{self._instance_id}',
         )
         self._connection_pool_shutdown = False
 
@@ -1348,7 +1370,11 @@ class ThreadSafeConnection:
         with self._channel_waiters_lock:
             if self._closed_reason is not None:
                 raise self._closed_reason
-            ch = ThreadSafeChannel(result[0], self)
+            ch = ThreadSafeChannel(
+                result[0],
+                self,
+                work_queue_maxsize=self._work_queue_maxsize,
+                work_queue_put_timeout=self._work_queue_put_timeout)
             self._channels.append(ch)
         return ch
 
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -35,6 +35,43 @@ DEFAULT_RPC_TIMEOUT = 10
 # Matches MAX_QUEUE_LENGTH in the RabbitMQ Java client's WorkPool.
 DEFAULT_WORK_QUEUE_MAXSIZE = 1000
 
+# Default seconds the IOLoop thread waits for space on a full work queue
+# before raising WorkQueueFullError and tearing down the connection.
+# While the queue is full the IOLoop blocks - it sends no heartbeats and
+# reads no frames - so a wedged consumer stalls the connection for up to
+# this long.  Keep an override comfortably under the negotiated
+# heartbeat-death window (twice the heartbeat interval) so the stall
+# surfaces as a clean WorkQueueFullError before the broker drops the link
+# with an opaque StreamLostError.
+DEFAULT_WORK_QUEUE_PUT_TIMEOUT = 30.0
+
+
+def _validate_put_timeout(put_timeout: float) -> float:
+    """
+    Validate a public work-queue put timeout.
+
+    A ``None`` (infinite) timeout is rejected because it lets a full work queue block the IOLoop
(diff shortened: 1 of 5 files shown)


lukebakken:
@jholaj I pushed two commits addressing the shutdown-hang finding (Finding 2) on top of your branch. Could you review them when you have a chance?

## What changed

- `ee67efb7` - `_BoundedWorkPool.shutdown()` no longer enqueues an in-band sentinel via an untimed `put()`. It sets a shutdown flag the worker checks on a `get()` timeout, so shutdown can never block on a full queue. The drain-all contract is preserved.
- `1f709352` - `shutdown()` gained an optional `timeout` that bounds the worker join and reports whether the worker exited. That budget is threaded through the channel and connection close paths so `close()` / `abort()` honor their timeout even when a delivery callback is wedged. A worker still running when the budget expires is left to finish on its own, since it is a daemon thread and Python cannot forcibly stop a thread. The clean-shutdown courtesy drain on the idle IOLoop thread stays unbounded, so a healthy close still drains every queued delivery.

The rationale we settled on: a hang during shutdown is worse than a dropped delivery during shutdown, and the drop only happens when the caller's own timeout has already expired.

Three unit tests were added covering the non-blocking shutdown on a full queue, the bounded-join return value, and `close()` returning within its timeout against a wedged consumer.

lukebakken:
I converted to draft because I still have work to do on this PR.

lukebakken:
@jholaj OK, all set for you to take a look!

pika/adapters/thread_safe_connection.py
```diff
@@ -1179,18 +1451,46 @@ def _on_connection_closed(self, _connection, reason) -> None:
...
+            (the clean-shutdown default). A finite budget is split across channels so the whole
+            sweep stays bounded.
+        """
         with self._channel_waiters_lock:
             channels = list(self._channels)
+        if not channels:
+            return
+        if timeout is None:
+            for ch in channels:
+                ch._shutdown_pool()
```
jholaj, line 1513: Small perf catch with the polling design: each idle worker only notices the flag when its `get(timeout=0.1)` expires + this loop joins channels one at a time - so the waits add up instead of overlapping. Each join waits somewhere between 0 and 0.1s (wherever that worker happens to be in its poll window), so:
> ```
> total ≈ N x ~0.05-0.1s
> 20 channels -> 1.2 s measured (used to be instant with the sentinel)
> 100 channels -> up to 10s, i.e. the entire default close() budget
> ```

And since this runs on the IOLoop tail, which `close()` joins with that 10s budget, at circa 100+ consumer channels a completely healthy close would blow the budget and take the force path for no reason.

Cheap fix is: set all the flags first, then join, so the workers wake concurrently:

> ```python
> for ch in channels:
>     ch._consumer_work_pool.shutdown(wait=False)   # flag only, never blocks
> for ch in channels:
>     ch._shutdown_pool(...)
> ```

0.1s flat with the same 20 channels. One catch though: the first loop has to go straight to the pools, because `_shutdown_pool()` would set its `_pool_shutdown` flag on the first pass and skip the join on the second.

pika/adapters/thread_safe_connection.py
```diff
@@ -14,23 +14,227 @@
...
+            ``False`` if *timeout* expired with the worker still alive.
+        """
+        with self._lock:
+            self._shutdown = True
+            thread = self._thread
+        if thread is None:
+            return True
+        if not wait:
+            return not thread.is_alive()
+        thread.join(timeout=timeout)
```
jholaj, line 264: This blows up when `close()` is called from inside a consumer callback - we end up joining the worker's own thread and Python doesn't want to:

> ```python
> def on_message(ch, method, properties, body):
>     ch.basic_ack(method.delivery_tag)
>     ch.close()   # RuntimeError: cannot join current thread
> ```

Docstrings says this is safe to call from any thread. The old way had the same problem, so nothing these commits broke - but since we're touching these lines anyway:

> ```suggestion
>         if thread is threading.current_thread():
>             # A delivery callback is closing its own channel. The worker cannot join itself.
>             # It exits once the callback returns.
>             return False
>         thread.join(timeout=timeout)
> ```

The flag is already set at this point, so the worker just exits on its own once the callback returns.

pika/adapters/thread_safe_connection.py
```diff
@@ -14,23 +14,227 @@
...
+            self._queue.put((fn, args), timeout=self._put_timeout)
+        except queue.Full:
+            raise WorkQueueFullError(
+                f'work queue full for {self._put_timeout} seconds '
+                f'(maxsize={self._queue.maxsize})') from None
+
+    def _run_worker(self) -> None:
+        while True:
+            try:
+                item = self._queue.get(timeout=self._SHUTDOWN_POLL_INTERVAL)
```
jholaj: Hmmm, If I'm not mistaken, I think the polling could go away completely with a flag + best effort sentinel.

Keep the flag as the source of truth and use the sentinel just as a wakeup ping - `put_nowait`, dropped if the queue is full. It can't miss: a worker blocked in `get()` means the queue is empty, so the ping lands and a full queue means the worker is busy and re-checks the flag between items.

Something like:
```python
def _run_worker(self):
    while True:
        if self._shutdown and self._queue.empty():
            return
        item = self._queue.get()          # plain blocking get, no timeout
        if item is self._SENTINEL:
            continue                      # wakeup ping, re-check at loop top
        fn, args = item
        fn(*args)

def shutdown(self, ...):
    # ...set flag under lock...
    try:
        self._queue.put_nowait(self._SENTINEL)
    except queue.Full:
        pass   # worker is busy, it re-checks between items
 
```

That would kill the idle wakeups and the poll-tick delay in one go - and the sequential joins in `_shutdown_all_consumer_pools()` along with them. What do you think?

jholaj (commented):
LGTM overall, thanks for seeing this one through. 

Flagged a couple of things inline, would love your opinion on them. 
Thanks!

lukebakken:
Thanks for the review, @jholaj - all three are good catches. Pushed fa1685b1 addressing them.

## Sentinel instead of polling (your line 189 comment)

Took your suggestion as-is. `_run_worker` now does a plain blocking `get()`, and `shutdown()` enqueues a best-effort `put_nowait(_SENTINEL)` wakeup that a full queue simply drops (a busy worker re-checks the flag between items). The worker checks `self._shutdown and self._queue.empty()` at the loop top so it still drains every queued item before exiting. This removes both the idle wakeups and the poll-tick latency, exactly as you described.

## Sequential joins in `_shutdown_all_consumer_pools` (your line 1469 comment)

I did not add a separate change here, because the sentinel rework fixes it at the source - which you called out yourself ("would kill ... the sequential joins ... along with them"). With instant wakeups each `thread.join()` in that sweep returns immediately, so the N x 0.1s accumulation is gone. Adding the two-loop flag-then-join restructure on top would be redundant complexity for a problem that no longer exists. Happy to revisit if you would still prefer the explicit two-loop form for clarity.

## `close()` from inside a consumer callback (your line 235 comment)

Fixed, with one tweak. `shutdown()` now returns early when `thread is threading.current_thread()`, so a delivery callback that closes its own channel no longer hits `RuntimeError: cannot join current thread`. I return `True` rather than `False` there: the worker will drain and exit on its own once the callback returns, and returning `False` would make `_shutdown_pool` log the "did not drain ... abandoning its worker thread" warning, which is misleading for a deliberate self-close.

## Verification

Full unit suite (900 passed), ruff, yapf, docformatter, and mypy all clean. Acceptance suites against a live broker pass: the bounded-queue suite (3 tests) and the existing thread_safe suite (7 tests). Added `test_shutdown_from_worker_thread_does_not_join_itself` for the self-join guard; the full-queue drop path is already covered by `test_shutdown_does_not_block_when_queue_full`.

pika/adapters/thread_safe_connection.py
jholaj, line 240: nit: `True` is the right value here for the warning, but the `:returns:` docstring defines it as "the worker has exited" and at this point it hasn't yet (it's the thread we're running on - it exits after the callback returns). Same for `wait=False` from the worker, where any other thread would get `False`. Just worth a sentence so nobody reads too much into it later.

jholaj (commented):
The two-loop version has nothing left to fix, I'd leave it as-is.

One docstring nit inline, otherwise LGTM, thank you.

pika/adapters/thread_safe_connection.py
lukebakken: Good nit, thanks. Updated the `:returns:` docstring to spell out the self-call case: when `shutdown()` is called from the worker thread itself it returns `True` without joining even though the worker has not exited yet (it exits once the calling callback returns), so nobody reads the return value as "the thread is already gone."

lukebakken (approved):
## Approving

Bounding the `ThreadSafe*` work queues is the right fix for #1600, and the branch has come together well. The core change replaces the unbounded executor with a single-worker bounded `queue.Queue` pool, surfaces overflow to applications as a real `WorkQueueFullError` (catchable by type, and now a subclass of `AMQPConnectionError` so base-class connection-loss handlers see it too), and bounds every shutdown join so `close(timeout=...)` is honored even against a wedged consumer.

The follow-up hardening pass covered the remaining edge cases: a submit/shutdown race that could strand work in a drained queue, up-front signaling of every channel pool before joining any so a slow channel cannot burn the whole close budget, and atomic claim of the shutdown flag so concurrent `close()` calls do not both run the join. Test coverage is solid, including acceptance tests that exercise the overflow-closes-connection and close-honors-timeout paths against a live broker.

CI is fully green across the matrix (all OS/Python/TLS legs, mypy, lint, yapf, CodeQL, docs).

Nice work, @jholaj.

PR #1652, opened 2026-07-19
Fix: pyright errors

alonfaraj:
## Proposed Changes

Follow-up to #1649, continuing the effort to make Pika type-check cleanly
without changing runtime behavior. This round covers both the library
(`pika/`, 26 files) and the test suites (28 files).

**Error count (excluding the auto-generated `pika/spec.py`, which is unchanged
at 133 errors):**

|                     | Errors  |
| ------------------- | ------- |
| Before (main)      | **635** |
| After (this branch) | **276** |
| **Resolved**        | **359** |

Changes are grouped by the class of error they resolve:

**1. Annotations too narrow for the actual values (`reportArgumentType` / `reportAttributeAccessIssue`)**
- Widen `Connection.known_hosts` to `str | bytes | None` (matching
  `decode_short_string`'s return) and drop the pyright ignore it needed.
- Widen `TwistedChannel.callback_deferred(replies)` from `list[type[Method]]`
  to `Sequence[...]` — list invariance rejected valid calls like
  `[spec.Basic.CancelOk]`; the value is only forwarded to
  `Channel.add_callback`, which already takes `Sequence`.
- Replace the remaining `# type:` declaration comments with real annotations
  (`connection_class: type[pika.adapters.BaseConnection]`, etc.).

**2. Optional narrowing (`reportOptionalMemberAccess` / `reportOptionalSubscript`)**
- Add `assert ... is not None` (or `assert isinstance(...)`) before accesses
  where the invariant already holds: `basic_get`/`consume` result tuples,
  `_closing_reason`, `_timeout_heap`, `_queue_consumer_generator`,
  `ready`/`closed` Deferreds, `self.connection` in async adapter tests, and
  friends. Existing `assertIsInstance` guards were converted to
  `assert isinstance` so the checker narrows on them too (same runtime failure
  semantics).

**3. Incorrect test values that only worked by accident (`reportArgumentType`)**
- `str` bodies passed to `bytes` parameters: `basic_publish` (57 call sites in
  the blocking acceptance tests alone), `_on_getok`, `_on_return`,
  `_set_content`, `dataReceived`, `_adapter_emit_data`.
- The `spec.Basic.GetEmpty` **class** passed where an instance is required;
  the numeric channel number stuffed into `Channel.OpenOk`'s deprecated string
  `channel_id` field; `int` passed to `bool` `redelivered`; string delivery
  tags passed to `int` parameters.
- String/None placeholders passed to callbacks typed `Connection` /
  `Channel` / `Exception` — now `mock.Mock()` for ignored parameters and real
  exception instances (with pass-through assertions preserved) where the value
  is used.

**4. Incompatible overrides (`reportIncompatibleMethodOverride`)**
- Align the `start()` parameter name between `AsyncTestCase` /
  `BoundQueueTestCase` and the `AsyncAdapters` mixin (`adapter` →
  `adapter_class`) — one rename cleared the error on every async adapter test
  class.

**5. Latent test bugs surfaced by the checker (`reportPossiblyUnboundVariable`)**
- Context-manager targets (`channel`/`connection`) were bound *inside*
  `assertRaises` blocks; an exception raised before the `as` binding would be
  swallowed and turn the post-block assertion into a `NameError`. The objects
  are now bound before the block (identical semantics — `__enter__` returns
  `self`).

## Types of Changes

- [x] Bugfix 
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Cosmetics 

## Checklist

## Fixes

- Fixes #

### DIFF
--- a/examples/blocking_delivery_confirmations.py
+++ b/examples/blocking_delivery_confirmations.py
@@ -19,7 +19,7 @@ channel.confirm_delivery()
 try:
     channel.basic_publish(exchange='test',
                           routing_key='test',
-                          body='Hello World!',
+                          body=b'Hello World!',
                           properties=pika.BasicProperties(
                               content_type='text/plain',
                               delivery_mode=pika.DeliveryMode.Transient))
--- a/examples/blocking_publish_mandatory.py
+++ b/examples/blocking_publish_mandatory.py
@@ -20,7 +20,7 @@ channel.confirm_delivery()
 try:
     channel.basic_publish(exchange='test',
                           routing_key='test',
-                          body='Hello World!',
+                          body=b'Hello World!',
                           properties=pika.BasicProperties(
                               content_type='text/plain',
                               delivery_mode=pika.DeliveryMode.Transient),
--- a/examples/comparing_publishing_sync.py
+++ b/examples/comparing_publishing_sync.py
@@ -7,7 +7,7 @@ connection = pika.BlockingConnection(parameters)
 channel = connection.channel()
 
 channel.basic_publish(
-    'test_exchange', 'test_routing_key', 'message body value',
+    'test_exchange', 'test_routing_key', b'message body value',
     pika.BasicProperties(content_type='text/plain',
                          delivery_mode=pika.DeliveryMode.Transient))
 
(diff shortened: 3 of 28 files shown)


lukebakken (approved):
Approving after rebasing onto main and resolving the conflicts with the recently merged frame-buffer work (#1653).

Conflict resolution: kept the `bytearray` frame buffer and `_processing_frame_buffer` flag from #1653 while applying this PR's `known_hosts: str | bytes | None` widening and the removed pyright suppression. Fixed `test_on_data_available` to use a `bytearray` buffer, since #1653 changed `_frame_buffer`'s type out from under the test. Full CI is green on the rebased branch: all test-modern and test-legacy jobs, mypy on all three platforms, ruff, docformatter, and CodeQL.

PR #1655, opened 2026-07-22
Optimization: Speed up field-table codec

alonfaraj:
## Proposed Changes

Speed up the AMQP field-table codec in `pika/data.py`, plus two tiny allocation
fixes on the marshal path. No API or behavior changes — same values decoded,
same bytes produced, same exceptions raised.

**What changed:**

- `decode_value` previously allocated a 1-byte `bytes` slice for every field's
  type tag (`encoded[offset:offset + 1]`) and compared it through a ~20-branch
  `elif` chain of bytes literals. It now dispatches on the raw integer byte
  (`encoded[offset]`, no allocation per field) against module-level int
  constants (`_FIELD_LONG_STRING = ord('S')`, …), with the most common field
  types checked first.
- Single-octet reads (`decode_short_string` length byte, and the `t`/`B`/`D`
  branches of `decode_value`) index the buffer directly instead of calling
  `struct.unpack_from('B', ...)`.
- `encode_short_string` packs the length with `bytes((length,))` instead of
  `struct.pack('B', length)`.
- `Frame._marshal` reuses a hoisted `_FRAME_END_BYTE` module constant instead
  of constructing `bytes((spec.FRAME_END,))` on every marshaled frame.

**Why:** header-table decoding runs for every consumed message that carries
headers (tracing headers, retry/dead-letter metadata, `x-` arguments), and the
per-field slice allocation plus bytes-compare chain dominates its cost.

**Benchmarks** (interleaved in-process A/B, original vs patched alternating,
6 rounds each, median; decoded values asserted identical to the original
implementation):

| workload | original | patched | delta |
|---------:|---------:|--------:|------:|
| header-table decode (8 mixed-type fields) | 219,536 tables/s | 310,981 tables/s | **+41.7%** |
| publish-side marshal (`Basic.Deliver` + `Header`) | 470,010 msg/s | 478,982 msg/s | +1.9% |
| e2e consume, 100 B messages, no headers | 168,221 msg/s | 171,037 msg/s | +1.7% |

As expected, the win is concentrated in workloads that carry header tables;
header-less traffic sees only a small improvement.

<details>
<summary><b>Minimal repro for the table-decode number</b></summary>

```python
# bench_table.py — run on main, then on this branch
import timeit
import decimal
from datetime import datetime, timezone

import pika.data as data

table = {
    'x-string': 'hello world, this is a header value',
    'x-int': 42,
    'x-long': 9123456789,
    'x-bool': True,
    'x-decimal': decimal.Decimal('3.14159'),
    'x-timestamp': datetime.fromtimestamp(1750000000, timezone.utc),
    'x-array': [1, 2, 3, 'four'],
    'x-nested': {'a': 1, 'b': 'two'},
}
pieces = []
data.encode_table(pieces, table)
blob = b''.join(pieces)

n = 50_000
best = min(timeit.repeat(lambda: data.decode_table(blob, 0), number=n, repeat=7))
print(f'{n / best:,.0f} tables/s')
```

```bash
git checkout main && python bench_table.py                       # ~220k tables/s
git checkout perf/data-codec-micro-opts && python bench_table.py # ~305k tables/s
```

</details>

## Types of Changes

- [ ] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

## Checklist

## Fixes

None — standalone performance improvement.

## Further Comments

- The unknown-tag error path still raises
  `InvalidFieldTypeException(bytes((kind,)))`, so error messages are unchanged.
- `decode_value` gained an explicit `value: Any` annotation because the
  branch reordering changed which assignment mypy sees first.
- The `bytes((spec.FRAME_END,))` comparison on the *decode* side
  (`frame.decode_frame`) is intentionally untouched:
 #1653 
  replaces that line with an int comparison, and leaving it alone keeps these
  two PRs conflict-free in either merge order.

### DIFF
--- a/pika/data.py
+++ b/pika/data.py
@@ -11,6 +11,29 @@ from typing import Any
 from pika import exceptions
 from pika._utils import as_bytes
 
+# AMQP field-table value type tags, as integer byte values. decode_value
+# dispatches on the raw byte rather than a 1-byte slice to avoid a bytes
+# allocation per decoded field.
+_FIELD_BOOL = ord('t')
+_FIELD_SHORT_SHORT_INT = ord('b')
+_FIELD_SHORT_SHORT_UINT = ord('B')
+_FIELD_SHORT_INT = ord('U')
+_FIELD_SHORT_UINT = ord('u')
+_FIELD_LONG_INT = ord('I')
+_FIELD_LONG_UINT = ord('i')
+_FIELD_LONG_LONG_INT = ord('L')
+_FIELD_LONG_LONG_INT_ALT = ord('l')
+_FIELD_FLOAT = ord('f')
+_FIELD_DOUBLE = ord('d')
+_FIELD_DECIMAL = ord('D')
+_FIELD_SHORT_INT_ALT = ord('s')
+_FIELD_LONG_STRING = ord('S')
+_FIELD_BYTE_ARRAY = ord('x')
+_FIELD_ARRAY = ord('A')
+_FIELD_TIMESTAMP = ord('T')
+_FIELD_TABLE = ord('F')
+_FIELD_VOID = ord('V')
+
 
 def encode_short_string(pieces: list[bytes], value: str) -> int:
     """
@@ -37,7 +60,7 @@ def encode_short_string(pieces: list[bytes], value: str) -> int:
     if length > 255:
         raise exceptions.ShortStringTooLong(encoded_value)
 
-    pieces.append(struct.pack('B', length))
+    pieces.append(bytes((length,)))
     pieces.append(encoded_value)
     return 1 + length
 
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
Reviewed at medium effort. LGTM.

I verified behavior preservation independently rather than relying on the description: all 19 type-tag constants map to the same struct format, offset advance, and semantics as the branches they replace (nothing swapped, including the `L`/`l` and `U`/`s` pairs). The single-byte direct-index reads (`bool(encoded[offset])`, `encoded[offset]`, `bytes((length,))`) are value-equivalent to the old `struct` calls, and the unknown-tag error still raises `InvalidFieldTypeException` with the same bytes payload. The one behavioral nuance, a truncated buffer now raising `IndexError` instead of `struct.error` on single-byte reads, is unreachable through the normal decode path: `decode_table` only runs after `decode_frame` has buffered the complete frame and validated the FRAME_END marker, and neither exception is caught around it. The `_FRAME_END_BYTE` hoist is a clean, correct micro-win. `hatch` typecheck, lint, fmt-check, and docfmt-check all pass, and the data and frame unit tests are green.

PR #1662, opened 2026-07-26
Add Ruff quote style (`Q`)

alonfaraj:
## Proposed Changes

`CLAUDE.md` documents a single-quote string convention, but nothing enforced it and the codebase had drifted. This adds the linter rule that enforces it and applies the mechanical conversion.

Worth calling out why `[tool.ruff.format] quote-style = "single"` was not already doing this job: it configures `ruff format`, which this project does not use (yapf is the formatter), so it was inert. The `Q` rules are linter rules and read `[tool.ruff.lint.flake8-quotes] inline-quotes`, which defaults to `double`. Enabling `Q` without that section would have had ruff rewriting single quotes into double ones, the opposite of the intended convention.

Two commits, kept separate so the mechanical churn is easy to skip while reviewing: one adds `Q` to the rule selection and sets `inline-quotes = "single"`, the other is pure `ruff check --fix` output.

Docstrings are untouched, since `docstring-quotes` and `multiline-quotes` keep their `double` defaults, so PEP 257 style is preserved and `docformatter` is unaffected. Strings containing a single quote keep their double quotes via the `avoid-escape` default, matching the carve-out in `CLAUDE.md`.

## Types of Changes

- [ ] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [x] Cosmetics 

## Checklist

## Further Comments

### DIFF
--- a/examples/async_as_callback.py
+++ b/examples/async_as_callback.py
@@ -36,7 +36,7 @@ class Headers(BaseModel):
     priority: Priority
     task_type: str | None = None
 
-    @validator("priority", pre=True)
+    @validator('priority', pre=True)
     def _convert_priority(self, value):
         return Priority[value]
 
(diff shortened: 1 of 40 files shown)


lukebakken (approved):
Thanks! I do love some `'` characters!

PR #1664, opened 2026-07-27
Upload coverage to Codecov once after all tests finish

alonfaraj:
## Problem

The `codecov/patch` status and the Codecov PR comment appear long before CI has finished, computed from a fraction of the test matrix, and then change as the rest of the run completes.

Two causes:

1. **Every matrix job uploaded separately.** Codecov notifies as soon as the first upload is processed, then re-notifies on each subsequent one. The matrix produces 48 uploads (`test-modern` 5 Python x 3 OS x 2 TLS, plus `test-legacy` 3 x 3 x 2), so the first report to land triggered the status.
2. **`wait_for_ci: true` was inert.** It gates on the commit's CI statuses, but on a PR branch every workflow here fires only via `pull_request`, so the check runs attach to the merge commit. `codecov-action` attributes coverage to the PR head commit. Codecov polls the head commit, finds no pending checks, and concludes CI is done.

## Changes

Each matrix job now publishes `coverage.xml` as an artifact rather than uploading it. A new `coverage` job runs after both test workflows, downloads all reports, and makes one Codecov upload. One upload, one notification, sent only after every test job has finished.

`after_n_builds` was the other candidate fix, but it would have made the notification depend on 48 independent jobs, any one of which could die before its upload and strand the counter permanently. Gating on the job graph is deterministic.

Also in this PR:

- A guard step between download and upload fails the job if no reports were collected. With a single upload point and `fail_ci_if_error: false`, a bad artifact pattern would otherwise pass green and silent.
- A concurrency group on `main.yaml`. There wasn't one, so three pushes in quick succession ran three complete 48-job matrices at once. `cancel-in-progress` is an expression rather than a bare `true` so that superseded PR runs are cancelled while pushes to `main` queue, keeping coverage for every commit on the default branch.
- `retention-days: 7` on the artifacts, so re-running a failed `coverage` job the next day still finds its inputs. Reports are ~225K each and public repos are not billed for artifact storage.
- The `codecov` dependency is dropped from the default hatch environment. It is the deprecated Python uploader and nothing invokes it; uploads go through `codecov/codecov-action`, which fetches its own.

## Notes for reviewers

**Flags are gone.** A single upload carries one flag set, so the per-OS and per-Python breakdown is no longer available and the existing flags on the Codecov dashboard will go stale. Coverage is now a union across platforms, which means a line exercised on Linux but never on Windows reads as covered. Restoring flags is possible by fanning the `coverage` job across an OS matrix and setting `after_n_builds` to match, at the cost of reintroducing a multi-way counter. Happy to do that in a follow-up if the platform split is worth keeping.

**`coverage` is deliberately not in `tests-passed`.** Codecov upload flakiness should not block merges, and the guard step surfaces the silent-empty case in the job's own log.

**The flag table in the Codecov comment is a transitional artifact.** Every row shows `?`, meaning no data for that flag on this branch. The comparison base is a commit on `main`, which still carries all 48 flags from the old per-job uploads. It confirms the change worked, and it also confirms carryforward is off, so stale flag data is not leaking into the merged coverage. The table goes away once `main` itself carries a flagless report.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

## Fixes

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,2 @@
 [run]
 omit = pika/spec.py
-relative_files = true
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,3 @@
 [run]
 omit = pika/spec.py
+relative_files = true
--- a/.github/workflows/_test.yaml
+++ b/.github/workflows/_test.yaml
@@ -20,9 +20,6 @@ on:
         required: false
         type: boolean
         default: false
-    secrets:
-      CODECOV_TOKEN:
-        required: false
 
 jobs:
   build-windows:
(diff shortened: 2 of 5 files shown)


lukebakken (approved):
Thank you!

PR #1666, opened 2026-07-29
ruff: extend the default rule set and fix the fallout

### LINKED ISSUE #1663: `ruff`: select vs extend-select approach
## Summary

Recent ruff versions ship a [much larger default rule set](https://astral.sh/blog/ruff-v0.16.0#better-default-rule-set) than they used to (59->413).
 Our `[tool.ruff.lint] select` list was written when ruff's default was a handful of pycodestyle and pyflakes rules, so the list read as an expansion of the defaults. That is no longer true: today it is a *narrowing* of them.
I think we should make a decision on which direction we want to go before this drifts further.

## What changed

Ruff's current default now covers roughly 37 rule prefixes. Our list has 21 entries. 
Cross-referencing the two, almost every prefix we select is already enabled by default, and only one entry in the whole list adds anything ruff would not have enabled on its own. 
The practical effect of using `select` is therefore to switch *off* around 18 prefixes that ruff now turns on out of the box.

So the config is doing close to the opposite of what its comments imply. Someone reading it reasonably concludes we have opted into a broad rule set, when in fact we have opted out of most of ruff's current recommendations.

## The two options

Switching `select` to `extend-select` surfaces roughly a hundred violations, spread across exception handling, logging practice, timezone-naive datetime calls, control-flow simplification, type-stub conventions, and a handful of test and example specific patterns. Essentially none of them are autofixable, so this is real review work rather than a mechanical pass.

**Option A: keep `select`.** CI stays green and the enabled set is explicit and reviewable. The cost is that we are pinned to a hand-maintained subset that silently diverges from ruff's defaults every release. Nobody is going to audit the full default set on each ruff release to keep our list current, which is how we ended up here in the first place.

**Option B: switch to `extend-select`.** We inherit ruff's evolving defaults and fix violations as they appear. 
Optionally - disable too-noisy rules, undesired etc. (like we already have done).
This is what `CLAUDE.md` already says our policy is:

> Ruff is intentionally left unpinned so CI always installs the latest release. When ruff stabilizes a new rule, CI surfaces the violations immediately and we fix them, keeping the codebase current with ruff's evolving idioms.

The cost is the initial backlog, plus the ongoing reality that unpinned ruff combined with inherited defaults means a ruff release can turn CI red on code nobody touched.

Personally, I would stick with B and fix issues as they appear..

Any thoughts on this?

alonfaraj:
## Proposed Changes

`[tool.ruff.lint]` used `select`, which **replaces** ruff's default rule set with exactly the listed prefixes. This PR switches it to `extend-select`, so those prefixes now **augment** the defaults instead of standing in for them. The enabled rule count goes from 339 to 527.

This follows the policy in `AGENTS.md`: ruff is deliberately left unpinned so CI surfaces newly stabilized rules and we fix them, keeping the codebase current with ruff's evolving idioms. `select` was quietly opting out of everything ruff considers default, so that mechanism was only working for the 21 prefixes we happened to list.

Running the new config against `main` reports **107 violations**. This branch clears all of them: 96 fixed in code, 11 suppressed at the line with a documented reason. Two additional rules are ignored in config rather than fixed (see below).

No behaviour change is intended anywhere, with two deliberate exceptions noted under Further Comments.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [x] Breaking change 
- [x] Documentation 
- [x] Cosmetics 

Breaking change is ticked for one narrow reason: `Channel.confirm_delivery()` now raises `TypeError` instead of `ValueError` for a non-callable `ack_nack_callback`. Details under Further Comments.

## Checklist

No new tests: every change is either a lint-driven refactor with no new behaviour, or covered by an existing test. One existing test was updated for the exception-type change (`test_confirm_delivery_with_bad_callback_raises_value_error` becomes `..._raises_type_error`). Docstrings were corrected where an annotation or `:raises` clause was wrong.

## Fixes

- Fixes #1663

### Error groups fixed

**Exception types** (`TRY004` x3, `TRY002` x2). `TRY004` wants `TypeError` for failed type checks. Two of the three sites were genuine inconsistencies: `Channel.confirm_delivery()` raised `ValueError` for a non-callable callback while `validators.require_callback`, `validators.rpc_completion_callback` and `check_callback_arg` all raise `TypeError` for the same mistake, and `confirm_delivery` itself calls `rpc_completion_callback` on the next line. `TRY002` replaced two bare `raise Exception(...)` calls in an acceptance test with a named `_ConnectionRejectedError`.

**Exception logging** (`TRY401` x17, `S110` x3, `LOG015` x6). `logging.exception()` already appends the active exception's traceback, so passing the exception into the message duplicated it in every record; `TRY401` removed the redundant copy. `S110` flagged three `try`/`except`/`pass` blocks, all genuine best-effort paths (teardown helpers, a stderr dump guard) where swallowing is correct but silence was not, so each now logs. `LOG015` replaced root-logger calls with module loggers.

**Type annotations** (`PYI036` x8, `PYI034` x4, `PYI041` x2). `PYI036` caught real bugs, not style: both `__exit__` methods were annotated `exc_type: Exception` (it is the exception *class*) and `value: TracebackException` (it is the exception *instance*, not a `traceback.TracebackException`), and all three arguments omitted `| None` even though they are `None` on normal exit. A type checker following the old signatures accepted wrong calls and rejected correct ones. `PYI034` switched four `__enter__` methods to `Self`.

**Test assertion precision** (`B017` x4, `B018` x2). `B017` narrowed `assertRaises(Exception)`; the actual types were confirmed by running the tests, not inferred (`OSError`/ENOTCONN for `getpeername()` on an unconnected socket, and `AMQPConnectionError` for the open-error path, which also required making the injected fixture realistic since `ThreadSafeConnection` re-raises verbatim). `B018` replaced a bare `base_class._on_stream_connected` expression, used as a class-creation-time tripwire, with an explicit `hasattr()` assertion.

**Control flow and simplification** (`SIM117` x11, `SIM102` x5, `SIM211` x3, `SIM113` x3, `SIM114` x2, `PLR0402` x10, `PLR1730` x4). Mostly mechanical. `SIM117` was the interesting one: parenthesized context managers are a syntax error before Python 3.10 and `requires-python` is `>=3.7`, so the usual rewrite is unavailable. Joining the managers on one line leaves yapf splitting inside call brackets, which reads worse than the nesting it replaces, so the five remaining sites use `contextlib.ExitStack`, which preserves enter and exit order exactly and needs no line continuation.

**File modes** (`EXE001` x1, `EXE002` x2). `.ci/smoke_test.py` had a shebang but mode 644; nothing executes it directly (`release.yaml` runs `python .ci/smoke_test.py`), so the shebang was removed. Conversely `io_services_utils.py` and `selector_ioloop_adapter.py` were mode 755 with no shebang, an accidental bit, now cleared.

**Attribute access** (`B010` x4 fixed, `B009` x10 suppressed). See below.

### `B009` is suppressed, not fixed

This is the one decision worth reviewing. The 10 `B009` reports are all `getattr(select, 'epoll')`, `getattr(select, 'KQ_FILTER_READ')`, `getattr(socket, 'TCP_KEEPIDLE')` and similar. Those lookups are the portability mechanism: the attributes exist only on some platforms, so typeshed guards their stubs behind `sys.platform`, and resolving by name deliberately keeps them out of that per-platform view.

Applying the autofix leaves runtime unchanged (dot access raises the same `AttributeError` the surrounding `try`/`except` already handles) but breaks `mypy` on every platform, differently each time: 4 errors on linux, 2 on darwin, 6 on win32. Since the mypy workflow runs on all three, the autofix turns one green job into three red ones.

Neither alternative works. `# type: ignore[attr-defined]` is itself reported by `warn_unused_ignores` on whichever platform *does* have the attribute. `assert hasattr(...)` narrows only one module attribute per scope, so it covers a single lookup but not the four `KQ_*` constants read together. A `_select_attr(name)` wrapper does work, but adds indirection to hide a false positive.

So each site keeps `getattr()` with `# noqa: B009` and a comment naming the platform that lacks it. `RUF100` is enabled, so CI will flag these directives if a later ruff release stops reporting the rule. The single `PYI013` report is suppressed the same way, for the same reason: see below.

`B010` (`setattr(obj, 'foo', v)` to `obj.foo = v`) took the autofix; all four sites are in `tests/unit/amqp_object_tests.py`, and `mypy.ini` sets `packages = pika`, so tests are never type checked.

### Rules ignored in config

Added to the existing `ignore` list rather than fixed:

| Rule | Reports on `main` | Note |
| --- | --- | --- |
| `BLE001` blind except | 33 | Mostly adapters catching broad exceptions around user callbacks and transport teardown, in `io_services_utils.py`, `thread_safe_connection.py` and `twisted_connection.py`, plus tests and examples. |
| `DTZ005` naive `datetime.now()` | 8 | All of them interpolated into `LOGGER.debug()` calls in one acceptance test. |

These two are excluded from the 107 figure above, since the ignore entries land in the same commit as the config switch.

### Commit separation

22 commits. The first 21 are one ruff rule per commit (or one pair, where two rules share a fix), each with a body explaining the rule and any judgement call. The last folds in review feedback:

```
349765c9 ruff: address review feedback and restore TCP_USER_TIMEOUT fallback
c6820939 ruff: extend the default rule set instead of replacing it
72f40a31 ruff: fix B009 + B010
275b297c ruff: fix B018
276841c5 ruff: fix TRY002
49736ca7 ruff: fix SIM117
aeecca00 ruff: fix B017
22b8b24d ruff: fix TRY401
bdf397b5 ruff: fix EXE001 + EXE002
dbb99ac9 ruff: fix PYI036
c68efed5 ruff: fix PYI034
7b1c6b4d ruff: fix S110
c436e82d ruff: fix TRY004
7268f92c ruff: fix SIM117
9e2a5e94 ruff: fix SIM102
e27967e9 ruff: fix LOG015
1a9c3cc6 ruff: fix SIM113
24eba379 ruff: fix SIM211
3c59e5f3 ruff: fix SIM114
93ae49b7 ruff: fix PYI041
4b9fbcc8 ruff: fix PLR1730
c77ddd31 ruff: fix PLR0402
```

The config switch is deliberately the **last of the per-rule commits**. Until it lands, `lint-check` still runs against the narrower rule set, so every intermediate commit is green and the branch stays bisectable; the review-feedback commit that follows it is green under the full set. Reviewing commit by commit is the intended path; the combined diff mixes 20 unrelated rules.

Two untidy spots in that history, both cosmetic: `SIM117` appears twice (7268f92c predates a rebase onto `main`, whose work-pool changes introduced five new sites, fixed in 49736ca7), and two of the eight `PYI036` annotations were incidentally corrected in 7268f92c rather than in the `PYI036` commit.

### Behaviour change to be aware of

`Channel.confirm_delivery()` previously raised `ValueError` when `ack_nack_callback` was not callable, and its docstring documented `:raises ValueError:`. It now raises `TypeError`, matching every other callback validator in the library. Downstream code catching `ValueError` around that call will need updating. The docstring and the corresponding unit test were updated. A changelog entry may be warranted; this PR does not add one.

The other exception-type change from `TRY004`, in `_AsyncStreamConnector.__init__` (`ssl_context` validation), is on an underscore-prefixed internal class with no documented `:raises` contract, so it is not part of the public surface.

### Review feedback (349765c9)

A review pass over the branch turned up four things, all fixed in the final commit.

**`PYI013` moved from config `ignore` to a per-line `noqa`.** A config entry disables the rule for every current and future class body in the repo, and only `ConnectionParameters._DEFAULT` needs it. That line already carried `# noqa: PIE790` for the same `...`, so the directive simply becomes `# noqa: PIE790, PYI013`, matching the per-line approach already taken for `B009`. Removing the `...` outright is not an option: `docformatter` then requires a blank line after the class docstring and `yapf` requires its removal, and the two cannot both be satisfied. A `# yapf: disable`/`# yapf: enable` block does work, but spending two directive lines plus a blank line to route around a lint rule puts the workaround in the wrong tool.

**Two `ignore` comments were wrong or thin.** The `DTZ005` comment read "use `datetime.now(tz=timezone.utc)` instead of `datetime.utcnow()`", which describes `DTZ003`, a different rule. `BLE001` said only "blind except" with no rationale. Both now name the rule they actually suppress and say why the code is correct as written.

**`IOLoop._get_poller` repeated one condition three times.** `not SELECT_TYPE or SELECT_TYPE == '...'` appeared in all three poller branches, and yapf wrapped the resulting compound conditions inconsistently, which is what made the block hard to read in the first place. A local `requested(name)` helper collapses each branch to a single line. Poller selection was probed for all five `SELECT_TYPE` values before and after: identical results.

**`pika/tcp_socket_opts.py`: the dead branch was a real regression, so it is fixed here.** The PR originally left this alone to keep the diff lint-scoped, and flagged it as pre-existing. On investigation it is not merely dead code. The `except AttributeError` branch sets `TCP_USER_TIMEOUT` to `18` on Linux 2.6.37 and later, for builds where Python lacks the constant. Commit d36b4052 wrapped the `try` in `if hasattr(socket, 'TCP_USER_TIMEOUT')`, which made that branch unreachable and silently dropped the fallback. The `try`/`except` was always the guard, so the wrapper is removed. Verified by simulating a build without `socket.TCP_USER_TIMEOUT`: the fallback now sets `18` again. `mypy` stays green on linux, darwin and win32.


lukebakken (approved):
## Approving

Reviewed the full 22-commit series commit by commit, as the body recommends. This is careful, well-documented work: the `select` to `extend-select` switch is the correct fix for #1663, and splitting the 96 code fixes into one commit per rule made each judgement call reviewable in isolation. Ordering the config switch last so every intermediate commit stays lint-green and bisectable is the right call.

I verified the changes that carried real risk rather than taking them as mechanical:

- **`SIM117`.** Parenthesized context managers are a syntax error before Python 3.10 and `requires-python` is `>=3.7`, so `contextlib.ExitStack` is the right escape hatch. Checked that enter and exit order is preserved exactly at each converted site.
- **`PYI036`.** These were genuine annotation bugs, not style: `exc_type` is the exception class, `value` is the instance (not a `traceback.TracebackException`), and all three arguments are `None` on normal exit. The old signatures both accepted wrong calls and rejected correct ones.
- **`B009` suppressed rather than autofixed.** Agreed, and the reasoning holds. Name-based lookup is the portability mechanism, since typeshed guards these stubs behind `sys.platform`. The autofix would trade one green mypy job for three red ones, `type: ignore` trips `warn_unused_ignores` on whichever platform does have the attribute, and `RUF100` will flag the directives if a future ruff release stops reporting the rule.
- **`SIM113`, `TRY401`.** Confirmed the post-loop counter variable is still live where it is read, and that no exception binding became unused after the log-format edits.
- **`TRY004`.** `Channel.confirm_delivery()` moving from `ValueError` to `TypeError` is the only public-surface break, and it makes the method consistent with `validators.rpc_completion_callback`, which it calls on the very next line. Worth a changelog entry when this is released.
- **`EXE001`/`EXE002`.** Checked the invocation path: `release.yaml` runs `python .ci/smoke_test.py`, so dropping the shebang breaks nothing, and the two 755 bits on library modules were accidental.

No correctness bugs found.

I pushed one commit of my own, 349765c9, covering review feedback: `PYI013` moved from a config `ignore` to the per-line `noqa` that line already carried for `PIE790` (a config entry would disable the rule repo-wide for one site that needs it), corrected the `DTZ005` ignore comment, which described `DTZ003`, gave the `BLE001` ignore an actual rationale, and extracted the thrice-repeated `SELECT_TYPE` check in `IOLoop._get_poller` into a local helper. Poller selection was probed for all five `SELECT_TYPE` values before and after with identical results.

That commit also fixes the item the body had listed as pre-existing and out of scope. The `except AttributeError` branch in `tcp_socket_opts.py` is not merely dead: it sets `TCP_USER_TIMEOUT` to 18 on Linux 2.6.37 and later for builds where Python lacks the constant, and the `hasattr` wrapper added in d36b4052 (mine, not yours) made it unreachable and silently dropped that fallback. The `try`/`except` was always the guard, so the wrapper is gone.

Local verification: `lint-check`, `fmt-check`, `docfmt-check` and `typecheck` all clean, mypy green on linux, darwin and win32, 903 unit tests passing. CI is green across all 60 checks. The one red mark, `codecov/patch` at 81.57% against a 100% target, was already failing before my commit at 81.15%, and a lint refactor that adds no new logic is exactly the case that gate reads wrong.

Thanks for the thorough PR body. The per-rule rationale made this much faster to review than the combined diff would have been.

PR #1667, opened 2026-07-30
fix: Stop the getaddrinfo acceptance tests from depending on live DNS

### LINKED ISSUE #1661: Flaky getaddrinfo acceptance tests depend on live DNS
## Summary

The `getaddrinfo` acceptance tests in `tests/acceptance/io_services_tests.py` perform live DNS lookups against the public internet (`www.google.com` and a deliberately non-existent `www.google.comSSS`). When outbound DNS is slow or unavailable on the CI runner they time out, failing the build for reasons unrelated to the change under test.

## Observed failure

Seen on the `macos-15-intel` runner (Python 3.9, both tls=true and tls=false) in a recent CI run. All other legs (Linux and Windows, every Python version) passed.

```
FAILED tests/acceptance/io_services_tests.py::TestGetaddrinfoWWWGoogleDotComPort80::test_with_asyncio_io_services - AssertionError: The test timed out.
FAILED tests/acceptance/io_services_tests.py::TestGetaddrinfoNonExistentHost::test_with_asyncio_io_services - AssertionError: The test timed out.
FAILED tests/acceptance/io_services_tests.py::TestGetaddrinfoWWWGoogleDotComPort80::test_with_select_connection_io_services - AssertionError: The test timed out.
FAILED tests/acceptance/io_services_tests.py::TestGetaddrinfoNonExistentHost::test_with_select_connection_io_services - AssertionError: The test timed out.
FAILED tests/acceptance/io_services_tests.py::TestGetaddrinfoWWWGoogleDotComPort80::test_with_tornado_io_services - AssertionError: The test timed out.
FAILED tests/acceptance/io_services_tests.py::TestGetaddrinfoNonExistentHost::test_with_tornado_io_services - AssertionError: The test timed out.
```

The timeout also produces a follow-on teardown error, `RuntimeError: Cannot close a running event loop`, because the event loop is still running the in-flight `getaddrinfo()` when the test times out.

Because GitHub Actions matrices default to `fail-fast: true`, the sibling `tls=false` leg was `cancelled` rather than run to completion, and the `tests-passed` aggregator went red as a result.

## Root cause

The tests call `nbio.getaddrinfo('www.google.com', 80, ...)` and `nbio.getaddrinfo('www.google.comSSS', 80, ...)` directly. These resolve real names over the network, so the tests depend on:

- outbound DNS being reachable and fast from the runner, and
- `www.google.com` continuing to resolve and `www.google.comSSS` continuing to not resolve.

None of that is under the test suite's control, so the tests are inherently flaky in constrained or slow-DNS CI environments.

## Possible fixes (to be decided)

- Resolve a loopback / localhost name instead of a public host, so no external DNS is needed.
- Monkeypatch / stub the resolver so `getaddrinfo` returns a canned result and a canned `gaierror`, testing the async plumbing without the network.
- Mark these tests to skip when a quick DNS reachability probe fails, so a runner without outbound DNS skips rather than fails.

## Notes

- Not a regression: these tests predate current work and fail only on network-constrained runners.
- Surfaced while landing unrelated changes on #1651, where the red board was traced entirely to these DNS timeouts on one runner; the changes under review passed on every leg that ran to completion.

alonfaraj:
## Proposed Changes

The `getaddrinfo` acceptance tests resolved `www.google.com` and `www.google.comSSS` against the public internet. On a runner where outbound DNS is slow or unavailable they blow through the 15 second per-test timeout and fail with `AssertionError: The test timed out.` That is exactly how #1661 was reported: six failures on `macos-15-intel` py3.9, both TLS legs, while every Linux and Windows leg passed.

Nothing in those tests is about the public internet. What they cover is the async plumbing of the `getaddrinfo` adaptations: that a result is delivered on the loop thread, that a resolution failure arrives as a value rather than an exception, and that a pending lookup can be cancelled. The network dependency buys no coverage and costs determinism, so this removes it.

Three commits, each independently reviewable:

### 1. tests: stop the getaddrinfo tests from using live DNS

The issue lists three candidate fixes as alternatives, but the five call sites assert different things and only one of them actually needs a stub, so this picks per test:

- `TestGetaddrinfoWWWGoogleDotComPort80` becomes `TestGetaddrinfoLocalhostPort80` and resolves `localhost`. That comes from the hosts file, so no network is involved, while a real `socket.getaddrinfo()` call still covers argument forwarding and the shape of the result end to end. Every existing assertion is unchanged. An IP address literal would also avoid the network, but it would stop exercising name resolution, which is the point of the test.
- `TestGetaddrinfoNonExistentHost` stubs the resolver. This is the one case that cannot be made hermetic with a real hostname: a name that does not resolve is not obtainable without the network, not even under the `.invalid` TLD reserved by RFC 6761, because the stub resolver still puts the query on the wire and waits for `NXDOMAIN`. Patching `socket.getaddrinfo()` is a single seam that covers all three adaptations, since both `selector_ioloop_adapter._AddressResolver` and asyncio's `loop.getaddrinfo()` bottom out in it. The assertion under test is unchanged: the `gaierror` must arrive as the value passed to `on_done`.
- The two cancellation tests block the resolver. They never waited for a result, so they did not time out, but they did leave a public DNS lookup running past the end of the test, which is what their `Event loop is closed` notes described. A lookup that cannot complete also removes the race their other note warned about, so the cancellation window is now deterministic rather than probabilistic.

### 2. tests: keep cleanup from burying a test's timeout failure

`create_nbio()` registered `nbio.close` directly as a cleanup. Cleanups run on the main thread, after `run_in_thread_with_timeout` has abandoned the thread running `start()`, so on a timeout the loop is still running and closing it fails: asyncio and tornado raise `RuntimeError: Cannot close a running event loop`, and `select_connection` asserts `Cannot call close() before start() unwinds.` Every timed-out test therefore reported that unrelated teardown error next to the timeout, which is the follow-on error described in the issue.

Closing now goes through `_close_nbio()`, which requests a stop that only a still-running loop honors, then closes once and warns rather than raises if the close does not take. The DNS timeouts surfaced this, but any future timeout in this suite would have been reported the same way.

### 3. ci: stop a failing matrix leg from cancelling its siblings

All three jobs in the reusable test workflow declare a matrix with no `fail-fast` key, so it defaulted to `true`. When the macOS leg failed, its `tls=false` sibling was cancelled rather than run and `tests-passed` went red with half the picture. This sets `fail-fast: false`. It makes no test less flaky; it keeps one failing leg from erasing the results of the twenty-odd that would have passed, which is what makes a flaky failure hard to tell apart from a real one.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

## Fixes

- Fixes #1661

### Why not the third option in the issue

A DNS reachability probe that skips the tests was the one option not taken. It adds its own latency, silently deletes coverage on exactly the runners where it fires, and would not have helped the reported failure: the macOS runner resolves names, just slowly, so it passes a probe and then still exceeds the timeout.

### DIFF
--- a/.github/workflows/_test.yaml
+++ b/.github/workflows/_test.yaml
@@ -26,6 +26,7 @@ jobs:
     name: build/test on windows-latest py${{ matrix.python-version }} tls=${{ matrix.test-tls }}
     runs-on: windows-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: ${{ fromJSON(inputs.python-versions) }}
         test-tls: [true, false]
@@ -81,6 +82,7 @@ jobs:
     name: build/test on ${{ inputs.ubuntu-runner }} py${{ matrix.python-version }} tls=${{ matrix.test-tls }}
     runs-on: ${{ inputs.ubuntu-runner }}
     strategy:
+      fail-fast: false
       matrix:
         python-version: ${{ fromJSON(inputs.python-versions) }}
         test-tls: [true, false]
@@ -110,6 +112,7 @@ jobs:
     name: build/test on ${{ inputs.macos-runner }} py${{ matrix.python-version }} tls=${{ matrix.test-tls }}
     runs-on: ${{ inputs.macos-runner }}
     strategy:
+      fail-fast: false
       matrix:
         python-version: ${{ fromJSON(inputs.python-versions) }}
         test-tls: [true, false]
--- a/tests/acceptance/io_services_tests.py
+++ b/tests/acceptance/io_services_tests.py
@@ -1,11 +1,14 @@
 """Tests of nbio_interface.AbstractIOServices adaptations."""
 
 import collections
+import contextlib
 import errno
 import logging
 import os
 import socket
+import threading
 import unittest
+from unittest import mock
 
 import pika._utils
 from pika.adapters.utils import nbio_interface
@@ -403,8 +406,49 @@ class TestSocketWatchersAfterLocalPeerShutsReadWrite(SocketWatcherTestBase,
         self._check_socket_watchers_fired(s1, expected)
 
 
-class TestGetaddrinfoWWWGoogleDotComPort80(AsyncServicesTestBase,
-                                           IOServicesTestStubs):
+@contextlib.contextmanager
+def _failing_resolver():
+    """
+    Patch `socket.getaddrinfo()` to fail as if the host did not exist.
+
+    A name that does not resolve cannot be obtained without the network, not even under the
+    `.invalid` TLD reserved by RFC 6761, because the stub resolver still sends the query and waits
+    for `NXDOMAIN`. Patching `socket.getaddrinfo()` covers every adaptation under test, since both
+    `selector_ioloop_adapter._AddressResolver` and asyncio's `loop.getaddrinfo()` bottom out in it.
+    """
+
+    def getaddrinfo_failure(*args, **kwargs):
+        raise socket.gaierror(socket.EAI_NONAME, 'Name or service not known')
+
+    with mock.patch.object(socket, 'getaddrinfo', getaddrinfo_failure):
+        yield
+
+
+@contextlib.contextmanager
+def _blocked_resolver():
+    """
+    Patch `socket.getaddrinfo()` to block until the context manager exits.
+
+    The cancellation tests need a lookup that is still pending when they cancel it. Blocking the
+    resolver removes the race with a lookup that completes first, and keeps an abandoned lookup from
+    outliving the test in a resolver thread.
+    """
+    released = threading.Event()
+
+    def getaddrinfo_blocked(*args, **kwargs):
+        released.wait()
+        # Every caller cancels the lookup, so nothing observes this result
+        return []
+
+    try:
+        with mock.patch.object(socket, 'getaddrinfo', getaddrinfo_blocked):
+            yield
+    finally:
+        released.set()
+
+
+class TestGetaddrinfoLocalhostPort80(AsyncServicesTestBase,
+                                     IOServicesTestStubs):
 
     def start(self):
         # provided by IOServicesTestStubs mixin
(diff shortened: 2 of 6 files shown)


lukebakken (approved):
## Approving

Verified locally against a live broker, plus the full 61-check CI run on 44ab9253.

### What the PR does

The getaddrinfo acceptance tests previously resolved real hostnames, so they were slow and failed on hosts with no DNS or with a resolver that answers for anything. They now patch `socket.getaddrinfo`, which is the single seam all three adaptations bottom out in: `_AddressResolver._resolve()` calls it directly, and asyncio's `loop.getaddrinfo()` reaches it through `run_in_executor`. The 12 getaddrinfo tests run in 0.16s and no longer touch the network.

### Verified rather than assumed

- The synthesized `gaierror` matches the shape the real resolver raises, so the error-path assertions still mean what they meant.
- `run_in_thread_with_timeout` abandons the thread it runs `start()` in, leaving the loop running. I confirmed the resulting `close()` failure per adapter: `RuntimeError` from asyncio and tornado, `AssertionError` from `select_connection.IOLoop`. That is what the cleanup helper has to survive.
- Tornado really does allow only one close attempt: the first succeeds, the second raises `KeyError`, which is why the helper does not retry.
- `-O` is not used anywhere in the test invocation, so the `select_connection` assert is live and not stripped.

### Review feedback (44ab9253)

Three things the first pass left too broad. Disclosure: 44ab9253 is my own commit, so that part of this PR is not independently reviewed.

- `_close_nbio` caught `Exception`, so a genuine `close()` regression was downgraded to a log line and the test still passed. Narrowed to the two errors an abandoned loop actually produces; a 5-case probe confirms genuine `ValueError`/`KeyError` now fail the test again while both abandoned-loop errors stay tolerated.
- `_blocked_resolver()` waited forever. An abandoned test therefore left `socket.getaddrinfo` patched for every later test in the process and hung interpreter exit, because `concurrent.futures` joins the asyncio executor thread the stub is parked in. Probed both ways: unbounded exits 124, bounded exits 0. The bound sits above the per-test timeout, so only an already-abandoned test can reach it.
- The two `thread_safe` acceptance modules each had their own `_safe_close`. Now one `safe_close` in `tests/misc/test_utils.py`, forwarding `*args`/`**kwargs` so the bounded-queue call site keeps its `timeout`.

### Notes

- The branch is 2 commits behind `main`. The only content in that gap is a README badge fix (9ea5d5a8), which cannot affect these tests.
- `codecov/patch` is green here, unlike on #1666.

Issue #1669, opened 2026-08-02
Remove the final 7 `type: ignore` comments in 2.0


lukebakken:
## Summary

Split out of #1618. Seven `type: ignore` suppressions cannot be removed in 1.5.0 because they live in adapters that #1608 deprecates but does not remove until 2.0, or because removing them means changing a public contract.

The counts and diagnoses below were measured on 88df7ca2 by stripping the suppressions and reading mypy's actual output.

### gevent `[import-untyped]` (4)

```
pika/adapters/gevent_connection.py:13  import gevent
pika/adapters/gevent_connection.py:14  import gevent._interfaces
pika/adapters/gevent_connection.py:15  import gevent.hub
pika/adapters/gevent_connection.py:16  import gevent.socket
```

gevent ships no `py.typed`. `types-gevent` exists on PyPI and is actively maintained, and installing it does clear all four, but it is not a free win:

- `gevent._interfaces` is not covered by the stubs, so line 14 turns into `[import-not-found]`. The adapter references `gevent._interfaces.ILoop` in two public signatures (`GeventConnection.__init__` and `create_connection`), so this is not a trivially droppable import.
- Six new genuine type errors surface in the adapter once the stubs are in place: a missing annotation on `_waiter`, an `Optional` assignment to a `Hub`-typed attribute, a `functools.partial` vs `Callable[[], None]` mismatch, a greenlet assigned to a `None`-typed attribute, and an `Exception` assigned to a `getaddrinfo`-result-typed attribute.

These disappear entirely when the gevent adapter is removed per #1608, which is strictly less work than adopting the stubs and then deleting the adapter anyway.

### Twisted `ITransport` (2)

```
pika/adapters/twisted_connection.py:1122  self._transport.loseConnection()   [misc]
pika/adapters/twisted_connection.py:1130  self._transport.write(data)        [call-arg]
```

`twisted.internet.interfaces.ITransport` is a `zope.interface.InterfaceClass` instance, not a class, so mypy cannot interpret it as a type. Methods declared on it have no `self` parameter, which is why mypy reports "does not accept self argument" and "Too many arguments".

The `mypy-zope` plugin resolves both, but it is a wash: enabling it introduces two different errors in the same file (`callLater` `[arg-type]` and `callFromThread` `[union-attr]`), so the suppression count does not improve while the project takes on a plugin dependency. Not worth it for code slated for removal.

### Twisted `ClosableDeferredQueue.put` override (1)

```
[...]

Issue #1564, opened 2026-05-01
Evaluate pyright as a type checker


lukebakken:
## Summary

Evaluate adding pyright as a type checker alongside or instead of mypy. PR #1529 (type annotations) includes 76 `pyright: ignore` comments from the contributor's pyright-based workflow. These comments are currently inert in CI since pika only runs mypy.

## Motivation

Pyright catches type errors that mypy does not, particularly around `None` safety, attribute access on union types, and incompatible method overrides. It is the engine behind Pylance (VS Code's Python type checking) and has growing adoption in the Python ecosystem.

## Considerations

- Pyright and mypy disagree on some typing edge cases (e.g., `namedtuple` patterns, `type: ignore` scoping). Running both may require reconciling conflicting diagnostics.
- Pyright is stricter by default, which may surface additional issues that need `pyright: ignore` or code changes.
- Adding a second type checker increases CI time and maintenance burden.
- Pyright is written in TypeScript and distributed via npm/PyPI (`pyright`), which is a different dependency profile than mypy.

## Proposed work

- [ ] Run pyright against the codebase and assess the current error count
- [ ] Decide whether to run pyright alongside mypy or replace mypy with pyright
- [ ] If adopting, add a `pyrightconfig.json` and a CI workflow
- [ ] Resolve or suppress any new diagnostics

PR #1670, opened 2026-08-02
Remove six `type: ignore` comments from three adapters

lukebakken:
## Summary

First of three PRs for #1618, covering 6 of the 14 suppressions fixable in 1.5.0.

| File | Removed | Change |
|------|---------|--------|
| `adapters/base_connection.py` | 4 `[assignment]` | `_StreamingProtocolShim`'s four `None` class attributes become real `@override` methods delegating to `conn._proto_*` |
| `adapters/blocking_connection.py` | 1 `[assignment]` | `_impl` declared without a value; the two readers that can run on a half-built instance use `getattr` |
| `adapters/select_connection.py` | 1 `[attr-defined, assignment, unused-ignore]` | `sys.platform` guards in `KQueuePoller` in place of the suppression |

Repo-wide `type: ignore` count drops from 21 to 15. The remaining 15 are PRs 2 and 3 of #1618 plus the 7 deferred to #1669.

## The kqueue suppression was hiding unchecked code

This is the substantive part of the PR. The suppression listed `assignment`, which left `self._kqueue` inferred as `None`. Every statement after the `if self._kqueue is None: return` guard was therefore unreachable, so mypy has never type-checked the body of `KQueuePoller`.

Measured with `warn_unreachable = True` under `--platform darwin`:

- On main: 4 unreachable statements inside `KQueuePoller`, 27 repo-wide.
- After this change: 0 inside `KQueuePoller`, 23 repo-wide, with no new unreachable code anywhere.

To confirm the restored checking is real and not cosmetic, I injected two errors into a scratch copy of the poller: `select.KQ_EV_NOPE` in place of a valid constant, and a `str` where `kevent()` requires an `int`. Both are caught under `--platform darwin` and both are invisible on Linux.

CI already enforces this. `mypy --platform` defaults to `sys.platform` and `mypy.ini` sets no `platform` key, so the existing `macos-latest` leg of the mypy workflow checks the poller body as soon as the suppression is gone. No workflow change is needed. There are also 17 kqueue tests in `tests/unit/select_connection_ioloop_tests.py` gated on `hasattr(select, 'kqueue')`; they skip on Linux and run on the macOS legs.

### Why the guard is repeated in three methods

Each of `_map_event`, `_init_poller`, and `_modify_fd_events` opens with the same `sys.platform` test. I tried two ways to avoid the repetition, and both fail:

- Aliasing the test to a module-level constant. mypy does not narrow through the alias, so `select.kqueue` still errors on Linux.
- Defining the class under a single `if sys.platform != 'linux' and sys.platform != 'win32':`. This breaks the `KQueuePoller` reference in `_get_poller()` at `select_connection.py:466` with `[name-defined]` on Linux.

Inline `sys.platform` checks are the only form mypy narrows on, which is also how typeshed declares these attributes.

## Behavior changes

Constructing `KQueuePoller` directly on Linux now raises `NotImplementedError` instead of `AttributeError` from `select.kqueue`. `_get_poller()` only constructs it when `select.kqueue` exists, so neither path is reachable in normal use, and both are errors either way.

`BlockingConnection._impl` no longer exists as an attribute until `_create_connection()` succeeds, where it was previously `None`. Two members can run against a half-built instance and both read the attribute with `getattr`:

- `_cleanup()`, which `_create_connection()` itself calls from its `except` block before re-raising.
- `__repr__()`, which runs from any traceback or debugger rendering the frame's locals.

I enumerated these rather than assuming: 14 members read `_impl`, and the other 12 are public API reachable only after a successful workflow. Verified end to end against an unreachable port, where `_cleanup()` runs with `_impl` absent from `__dict__`, `repr()` renders `impl=None`, and the real `AMQPConnectionError` surfaces intact.

Nothing else changes at runtime. The shim's four methods do exactly what the previously assigned bound methods did.

## Review fixes

alonfaraj caught that the first push had `__repr__` still reading `_impl` directly, which turned a broker-unreachable `AMQPConnectionError` into an `AttributeError` from the repr. Fixed with a regression test that captures the genuine half-built instance via `autospec=True` and exercises both readers. Worth recording that type checking cannot catch this shape: `self._impl: SelectConnection` with no value tells mypy the attribute exists, and mypy does not verify that every path assigns it, even with `possibly-undefined` and `warn_unreachable` enabled.

## Testing

- mypy clean under the default platform plus `--platform darwin`, `--platform win32`, and `--platform linux`.
- `hatch run test`: 1462 passed, 61 skipped, against a live RabbitMQ broker.

## Side note for #1564

This drops `select_connection.py` from 17 `pyright: ignore` comments to 4. Untested, since pyright has not been run against the tree.

See #1618

### DIFF
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -10,6 +10,7 @@ import errno
 import heapq
 import logging
 import select
+import sys
 import threading
 import time
 from typing import TYPE_CHECKING, Any, Callable, Sequence, Union
@@ -1091,12 +1092,22 @@ class SelectPoller(_PollerBase):
 
 
 class KQueuePoller(_PollerBase):
-    """KQueuePoller works on BSD based systems and is faster than select."""
+    """
+    KQueuePoller works on BSD based systems and is faster than select.
+
+    `select.kqueue`, `select.kevent`, and the `KQ_*` constants exist only on BSD
+    platforms, and typeshed declares them behind a `sys.platform` guard to match.
+    Every method that touches them therefore opens with the same guard, which
+    narrows the platform for the type checker and never fires at runtime:
+    `_get_poller()` only constructs this class when `select.kqueue` exists.
+    """
 
     def __init__(self, get_wait_seconds: Callable[[], float | None],
                  process_timeouts: Callable[[], None]) -> None:
         """Create an instance of the KQueuePoller."""
-        self._kqueue = None
+        # `Any` because the annotation cannot name `select.kqueue` on platforms
+        # where the type checker has no such attribute to refer to.
+        self._kqueue: Any | None = None
         super().__init__(get_wait_seconds, process_timeouts)
 
     @staticmethod
@@ -1107,13 +1118,13 @@ class KQueuePoller(_PollerBase):
         :param kevent: a kevent object as returned by kqueue.control()
         """
         mask = 0
-        # Looked up by name because these constants exist only on kqueue
-        # platforms; a direct `select.KQ_*` reference fails type checking
-        # everywhere else, and `warn_unused_ignores` rules out `type: ignore`.
-        kq_filter_read = getattr(select, 'KQ_FILTER_READ')  # noqa: B009
-        kq_filter_write = getattr(select, 'KQ_FILTER_WRITE')  # noqa: B009
-        kq_ev_eof = getattr(select, 'KQ_EV_EOF')  # noqa: B009
-        kq_ev_error = getattr(select, 'KQ_EV_ERROR')  # noqa: B009
+        if sys.platform == 'linux' or sys.platform == 'win32':
+            raise NotImplementedError('kqueue is available on BSD only.')
+
+        kq_filter_read = select.KQ_FILTER_READ
+        kq_filter_write = select.KQ_FILTER_WRITE
+        kq_ev_eof = select.KQ_EV_EOF
+        kq_ev_error = select.KQ_EV_ERROR
 
         if kevent.filter == kq_filter_read:
             mask = PollEvents.READ
@@ -1158,7 +1169,10 @@ class KQueuePoller(_PollerBase):
         """Notify the implementation to allocate the poller resource."""
         assert self._kqueue is None
 
-        self._kqueue = select.kqueue()  # type: ignore[attr-defined, assignment, unused-ignore]  # yapf: disable
+        if sys.platform == 'linux' or sys.platform == 'win32':
+            raise NotImplementedError('kqueue is available on BSD only.')
+
+        self._kqueue = select.kqueue()
 
     @override
     def _uninit_poller(self) -> None:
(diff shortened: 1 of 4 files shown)


pika/adapters/select_connection.py
```diff
@@ -1091,12 +1092,22 @@ def _unregister_fd(self, fileno: int, events_to_clear: int) -> None:
...
+    `_get_poller()` only constructs this class when `select.kqueue` exists.
+    """
 
     def __init__(self, get_wait_seconds: Callable[[], float | None],
                  process_timeouts: Callable[[], None]) -> None:
         """Create an instance of the KQueuePoller."""
-        self._kqueue = None
+        # `Any` because the annotation cannot name `select.kqueue` on platforms
+        # where the type checker has no such attribute to refer to.
+        self._kqueue: Any | None = None
```
alonfaraj: `Any | None` can be simplify to `Any`

alonfaraj (commented):
`_cleanup()` is not the only reader of `_impl` on a half-built instance. `__repr__` is too, and it still accesses the attribute directly, so without the `_impl = None` pre-assignment it raises.

Repro, with a real failure rather than a mock:

```python
import pika

URL = 'amqp://guest:guest@127.0.0.1:1/%2F?connection_attempts=1&socket_timeout=1'

def test_connect_to_dead_port():
    pika.BlockingConnection(pika.URLParameters(URL))
```

```console
$ pytest --showlocals test_repro.py

# main
self = <BlockingConnection impl=None>

# this branch
self = <[AttributeError("'BlockingConnection' object has no attribute '_impl'") raised in repr()] BlockingConnection object at 0x74cd0dd5f620>
```

This fires on the ordinary "broker is not reachable" path, which is exactly when someone is reading a traceback. Here `saferepr` contains it, but pdb's `p self` or any reporter that captures frame locals will propagate the `AttributeError` and bury the `AMQPConnectionError` you were chasing.

Fix:

```python
@override
def __repr__(self) -> str:
    impl = getattr(self, '_impl', None)
    return f'<{self.__class__.__name__} impl={impl!r}>'
```

pika/adapters/select_connection.py
lukebakken: Applied. Verified the class body stays reachable to mypy after the simplification, so the suppression removal this PR is about still holds.

lukebakken:
Good catch, and thanks for the runnable repro. Reproduced it exactly: `main` prints `impl=None`, this branch turned the `AMQPConnectionError` into a repr failure.

`__repr__` now reads `_impl` with `getattr()`, matching `_cleanup()`, plus a regression test.

I enumerated the rest rather than guessing: 14 members read `_impl`, and only `_cleanup()` and `__repr__()` can run before `__init__` completes. The other 12 are public API reachable only after a successful workflow. `BlockingChannel._impl` is assigned from a constructor argument on the first line of `__init__`, so it has no half-built window.

Worth noting for anyone reading later: type checking cannot catch this shape. `self._impl: SelectConnection` with no value tells mypy the attribute exists; mypy does not verify that every path assigns it, even with `possibly-undefined` and `warn_unreachable` enabled.

alonfaraj (approved):

PR #1671, opened 2026-08-02
Make `AbstractSelectorIOLoop` a structural `Protocol`

lukebakken:
## Summary

Second of three PRs for #1618, covering item 2: the 2 tornado `[arg-type]` suppressions. It also removes 6 obsolete `pyright: ignore` comments that fall out of the same change.

Based directly on the #1670 merge commit, so it contains that work and applies cleanly.

| File | Change |
|------|--------|
| `adapters/utils/selector_ioloop_adapter.py` | `AbstractSelectorIOLoop` becomes a `@runtime_checkable` `Protocol`; `READ`/`WRITE`/`ERROR` become `int` attribute declarations; 12 `@abc.abstractmethod` decorators and the `abc` import removed |
| `adapters/tornado_connection.py` | 2 `type: ignore[arg-type]` removed |
| `adapters/select_connection.py` | 3 obsolete `pyright: ignore` removed |
| `adapters/gevent_connection.py` | 3 obsolete `pyright: ignore` removed |
| `_utils.py` | `Protocol`/`runtime_checkable` re-exported with a 3.7 fallback |
| `tests/unit/selector_ioloop_adapter_tests.py` | new: 11 conformance tests |

## Why a Protocol

The class already documents itself as an interface that a loop may satisfy "but not necessarily derived from it", and `SelectorIOServicesAdapter.__init__` repeats that in its own docstring. A nominal base class cannot express it. `tornado.ioloop.IOLoop` implements every member but does not inherit from `AbstractSelectorIOLoop`, so both call sites handing one to `SelectorIOServicesAdapter` needed `type: ignore[arg-type]`. Structural typing is what the interface was describing all along.

## The six pyright suppressions

`READ`/`WRITE`/`ERROR` were abstract properties, but every implementation assigns a plain `int`. That mismatch is why all three flags in both `IOLoop` and `_GeventSelectorIOLoop` carried `pyright: ignore[reportAssignmentType, reportIncompatibleMethodOverride]`. Declaring them as `int` attributes on the Protocol makes the assignments correct.

Measured at the branch tip: pyright reports **6** diagnostics across the four affected files both before and after, and none concern `READ`, `WRITE`, or `ERROR`. `[tool.pyright]` in `pyproject.toml` is empty, so this is pyright at its defaults.

## Flags are variable members, not properties

Protocol variable members are not satisfied by read-only properties, so an implementation must expose plain `int` attributes. This is narrower than the abstract-property declaration it replaces, where an implementation could use either shape, and the trade-off is deliberate.

Declaring the flags as read-only property members is the more permissive alternative under mypy, and it was tested rather than assumed. pyright then reports **18** diagnostics on the same four files instead of 6, and the 12 additions are exactly `reportAssignmentType` x6 and `reportIncompatibleMethodOverride` x6: every diagnostic the six retired suppressions were covering, back again. Both in-tree loops use plain ints, and so does tornado's `IOLoop` (`READ` 1, `WRITE` 4, `ERROR` 24), so the narrower declaration costs nothing in the tree and keeps pyright at parity, which matters for #1564.

## The `@abc.abstractmethod` decorators

12 decorators are removed, 3 on the flags and 9 on the methods.

They are **not** inert under `Protocol`. `typing._ProtocolMeta` derives from `abc.ABCMeta`, so keeping them would populate `__abstractmethods__` and make an incomplete subclass fail at construction. Verified on 3.14: re-adding them populates all 12, and a subclass missing `remove_handler` raises `TypeError`.

They stay removed anyway, for a reason specific to this project's support window. pika supports 3.7, where `typing.Protocol` does not exist and the base falls back to `object`; there `__abstractmethods__` is absent entirely and an incomplete subclass instantiates fine. Keeping the decorators would buy runtime enforcement on 3.8 through 3.14 and nothing on 3.7, so the behavior would differ across legs of the same test matrix. mypy reports a missing method on every version either way.

## Runtime behavior

Measured by running the code on both 3.14.5 and 3.7.17:

- **`AbstractSelectorIOLoop()` raises `TypeError: Protocols cannot be instantiated`** on 3.8+, and succeeds on 3.7 under the fallback. Nothing in the tree constructs it.
- **`isinstance` still works** because of `@runtime_checkable`, and now also returns `True` for a loop that satisfies the interface without deriving from it. On 3.7 the fallback makes it nominal, so a non-deriving loop returns `False` there. Nothing in the repo does `isinstance` against this class.
- **`issubclass` raises `TypeError`** on 3.8+. A `Protocol` with non-method members cannot support it, and this one has three. Nothing in the tree calls it.
- **Both in-tree implementations still work.** `IOLoop` and `_GeventSelectorIOLoop` subclass the Protocol explicitly, instantiate fine, and implement all 9 protocol methods.

`AbstractSelectorIOLoop` lives in `pika/adapters/utils/`, which `CLAUDE.md` describes as "adapter internals". It is not in any `__all__` and not referenced in the docs.

## Tests

`tests/unit/selector_ioloop_adapter_tests.py` is new. Nothing in `tests/` referenced `AbstractSelectorIOLoop` before, even though structural conformance is the whole premise of this change.

11 tests, all asserted against classes rather than instances so no loop, hub, or file descriptor is created. Method names are derived from the protocol at import time, so a member added to the interface is immediately required of all three loops checked (`select_connection.IOLoop`, `_GeventSelectorIOLoop`, `tornado.ioloop.IOLoop`).

Confirmed the tests bite, by mutating the source four ways and checking each mutation fails:

| Mutation | Caught by |
|---|---|
| Protocol flags declared as properties | `test_flags_are_declared_as_variable_members` |
| `select_connection.IOLoop` flags as properties | `test_flags_are_int_attributes` |
| New protocol method no loop implements | `test_implements_every_protocol_method` and the `isinstance` test |
| `@abc.abstractmethod` re-added | `test_carries_no_abstract_methods` |

## Testing

- mypy clean under the default platform plus `--platform darwin` and `--platform win32`.
- `hatch run test`: 1473 passed, 61 skipped against a live RabbitMQ broker.
- Python 3.7.17 unit suite: 912 passed, 24 skipped. The 2 `isinstance` tests skip there by design, since the fallback base is nominal.
- pyright: 6 diagnostics before and after, with 6 fewer suppressions.

## Corrections to the original description

Three claims in the first version of this description were wrong, and are fixed above:

- It said pyright reported "the same 10 diagnostics" before and after. The count is 6. The parity holds; the number did not.
- It said the `@abc.abstractmethod` decorators "were already inert". That was true of the `object`-based class before the conversion and false after it, since `_ProtocolMeta` derives from `ABCMeta`. Thanks to @alonfaraj for catching both this and the flags narrowing.
- It said 8 decorators were removed. It is 12.

See #1618

### DIFF
--- a/pika/_utils.py
+++ b/pika/_utils.py
@@ -25,7 +25,25 @@ else:
         return func
 
 
-__all__ = ['override']
+# `Protocol` and `runtime_checkable` (PEP 544) landed in the stdlib in 3.8,
+# while pika still supports 3.7. Same dependency-free approach as `override`
+# above: type checkers run at 3.10 or later and see the real objects, while a
+# 3.7 runtime falls back to a plain base class and a no-op decorator. Protocols
+# in pika are only ever subclassed explicitly, never instantiated or passed to
+# `isinstance`, so the nominal fallback behaves identically at runtime.
+# Re-exported across pika (declared in __all__ below).
+if sys.version_info >= (3, 8):
+    from typing import Protocol, runtime_checkable
+elif TYPE_CHECKING:
+    from typing_extensions import Protocol, runtime_checkable
+else:
+    Protocol = object
+
+    def runtime_checkable(cls):
+        return cls
+
+
+__all__ = ['Protocol', 'override', 'runtime_checkable']
 
 RE_NUM = re.compile(r'(\d+).+')
 
--- a/pika/adapters/gevent_connection.py
+++ b/pika/adapters/gevent_connection.py
@@ -212,9 +212,9 @@ class _GeventSelectorIOLoop(AbstractSelectorIOLoop):
     # Gevent's READ and WRITE masks are defined as 1 and 2 respectively. No
     # ERROR mask is defined.
     # See https://www.gevent.org/api/gevent.hub.html#gevent._interfaces.ILoop.io
-    READ = 1  # pyright: ignore[reportAssignmentType, reportIncompatibleMethodOverride]
-    WRITE = 2  # pyright: ignore[reportAssignmentType, reportIncompatibleMethodOverride]
-    ERROR = 0  # pyright: ignore[reportAssignmentType, reportIncompatibleMethodOverride]
+    READ = 1
+    WRITE = 2
+    ERROR = 0
 
     def __init__(self, gevent_hub: gevent.hub.Hub | None = None) -> None:
         """
--- a/pika/adapters/select_connection.py
+++ b/pika/adapters/select_connection.py
@@ -403,9 +403,9 @@ class IOLoop(AbstractSelectorIOLoop):
     """
 
     # READ/WRITE/ERROR per `AbstractSelectorIOLoop` requirements
-    READ = PollEvents.READ  # pyright: ignore[reportIncompatibleMethodOverride, reportAssignmentType]
-    WRITE = PollEvents.WRITE  # pyright: ignore[reportIncompatibleMethodOverride, reportAssignmentType]
-    ERROR = PollEvents.ERROR  # pyright: ignore[reportIncompatibleMethodOverride, reportAssignmentType]
+    READ = PollEvents.READ
+    WRITE = PollEvents.WRITE
+    ERROR = PollEvents.ERROR
 
     def __init__(self) -> None:
         self._timer = _Timer()
--- a/pika/adapters/tornado_connection.py
+++ b/pika/adapters/tornado_connection.py
@@ -69,8 +69,7 @@ class TornadoConnection(base_connection.BaseConnection):
             nbio = custom_ioloop
         else:
             nbio = (selector_ioloop_adapter.SelectorIOServicesAdapter(
-                custom_ioloop or
-                ioloop.IOLoop.instance()))  # type: ignore[arg-type]
+                custom_ioloop or ioloop.IOLoop.instance()))
         super().__init__(
             parameters,
             on_open_callback,
@@ -98,7 +97,7 @@ class TornadoConnection(base_connection.BaseConnection):
         :param workflow: Optional connection workflow instance to use
         """
         nbio = selector_ioloop_adapter.SelectorIOServicesAdapter(
-            custom_ioloop or ioloop.IOLoop.instance())  # type: ignore[arg-type]
+            custom_ioloop or ioloop.IOLoop.instance())
 
         def connection_factory(
                 params: connection.Parameters | None) -> TornadoConnection:
--- a/pika/adapters/utils/selector_ioloop_adapter.py
+++ b/pika/adapters/utils/selector_ioloop_adapter.py
@@ -7,9 +7,9 @@ from __future__ import annotations
 import logging
 import socket
 import threading
-from typing import Any, Callable, Protocol, runtime_checkable
+from typing import Any, Callable
 
-from pika._utils import override
+from pika._utils import Protocol, override, runtime_checkable
 from pika.adapters.utils import io_services_utils, nbio_interface
 from pika.adapters.utils.io_services_utils import check_callback_arg, check_fd_arg
 
@@ -28,7 +28,8 @@ class AbstractSelectorIOLoop(Protocol):
 
     A `Protocol` rather than a base class so that loops implementing this
     interface without deriving from it, such as `tornado.ioloop.IOLoop`, satisfy
-    it structurally.
+    it structurally. `Protocol` reached the stdlib in 3.8, so on 3.7 this falls
+    back to a plain base class; see `pika._utils`.
     """
 
     # The values of the I/O loop's READ/WRITE/ERROR flags, which may be used
--- a/pika/adapters/utils/selector_ioloop_adapter.py
+++ b/pika/adapters/utils/selector_ioloop_adapter.py
@@ -4,11 +4,10 @@ as tornado's and our home-grown select_connection's I/O loops.
 
 from __future__ import annotations
 
-import abc
 import logging
 import socket
 import threading
-from typing import Any, Callable
+from typing import Any, Callable, Protocol, runtime_checkable
 
 from pika._utils import override
 from pika.adapters.utils import io_services_utils, nbio_interface
@@ -17,7 +16,8 @@ from pika.adapters.utils.io_services_utils import check_callback_arg, check_fd_a
 LOGGER = logging.getLogger(__name__)
 
 
-class AbstractSelectorIOLoop:
+@runtime_checkable
+class AbstractSelectorIOLoop(Protocol):
     """
     Selector-based I/O loop interface expected by
     `selector_ioloop_adapter.SelectorIOServicesAdapter`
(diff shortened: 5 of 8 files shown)


alonfaraj (commented):
## The `READ`/`WRITE`/`ERROR` comment claims something mypy rejects

`selector_ioloop_adapter.py:35-37` says the flags are attributes "so that implementations may use either". A bare annotation on a `Protocol` is a *settable* variable, so a read-only property does not satisfy it:

```python
class Loop(Protocol):
    READ: int

class PropLoop:
    @property
    def READ(self) -> int:
        return 1

def use(loop: Loop) -> None: ...

use(PropLoop())
```

```
error: Argument 1 to "use" has incompatible type "PropLoop"; expected "Loop"  [arg-type]
note: Protocol member Loop.READ expected settable variable, got read-only attribute
```

An explicit subclass fails too, with `Cannot override writeable attribute with read-only property`.

That is a narrowing, not just a stale comment: the declaration being replaced was `@property def READ`, and its docstring made attributes the shortcut, not the requirement. A third-party loop using properties satisfies `main` and not this branch.

Both in-tree loops use ints, so `READ: int` is fine. Suggested:

```python
    # The values of the I/O loop's READ/WRITE/ERROR flags, which may be used
    # with bitwise operators as expected. Protocol variable members are not
    # satisfied by read-only properties, so implementations must expose plain
    # `int` attributes.
```

## Optional: `@abc.abstractmethod` would no longer be inert under `Protocol`

Not a blocker, and the commit's conclusion may well still be the right one. But the reasoning behind it does not survive the conversion. "Already inert" held for an `object` base; under `Protocol` it does not, because `_ProtocolMeta` derives from `ABCMeta`, so keeping the decorators populates `__abstractmethods__` for real.

```python
@runtime_checkable
class P(Protocol):
    @abc.abstractmethod
    def start(self) -> None: """doc"""
    @abc.abstractmethod
    def stop(self) -> None: """doc"""

class Partial(P):
    def start(self) -> None: ...

Partial()              # TypeError: ... without an implementation for abstract method 'stop'
isinstance(duck, P)    # still True, structural conformance unaffected
```

I checked this against the real class rather than a toy: re-adding the nine decorators keeps `typecheck` clean and `hatch run unit` green, `IOLoop()` still instantiates, `isinstance(tornado.ioloop.IOLoop.current(), AbstractSelectorIOLoop)` is still `True`, and a subclass missing `remove_handler` raises at construction. As merged, `__abstractmethods__` is `frozenset()`.

Two reasons this is minor rather than a request for change: mypy already catches a missing method either way (`implicitly abstract because it has an empty function body`), so the only gain is runtime.

lukebakken:
Thanks @alonfaraj - both points were right, and the second one was right in a way I had not considered. Pushed as a third commit, plus a rewrite of the first commit's message and the PR description, which repeated the same errors.

## The flags comment

Confirmed, and it was a narrowing rather than a stale comment, exactly as you described. I took your suggested wording.

Before settling on it I tested the alternative, since read-only property *members* would be the genuinely more permissive declaration and would preserve what `main` accepted. Under mypy that works. pyright does not agree: 6 diagnostics on the four affected files becomes 18, and the 12 additions are `reportAssignmentType` x6 plus `reportIncompatibleMethodOverride` x6, which is every diagnostic the six retired suppressions were covering. Since both in-tree loops use plain ints, and so does tornado's `IOLoop` (`READ` 1, `WRITE` 4, `ERROR` 24), the narrower declaration costs nothing in the tree and keeps pyright at parity, which matters for #1564.

That reasoning is now in the comment so the trade is not re-litigated later.

## `@abc.abstractmethod`

You are right that "already inert" does not survive the conversion, and I had verified it against the pre-conversion class and then restated it as if it still held. `_ProtocolMeta` derives from `ABCMeta`, so keeping the decorators populates `__abstractmethods__` for real.

The decorators still stay out, but for a reason neither of us had stated: pika supports 3.7, where `typing.Protocol` does not exist. The second commit on this branch falls back to `object` there. Measured on both interpreters:

| | 3.14.5 | 3.7.17 |
|---|---|---|
| `__abstractmethods__` with decorators | all 12 populated | attribute absent |
| incomplete subclass construction | `TypeError` | succeeds |

So keeping them would enforce construction on 3.8 through 3.14 and do nothing on 3.7, giving different runtime behavior across legs of the same matrix. mypy reports the missing method on every version either way. The class docstring now says this instead of the old claim.

## The reason both slipped through

No test referenced `AbstractSelectorIOLoop` at all, even though structural conformance is the entire premise of the PR. The new `tests/unit/selector_ioloop_adapter_tests.py` has 11 tests, asserted against classes rather than instances so no loop, hub, or fd is created, with protocol method names derived from the class at import time so a new member is immediately required of all three loops.

I checked the tests actually bite by mutating the source four ways:

| Mutation | Caught by |
|---|---|
| Protocol flags as properties | `test_flags_are_declared_as_variable_members` |
| `select_connection.IOLoop` flags as properties | `test_flags_are_int_attributes` |
| New protocol method no loop implements | `test_implements_every_protocol_method` and the `isinstance` test |
| `@abc.abstractmethod` re-added | `test_carries_no_abstract_methods` |

## Corrections to the original description

Three claims were wrong. The first commit's message was reworded and the PR body rewritten:

- "pyright reports the same 10 diagnostics before and after" - the count is 6. Parity held, the number was carried from an earlier measurement instead of re-run.
- "the decorators were already inert" - your point 2.
- "8 `@abc.abstractmethod` decorators removed" - it is 12, three on the flags and nine on the methods. Found while re-checking the rest.

The code was fine in all three cases; the prose describing it was not, and that was the only part no check validates. Verification at the current tip: `lint-check`, `fmt-check`, `docfmt-check`, `typecheck` all exit 0; mypy clean under the default platform plus `--platform darwin` and `--platform win32`; `hatch run test` 1473 passed and 61 skipped against a live broker; 3.7.17 unit suite 912 passed and 24 skipped, where the two `isinstance` tests skip by design; pyright 6 diagnostics, unchanged from `main`.

alonfaraj (commented):
### `test_implements_every_protocol_method` is vacuous for the two in-tree loops

`getattr(self.loop_class, name)` resolves through the MRO, and both in-tree loops derive from the protocol, so a missing method finds the protocol's own do-nothing body rather than being absent. Add a member no loop implements:

```diff
+    def brand_new_member(self) -> None:
+        """Implemented by no loop."""
+
     def close(self) -> None:
```

```console
$ hatch run pytest tests/unit/selector_ioloop_adapter_tests.py -n 0 -q
FAILED ...::TornadoIOLoopConformanceTests::test_implements_every_protocol_method
FAILED ...::AbstractSelectorIOLoopTests::test_isinstance_accepts_a_loop_that_does_not_derive_from_it
2 failed, 9 passed

IOLoop                   callable=True  defined_by=AbstractSelectorIOLoop
_GeventSelectorIOLoop    callable=True  defined_by=AbstractSelectorIOLoop
IOLoop (tornado)         callable=False defined_by=None
```

Only tornado is genuinely checked, so "immediately required of `select_connection.IOLoop`, `_GeventSelectorIOLoop`, and `tornado.ioloop.IOLoop`" holds for tornado alone. Worth closing, since with the decorators gone an inherited stub is exactly the failure mode: a forgotten `remove_handler` would not raise, the descriptor would just never be removed.

Asking which class defined the member:

```python
        for name in PROTOCOL_METHOD_NAMES:
            owner = next((klass for klass in self.loop_class.__mro__
                          if name in vars(klass)), None)
            self.assertIsNotNone(owner, f'{name} is missing')
            self.assertIsNot(owner, AbstractSelectorIOLoop,
                             f'{name} is only the inherited protocol stub')
            self.assertTrue(callable(getattr(self.loop_class, name)),
                            f'{name} is not callable')
```

lukebakken:
Confirmed, and fixed in 77a87354.

Reproduced it exactly. Both in-tree loops derive from `AbstractSelectorIOLoop`, so `getattr(loop_class, name)` resolves a method the loop never defines through the MRO to the protocol's own body. Adding a member with a body that no loop implements passed vacuously for `select_connection.IOLoop` and `_GeventSelectorIOLoop`, and failed only for tornado's `IOLoop`, which does not derive. So "immediately required of all three loops" held for tornado alone, exactly as your `defined_by` output shows. And with the decorators gone, an inherited stub is the failure the test is supposed to catch: a forgotten `remove_handler` would resolve to the do-nothing body rather than raising.

I took your suggested check: walk `__mro__` for the class that defines each name and reject `AbstractSelectorIOLoop` itself, so an inherited stub fails the same as an absent method.

To confirm it bites, I added a bodied member to the protocol that no loop implements. Under the old test that failed on tornado only; under the new one it fails on all three:

```console
$ hatch run pytest tests/unit/selector_ioloop_adapter_tests.py -n 0 -q
FAILED ...::SelectConnectionIOLoopConformanceTests::test_implements_every_protocol_method
FAILED ...::GeventSelectorIOLoopConformanceTests::test_implements_every_protocol_method
FAILED ...::TornadoIOLoopConformanceTests::test_implements_every_protocol_method
FAILED ...::AbstractSelectorIOLoopTests::test_isinstance_accepts_a_loop_that_does_not_derive_from_it
4 failed, 7 passed
```

With the mutation reverted, at the pushed tip: `lint-check`, `fmt-check`, `docfmt-check`, `typecheck` all exit 0; `hatch run test` 1473 passed, 61 skipped against a live broker; 3.7.17 unit suite 912 passed, 24 skipped, the 2 `isinstance` tests skipping by design.

alonfaraj (approved):

PR #1672, opened 2026-08-03
Drop the six tractable Twisted `type: ignore` comments

lukebakken:
## Summary

Third of three PRs for #1618, covering item 3: the tractable 6 of the 9 Twisted suppressions. Based on `main` rather than stacked on #1670 or #1671, so it can be reviewed in any order.

| Code | Count | Change |
|------|-------|--------|
| `[func-returns-value]` | 2 | `DeferredQueue.put()` and `Channel.basic_publish()` both return nothing; call them as statements |
| `[arg-type]` | 2 | `UnroutableError` / `NackError` accept a structural type covering the three members their callers read |
| `[assignment]` | 2 | Narrowing fixes in `_on_connection_closed()` |

Repo-wide `type: ignore` count drops from 15 to 9, measured against `main` at 848adc29 with #1670 merged. The 3 that remain in `twisted_connection.py` are the two `ITransport` cases and `put`'s `[override]`, all deferred to #1669. With #1671 also merged the tree reaches the 7 tracked there.

All six marked genuine mismatches between the code and its annotations rather than gaps in mypy, so each is fixed at the code rather than silenced.

## `[func-returns-value]`

Both callees return nothing, so using the result was meaningless.

- `defer.DeferredQueue.put()` is annotated `-> None` and its body has no `return` statement, in both twisted 26.4.0 and the 22.10.0 that a 3.7 environment resolves.
- `pika.channel.Channel.basic_publish()` is annotated `-> None` and likewise has no `return`.

`ClosableDeferredQueue.put()` keeps its `Deferred[Any] | None` return type. The optionality is load-bearing, not leftover: the closed path returns `defer.fail(...)`, and `tests/acceptance/twisted_adapter_tests.py:67` asserts on that Deferred.

## `[arg-type]`

`UnroutableError` and `NackError` were annotated `Sequence[ReturnedMessage]`, but the Twisted adapter passes `ReceivedMessage`. The two are unrelated: `ReturnedMessage` is a slotted class with `method`, `properties`, and `body`, while `ReceivedMessage` is a `namedtuple` with those three plus `channel`.

#1618 proposed widening the annotation to accept both. A literal union turns out not to be possible: naming `ReceivedMessage` in `exceptions.py` means `exceptions` imports `twisted_connection`, which reaches `pika.connection` and back to `exceptions`. A `TYPE_CHECKING`-only import does not help either, because `pika.data` imports `exceptions` at runtime and `pika.spec` imports `pika.data`.

So `__init__` accepts `Sequence[ReturnedMessageLike]`, a structural type covering the three members a reader of `.messages` uses. Three details are not obvious and were each established by running mypy rather than reasoning:

- The members have to be **read-only properties**. Declared as plain variables, mypy rejects `ReceivedMessage` with `Protocol member expected settable variable, got read-only attribute`, because namedtuple fields are read-only.
- The **stored attributes are annotated `Sequence[Any]`**, not the protocol. Letting them infer from the parameter regressed downstream code: a protocol is not assignable to the nominal class it structurally describes, so a caller annotating a handler against `ReturnedMessage` started failing. A three-function reproduction is clean on 848adc29 and produces 3 `[arg-type]` errors without this. `Sequence[Any]` is the only annotation that satisfies all four constraints at once: both message classes accepted at the constructor, downstream `Sequence[ReturnedMessage]` annotations still checking, downstream protocol-annotated code checking, and pika's own mypy run clean. Pinning to `Sequence[ReturnedMessage]` instead just relocates the error inside `exceptions.py` as `[assignment]`. Reported by @alonfaraj.
- The name is **public**. An earlier revision used `_ReturnedMessageLike` to keep it out of the rendered nav, but downstream code should be able to name the type it is handed, so it is public and does render: 4 entries in the exceptions reference TOC. No `filters` entry added, since a type callers annotate against belongs in the reference.

The type is declared under `TYPE_CHECKING`, so `typing.Protocol` is never evaluated at runtime and this PR does not depend on the `pika._utils` version gate added in #1671. That also means an unguarded runtime import raises `ImportError`; consumers import it inside a `TYPE_CHECKING` block, which `from __future__ import annotations` or a quoted annotation makes usable. Confirmed on 3.7.17: `hasattr(pika.exceptions, 'ReturnedMessageLike')` is `False`, and both exceptions still construct and read `messages[0].body` from a real `ReceivedMessage`.

`ReturnedMessageLikeTests` in `tests/unit/exceptions_test.py` pins the three members on both concrete classes, passes each through both exceptions, and asserts the protocol stays absent from the module at runtime.

## `[assignment]`

Both suppressions in `_on_connection_closed()` came from reusing a name, not from anything mypy got wrong.

The first: the method opens with `for d in self._calls`, and `self._calls` is `set[Deferred[Any]]`, so `d` was already fixed as non-optional by the time `d, self.closed = self.closed, None` assigned the optional `self.closed` to it. Giving the loop its own name removes the conflict. The same tuple-swap against `self.ready` appears three times in this class and has never needed a suppression, because those methods have no loop above them.

The second: unwrapping a `Failure` by rebinding the `exception` parameter. Branching on the `isinstance` check instead means nothing is rebound.

That branch also required naming `Failure` in the signature. `twisted.python.failure.Failure` derives from `BaseException`, not `Exception` (`issubclass(Failure, Exception)` is `False`), so under the old `exception: Exception` annotation the `isinstance(exception, Failure)` branch was unreachable as far as a type checker was concerned. In production it is unreachable in fact too: `_TwistedConnectionAdapter.connection_lost()` unwraps the `Failure` at `twisted_connection.py:1154` before `_on_stream_terminated()`, and `Connection._error` is `Exception | None`. But `test_on_connection_closed_Failure` calls the method with a `Failure` directly and asserts the callback path fires, so the branch is contractually live and the annotation is what was wrong. This is a pre-existing inconsistency, not one this PR introduces.

## Behavior changes

None intended. The two statement conversions discard a value that was always `None`; `basic_publish()` without delivery confirmation still returns an already-fired Deferred, now explicitly `defer.succeed(None)` instead of `defer.succeed(result)` where `result` was `None`. The `[arg-type]` change is annotation-only. The `[assignment]` changes rename a local and restructure an `if` without altering which value reaches `d.callback()`.

## Testing

- `hatch run docs:build` exits 0.
- `hatch run test`: 1467 passed, 61 skipped against a live RabbitMQ broker, including all 67 Twisted acceptance tests and 26 covering the unroutable and nack paths.
- Python 3.7.17, with twisted 22.10.0 installed to match what CI's hatch environment provides: 975 passed, 22 skipped across the unit suite and the Twisted acceptance tests. Run because this PR touches `exceptions.py` and #1671 established that no local check catches a 3.7 stdlib gap.
- The downstream reproductions are committed as `tests/typing/`, so they are a standing check rather than a one-off: clean on `main`, 3 errors with the `Sequence[Any]` fix reverted, clean with it in place. See below.

## Typing fixtures

The `Sequence[Any]` annotation above was found by @alonfaraj, not by any check in this repo, and nothing here would have caught it coming back. `mypy.ini` sets `packages = pika`, so `hatch run typecheck` only ever reads pika's own source; nothing in the tree consumes `UnroutableError.messages` or `NackError.messages` from the outside. Reverting the fix leaves both the unit suite and the type check green.

`tests/typing/` closes that gap. Each file is a small downstream consumer that `mypy` checks as ordinary source, so `typecheck` now runs `python -m mypy pika/ tests/typing/`.

- `returned_messages.py` reads `.messages` the four ways callers do: annotated against the concrete `ReturnedMessage`, collected into a concrete list, forwarded to a concrete callback, and annotated against the exported protocol.
- `rejects_wrong_types.py` pins the constructor constraint with `# type: ignore[list-item]`. `mypy.ini` already sets `warn_unused_ignores = True`, so if the parameter stops being constrained the ignore becomes unused and the run still fails.

`pytest` collects nothing from the directory: the files carry no test functions and their names match none of the `python_files` patterns.

#1671 adds a fixture of its own for the same reason, and both branches define `typecheck` identically, so `pyproject.toml`, `AGENTS.md` and `.github/workflows/mypy.yaml` are byte-identical between the two and merge cleanly. One file does collide: both add `tests/typing/README.md`, each citing its own fixture as the motivating example, so a test-merge reports `CONFLICT (add/add)` there. Whichever of the two lands second needs that README resolved by keeping the shared framing, carrying both example paragraphs, and keeping the negative-fixture sentence below. No other path conflicts, and neither branch conflicts with #1674.

## Incidental finding

`docformatter` drops the space after `:returns:` unless some preceding field in the same docstring is itself wrapped across lines. It reformats `:returns: ``True`` if ...` to `:returns:``True`` if ...`, which is why the existing occurrence in `channel.py:229` passes and a new one on its own does not. Worked around by rewording the new `:returns:` line rather than changing the tool config. Pre-existing, and not addressed here.

See #1618


alonfaraj (commented):
## 1. Make the structural type public

`pika/exceptions.py:278` and `pika/exceptions.py:299`

`UnroutableError.messages` and `NackError.messages` now infer as `Sequence[_ReturnedMessageLike]`. That type is private and declared only under `TYPE_CHECKING`, so it is neither importable at runtime nor advertised as something to annotate against. Downstream code that type-checks against the blocking adapter breaks on upgrade:

```python
from typing import Sequence
from pika.adapters.blocking_connection import ReturnedMessage
from pika.exceptions import UnroutableError

def log_unroutable(messages: Sequence[ReturnedMessage]) -> None:
    for m in messages:
        LOGGER.warning('dropped %s -> %s', m.method.exchange, m.body)

try:
    channel.basic_publish(exchange='x', routing_key='k', body=b'...', mandatory=True)
except UnroutableError as err:
    log_unroutable(err.messages)
```

```
error: Argument 1 to "log_unroutable" has incompatible type
  "Sequence[_ReturnedMessageLike]"; expected "Sequence[ReturnedMessage]"  [arg-type]
```

`ReturnedMessage` is a nominal class, so the protocol is not assignable to it, and the same shape hits anything that stores the messages (`list[ReturnedMessage]`) or passes them to a callback typed `Callable[[Sequence[ReturnedMessage]], None]`. `UnroutableError`'s own docstring says "Used by BlockingChannel", so this lands on the most common consumer.

Suggested fix: rename to `ReturnedMessageLike`. If it should stay out of the rendered nav, add an explicit `filters` entry to `docs/modules/exceptions.md` rather than relying on the underscore prefix. Hiding it from the docs is the smaller concern of the two.

## 2. Correct the protocol docstring

`pika/exceptions.py:14-25`

The docstring says the exceptions "accept anything with the three members they read", but neither exception reads `method`, `properties`, or `body`. Both read only `len(messages)` and store the sequence (`exceptions.py:282-288` and `exceptions.py:303-309`).

The protocol's actual purpose is to describe what consumers of `.messages` can rely on, which is worth stating directly since that is also the argument for point 1.

lukebakken:
Both points confirmed and fixed in d5a94e77.

## 1

The regression reproduces exactly as described. A three-function version of your snippet (callback arg, `list[ReturnedMessage]`, `Callable[[Sequence[ReturnedMessage]], None]`) is clean against main at 848adc29 and produces 3 `[arg-type]` errors against the branch.

The rename alone does not fix it, though. I applied it to the branch tree and changed nothing else: all 3 errors remain verbatim, only with the leading underscore dropped from the message. mypy is objecting to protocol-versus-nominal assignability, not to the name, and no name makes a protocol assignable to `ReturnedMessage`.

What fixes it is annotating the stored attributes `Sequence[Any]` and keeping the protocol on the `__init__` parameters. Callers annotating against either concrete class check again, callers annotating against the protocol check too, and the constructors still accept only message-shaped sequences. I also tested pinning the attribute to `Sequence[ReturnedMessage]`, which just moves the error inside `exceptions.py` as `[assignment]`.

The rename is in the commit regardless, since downstream code should be able to name the type it is handed. That does put it in the rendered nav (4 TOC entries where the underscore produced none), and I have left it there rather than adding a `filters` entry: a type callers are meant to annotate against belongs in the reference. The PR body claimed the opposite and is now corrected.

## 2

Correct, neither exception reads `method`, `properties`, or `body`. The docstring now says the members describe what a reader of `.messages` can rely on, and records that the declaration lives under `TYPE_CHECKING`, so an unguarded runtime import raises.

Also added `ReturnedMessageLikeTests`: pins the three members on both `ReturnedMessage` and `ReceivedMessage`, passes each through both exceptions, and asserts the protocol stays absent from the module at runtime.

PR #1673, opened 2026-08-03
Fail the build when a Codecov upload is rejected

alonfaraj:
## Proposed Changes

`CODECOV_TOKEN` is empty, so `codecov-action` uploads tokenless. Codecov accepts that only for fork pull requests; pushes to `main` and same-repo pull requests are rejected. From #1670's coverage job ([run 30754080482](https://github.com/pika/pika/actions/runs/30754080482/job/91514224219)):

```
collected 48 coverage reports
 -> Token length: 0
Upload queued for processing failed: {"message":"Token required because branch is protected"}
```

That job's conclusion is `success`, because `fail_ci_if_error: false` swallowed the rejection. 848adc29 and the three commits #1670 contributed are unknown to Codecov.

Codecov does show a report on `main` dated 2026-07-30, which is what hides this. It belongs to d98d61b5, the head commit of #1667 from `alonfaraj:gh-1661`: the tokenless path accepted it, and Codecov re-labeled the branch to `main` once the pull request merged. Every report `main` holds arrived that way, so the baseline now degrades one merge at a time as branches move from forks into this repository.

Two changes. Neither fixes the upload, which needs the token; they make its absence visible and stop unrelated flakiness from discarding reports.

- `fail_ci_if_error` is derived from the event: `true` for pushes and same-repo pull requests, `false` for fork pull requests, which cannot reach the secret and must keep using the tokenless path. A rejected upload goes red instead of silently green.
- The `coverage` job had no `if:`, so one failed matrix leg skipped the upload and that commit recorded nothing. It now runs unless the workflow was cancelled and uploads whatever artifacts arrived. With no artifacts at all it stays quiet if the test jobs failed, and goes red if they passed, because then the `coverage-*` download pattern is what is broken.

## Types of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

CI only. No library code is touched.

## Checklist

There is no harness for workflow files. Verified instead by parsing `main.yaml`, confirming both expressions resolve as intended, and running the guard step's shell logic against all four combinations of zero or more reports crossed with tests passing or failing.

## Fixes

- Fixes #

## Further Comments

Decisions for maintainers:

**Alternative to a long-lived secret.** `use_oidc: true` plus `permissions: {id-token: write}` avoids the token entirely, at the cost of enabling OIDC on the Codecov side.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -36,6 +36,9 @@ jobs:
     name: upload coverage to Codecov
     runs-on: ubuntu-latest
     needs: [test-modern, test-legacy]
+    # Upload whatever the matrix produced. A single flaky leg must not throw
+    # away the coverage the other forty-odd legs already measured.
+    if: ${{ !cancelled() }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Download coverage artifacts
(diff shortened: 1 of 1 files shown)


alonfaraj:
@lukebakken 
Any ideas why upload fail for main branch? 
is it related? https://docs.codecov.com/docs/codecov-tokens#uploading-without-a-token

lukebakken:
We can't fail builds due to codecov. I currently do not have the correct permissions on `pika/pika` to manage codecov. I will reach out to @gmr to see if I can be made an org admin.

PR #1674, opened 2026-08-03
Type-annotate the generated `spec.py` and drop its mypy exemption

alonfaraj:
## Proposed Changes

`pika/spec.py` is generated without type annotations, so every field read off a decoded frame resolves to `Unknown | Any | None`. Two consequences follow from that.

First, `mypy.ini` carried `[mypy-pika.spec] ignore_errors = True`, making the generated module the only part of the package mypy never looked at. Because the generated `__init__` methods were untyped, mypy saw every spec field as `Any` and reported the package clean while four genuine type errors sat in it:

```
pika/connection.py:1888  reply_code    int | None                              -> int
pika/connection.py:1890  reply_text    str | bytes                             -> str
pika/connection.py:2351  heartbeat     int | Callable[[Connection, float], int] | None -> int
pika/adapters/blocking_connection.py:1883  delivery_tag  int | None            -> int
```

Three of those four are already reported by pyright on `main` today. The fourth, `reply_code`, is suppressed there with a `pyright: ignore` comment, which mypy does not honour.

Second, one of them is a live bug rather than a typing nit. `Connection.Close.decode()` fills `reply_text` from `data.decode_short_string`, which returns raw bytes when the payload is not valid UTF-8. `ConnectionClosed` then coerced it with `str()`, which yields the bytes repr. A broker close carrying non-UTF-8 text reached the user as `b'forced \xff'` through both `reply_text` and `__repr__`, mangling the text needed to diagnose why the connection dropped.

This annotates the generated output, fixes the four sites it exposes, and drops the mypy exemption.

## Numbers

pyright, before and after:

| target | before | after |
| --- | --- | --- |
| `pika/spec.py` | 133 errors | **0** |
| `pika/` (whole package) | 159 errors | **23** |

The 136 difference is 133 from `spec.py` plus the 3 pre-existing errors at the call sites above.

mypy reports success both before and after, but afterwards it checks all 31 source files including `pika/spec.py`, which it previously skipped entirely.

## What the generator now emits

- Each resolved AMQP domain maps to an annotation. `shortstr` and `longstr` admit bytes, matching what `data.decode_short_string` yields for a payload that is not valid UTF-8. A field defaulting to None is annotated optional, which also covers properties fields, since an absent flag decodes as None.
- `synchronous` becomes a class attribute rather than a property, so it overrides `amqp_object.Method.synchronous` in kind. Its only reader, `channel.py:1451`, accesses it on an instance, where the two are equivalent.
- `decode`, `encode`, the module-level lookup dicts and `has_content` are annotated. `props` names the concrete properties classes rather than the base, so `frame.Header` keeps the specific type it accepts and the per-message header path needs no cast or isinstance check.
- 132 generated `decode` and `encode` methods carry `@override`, which `enable_error_code = explicit-override` requires once the module is checked. Only the method classes get it; `amqp_object.Properties` declares no decode or encode to override.
- `from __future__ import annotations`, required for the `X | None` syntax on the 3.7 to 3.9 legacy matrix. No annotation is evaluated at runtime, and the `override` shim in `pika/_utils.py` is dependency-free with a no-op fallback below 3.12.

`pika/data.py` is also touched: `encode_short_string` passes its value through `as_bytes`, which takes `str | bytes`, and `encode_table` encodes its keys the same way. Both were annotated to accept only `str`, narrower than they behave. Widening them is an annotation fix with no behavior change.

## Impact on downstream type checking

pika ships `py.typed`, so this changes a public contract. Measured against typed user code doing the usual things (`basic_ack(delivery_tag=method.delivery_tag)`, `method.routing_key.upper()`, passing `method.exchange` back into `basic_publish`):

| checker | on `main` | with this change |
| --- | --- | --- |
| pyright | 6 errors | 6 errors, same sites, clearer messages |
| mypy | 0 errors | **6 errors** |

pyright users are unaffected, because those Optionals are already visible to it. mypy users go from clean to 6 errors on idiomatic code, because mypy had been treating the untyped module as `Any`.

Every one of those errors is true: `Basic.Deliver.delivery_tag` really is `int | None` under the generated constructor, and `routing_key` really can be `bytes`. But it will break downstream type-checking CI on upgrade.

## Types of Changes

- [x] Bugfix 
- [ ] New feature
- [x] Breaking change 
- [ ] Documentation
- [ ] Cosmetics

## Checklist

Full suite against a local RabbitMQ: 1464 passed, 61 skipped. `hatch run typecheck`, `fmt-check`, `docfmt-check`, `lint-check` and `docs:build` all pass. Every one of the seven commits passes all five gates individually, so the branch bisects cleanly.

Two tests cover the reply-text fix, in `tests/unit/exceptions_test.py`. Both fail without it, with the mangling described above:

```
E   - b'forced \xff'
E   + forced <replacement char>
```

Documentation added to `utils/codegen.py`: the regeneration recipe omitted the formatting step, without which the result differs from the committed file by line wrapping alone.

## Fixes

No linked issue.

## Further Comments

**Alternatives considered.**

`decode_short_string` could return `str` unconditionally, decoding with `errors='replace'`. That would remove the `str | bytes` union from every queue name, routing key and consumer tag, and make the `reply_text` handling here redundant. I prototyped it: roughly 14 hand-written lines across three files, fully green. I left it out because it reverses a deliberate and tested behavior (`test_decode_short_string_invalid_utf8` asserts bytes), it leaves an unresolved asymmetry with long strings, which keep their bytes fallback in `decode_value`, and measurably it does not reduce the downstream error count above, since that is driven by Optionality rather than by the bytes union.

Annotating method fields as non-Optional would reduce the downstream count, since a method class's `decode()` sets every field unconditionally. It is left out because it forces the constructor defaults to change from None to type-appropriate zeros, and today `spec.Connection.Close()` followed by `.encode()` raises `struct.error` on the None, which is arguably the correct fail-fast behavior. Defaulting to `0` would instead put a silently wrong `reply_code` on the wire. That trade is an API decision worth its own discussion.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -15,6 +15,46 @@ ch.basic_publish(exchange='', routing_key='q', body=b'hello')
 
 The other connection adapters (`BlockingConnection`, `SelectConnection`, `AsyncioConnection`, etc.) are **not** thread-safe. Each connection instance is confined to the thread that created it. The only safe cross-thread operation on these adapters is calling `add_callback_threadsafe` to schedule a callback in the connection's IOLoop thread. See [connection adapters](modules/adapters/index.md) for details.
 
+### Why does mypy report new errors on my consumer callback after upgrading?
+
+Pika ships a `py.typed` marker, so a type checker reads pika's own annotations. `pika.spec` is generated code, and it was previously excluded from pika's `mypy` configuration, which left every field of a decoded frame typed as `Any`. `Any` silences all checking, so code like this passed:
+
+```python
+def on_message(ch, method, properties, body):
+    ch.basic_ack(delivery_tag=method.delivery_tag)
+```
+
+`pika.spec` is now annotated and checked, so the same line reports:
+
+```
+error: Argument "delivery_tag" to "basic_ack" of "BlockingChannel"
+has incompatible type "int | None"; expected "int"
+```
+
+The errors are accurate rather than newly introduced behavior. Two properties of the generated code cause them:
+
+- **Fields are optional.** Every generated constructor defaults its arguments to `None`, so `delivery_tag` is `int | None`. A frame decoded off the wire always carries the field, but the type cannot express that.
+- **String fields may be bytes.** AMQP `shortstr` is a length-prefixed byte string with no declared encoding. Pika decodes it as UTF-8 and falls back to the raw `bytes` when that fails, so `routing_key`, `exchange`, and `consumer_tag` are `str | bytes | None`.
+
+An `assert` narrows the optional away, and `isinstance` narrows the `bytes`:
+
+```python
+def on_message(ch: BlockingChannel, method: Basic.Deliver,
+               properties: BasicProperties, body: bytes) -> None:
+    assert method.delivery_tag is not None
+    ch.basic_ack(delivery_tag=method.delivery_tag)
+
+    routing_key = method.routing_key
+    assert isinstance(routing_key, str)
+    ch.basic_publish(exchange='', routing_key=routing_key, body=body)
+```
+
+Narrowing only the optional is not always enough for a string field, since `str | bytes` remains. Whether that matters depends on the operation: `routing_key.upper()` checks either way, because `bytes` also has `upper()`, while passing the value back into `basic_publish`, using it as a `dict[str, ...]` key, or concatenating it with a `str` all require the `isinstance` narrowing above.
+
+Use `typing.cast` instead of `assert` where the check should carry no runtime cost. Note that `assert` statements are stripped under `python -O`.
+
+`pyright` reported most of these already, so its users see little change.
+
 ### How do I report a bug with Pika?
 
 The [main Pika repository](https://github.com/pika/pika) is hosted on [GitHub](https://github.com), and we use the issue tracker at [github.com/pika/pika/issues](https://github.com/pika/pika/issues).
--- a/mypy.ini
+++ b/mypy.ini
@@ -12,5 +12,3 @@ strict_equality = True
 extra_checks = True
 enable_error_code = explicit-override
 
-[mypy-pika.spec]
-ignore_errors = True
--- a/pika/adapters/blocking_connection.py
+++ b/pika/adapters/blocking_connection.py
@@ -1889,9 +1889,12 @@ class BlockingChannel:
                         # NOTE: we can't use basic_nack with the multiple option
                         # to avoid nacking messages already held by our client.
                         for message in pending_messages:
-                            self._impl.basic_reject(
-                                message.method.delivery_tag, requeue=True
-                            )  # pyright: ignore[reportArgumentType]
+                            delivery_tag = message.method.delivery_tag
+                            # The generated spec types delivery_tag as
+                            # optional, but the broker always sets it on a
+                            # delivery.
+                            assert delivery_tag is not None
+                            self._impl.basic_reject(delivery_tag, requeue=True)
 
                 # Cancel the consumer; impl takes care of rejecting any
                 # additional deliveries that arrive for a auto_ack=False
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -1883,13 +1883,16 @@ class Connection(abc.ABC):
         """
         LOGGER.debug('_on_connection_close_from_broker: frame=%s', method_frame)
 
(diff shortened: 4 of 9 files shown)


lukebakken (approved):
## Approving

Verified at 7cdad36a. Everything below was checked by construction rather than by reading the diff, since a generated-code PR is exactly where a plausible-looking diff can hide a behavior change.

## The generated file reproduces byte for byte

Regenerated `pika/spec.py` from `utils/codegen.py` against a local `rabbitmq_codegen` checkout, then applied the `yapf --in-place --style google pika/spec.py` step that b9665c56 adds to the recipe. The result is byte-identical to the committed file. The new documentation of that step is worth having: the generator emits unformatted output, so without it the recipe does not actually reproduce what is in the tree.

## No logic changed

`pika/spec.py` is +588/-611, which is too large to eyeball. Normalized both revisions by stripping annotations, dropping `@override`, and rewriting `AnnAssign` to `Assign`, then diffed the resulting ASTs. The only differences are mechanical:

- 67 `list()` to `[]`
- 66 property to class attribute
- 3 import changes

Also compared runtime behavior directly rather than trusting the AST work: a frame round-trip over every method and properties class gives identical results on `main` and on this branch, 44 encode/decode successes and 22 expected failures on both.

## The three new asserts are on cold paths

`_on_connection_close_from_broker`, `_send_connection_tune_ok`, and the `stop_consuming` reject loop. None is per-message, so there is no hot-path cost. Each documents a real invariant that the generated type cannot express.

## `reply_text` fix

`ConnectionClosed.__init__` decoding `bytes` rather than letting `str()` produce a repr is a genuine improvement: `str(b'...')` would embed the `b` prefix and escapes into the very text a user reads to diagnose a close.

One clarification on scope, since I initially had this wrong myself. The bytes branch is only reachable for genuinely invalid UTF-8. I built real `Connection.Close` wire payloads to check: `decode_short_string` returns `str` for all valid UTF-8, so a normal `CONNECTION_FORCED` style reply text was never mangled. The fix is correct and worth keeping, it just applies to a narrower case than the diff might suggest.

## Downstream impact is real and now documented

Removing `[mypy-pika.spec] ignore_errors = True` is a `py.typed` contract change, not an internal cleanup. That exemption made every decoded frame field read as `Any` for downstream type checkers, which silenced all checking at that boundary.

Measured with one unchanged consumer file checked against both trees: 0 errors against `main`, 5 against this branch, including `basic_ack(delivery_tag=method.delivery_tag)`, the most common line in every pika tutorial. Confirmed mypy was resolving each tree's `pika` rather than an installed copy before trusting the 0.

This is the right change and nothing should be reverted, but it will turn downstream CI red on upgrade with no change on the consumer's end. 7cdad36a adds a `docs/faq.md` entry covering both causes, the optional defaults and the `shortstr` bytes fallback, with the narrowing idiom for each. Every snippet there was type-checked as written and the quoted error text is real mypy output. Tracked for release notes in #1675.

## Merge with current `main`

`main` has moved since this PR was opened, and the interaction is worth checking rather than assuming: #1672 added `tests/typing/` fixtures to the `typecheck` scope, while this PR removes the `pika.spec` exemption. Neither PR's CI has seen that combination.

Merged the two locally and ran the full set on the merged tree: `typecheck` clean over 33 source files, `ruff` clean, `yapf` clean, 911 unit tests passed. No rebase needed.

## CI

The one red leg, `test-legacy / macos-15-intel py3.8 tls=false`, is unrelated to this PR. Both failures are `TestZ_PublishAndConsumeBig` timing out at 13.8s and 14.4s against a fixed 15s budget.

Ruled out this PR as the cause rather than assuming flake, since a big-message test timing out is how an encoding regression would present and this PR touches `pika/data.py`. That file's diff is annotations and docstrings only, and benchmarking the actual 15.7 MB marshalling path gives 1.4 ms on both trees with byte-identical output. The failing leg ran about 27% slower overall than its `tls=true` sibling on identical test counts, 91.06s versus 71.67s, and the same two tests passed there in 3.5s and 2.6s. Runner speed against a budget with no headroom. Filed as a checklist item in #1675.

## Other checks

`hatch run lint-check`, `fmt-check`, `docfmt-check`, `typecheck`, and `docs:build` all exit 0. Also ran the suite on Python 3.8, since this PR touches `exceptions.py`.

Nice work, and thanks in particular for documenting the formatting step. That was a real reproducibility gap.

PR #1676, opened 2026-08-04
Perf: codec hot paths

alonfaraj:
## Proposed Changes

Speeds up the paths profiling flagged in a publish/consume workload. No API changes, bytes on the wire unchanged. Four fixes, each measured independently:

1. **`Frame._marshal` copied the payload three times.** It built `struct.pack(...) + payload + FRAME_END`; each `+` allocates and copies both operands. A single `b''.join` allocates once, so large bodies gain most.
2. **`struct.pack`/`unpack_from` took format strings**, re-resolving the format every call. Pre-compiled `struct.Struct` is ~40% faster per call, and these run once *per field* on every frame across `data.py`, `frame.py`, and the generated `spec.py`. Biggest contributor.
3. **`decode_frame` sliced every frame** to compare against `b'AMQP'`, allocating 4 bytes per frame to find a header that appears once per connection. No frame type shares the value of `b'A'`, so a one-byte check rules it out first.
4. **Dispatch ran 22 `isinstance` checks per delivered message** across `ContentFrameAssembler.process`, `_handle_content_frame`, `_process_frame`, and `_is_method_frame`, re-deriving what `frame_type` and `Method.INDEX` already store as ints.

Also: `encode_short_string` built a fresh `bytes((length,))` per string though the length is already validated to fit a byte, and `struct.pack('>c', b'F')` produced a one-byte constant.

## Types of Changes

- [ ] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics
- [x] Performance (no category fits exactly; no intended behavior change)

## Checklist

No new tests: a perf change has nothing to assert that would not be flaky on CI. What it needs is proof behavior did not change, below. The suite passes at **every commit**, so the branch bisects cleanly; all four `hatch` gates are clean.

### Worth reviewing closely

Commit 4 narrows a contract: with `frame_type` comparisons replacing `isinstance`, an object subclassing a frame class without the matching `frame_type` is no longer recognized. Real frames always set it in `__init__` from the matching `spec.FRAME_*` constant, so nothing changes on the wire.

It surfaced in `test_on_data_available`, which mocked `frame.Heartbeat` but hardcoded `frame_type` to 2 where a real one carries 8. It passed before only because a mock satisfies `isinstance` regardless of attributes. The fix pairs each mocked class with the frame type a real instance carries: a fixture correction, but please confirm you agree.

That commit also drops `assert not isinstance(frame_value, frame.ProtocolHeader)` for a `cast`. It was pure mypy narrowing and unreachable, since `ProtocolHeader` sets `frame_type` to -1 and the check above returns on negatives.

### Verifying nothing changed

A differential wire-format harness against `main`, **282 checks**: every field-table type, every frame type, `bytes` and `bytearray` inputs, non-zero offsets, and input truncated at *every* byte position. Encoded bytes and decoded values identical.

`spec.py` is regenerated from `utils/codegen.py` in the same commit, and the generator reproduces the committed file byte for byte beforehand, so its diff is only the intended substitutions.

### Benchmarks

Interleaved A/B against a `main` worktree, alternating order each round, min-of-rounds plus head-to-head wins. CPython 3.14.4, Linux, idle machine, 15 rounds. Milliseconds, lower is better.

| workload | main | branch | delta | wins |
|---|---:|---:|---:|---:|
| publish/64k | 0.921 | 0.690 | -25.1% | 15/15 |
| publish/headers | 9.646 | 8.032 | -16.7% | 15/15 |
| table/roundtrip | 76.318 | 63.995 | -16.1% | 15/15 |
| publish/small | 4.976 | 4.276 | -14.1% | 14/15 |
| decode/small | 4.415 | 3.886 | -12.0% | 15/15 |
| decode/headers | 7.707 | 6.886 | -10.7% | 12/15 |
| consume/headers | 11.758 | 10.543 | -10.3% | 11/15 |
| consume/small | 8.120 | 7.439 | -8.4% | 10/15 |
| ack | 5.683 | 5.302 | -6.7% | 10/15 |
| decode/64k | 0.384 | 0.358 | -6.6% | 12/15 |

The publish and table rows are unambiguous at 15/15. `consume/small` and `ack` at 10/15 are the weakest: their deltas are consistent in sign across runs, but the win counts do not establish them on their own.

End to end, 20k messages with headers against local RabbitMQ 4.3.4, 21 rounds:

| phase | main | branch | delta | wins |
|---|---:|---:|---:|---:|
| publish | 49,532 msg/s | 50,738 msg/s | -2.4% | 16/21 |
| consume | 142,078 msg/s | 140,032 msg/s | +1.5% | 9/21 |

Publishing gains a consistent 2-4%, since marshaling is the publish-path work. **Consuming shows nothing** (9/21 is a coin flip) despite decode being 7-9% faster in isolation: socket reads and broker round-trips dominate. An I/O-bound client should not expect much here. This matters when pika itself is the bottleneck: large bodies, heavy header tables, or consumers doing little work per message.

### Reproducing

Scripts are uncommitted since there is no benchmark directory today; Drop both in the repo root (they need `pika` importable from the script's own directory) and run:

```bash
python3 ab.py --base main --rounds 15
```

Use enough rounds on an idle machine.

<details>
<summary><code>bench.py</code></summary>

```python
"""CPU-bound benchmarks for pika's codec and frame-dispatch paths.

No broker required: workloads drive the marshal/decode paths directly and
through a Connection subclass whose adapter writes to a list.

    python3 bench.py                       # run all workloads
    python3 bench.py --save out.json       # record timings
    python3 bench.py --profile consume/small
"""
import argparse
import gc
import json
import statistics
import time
from unittest import mock

from pika import channel, connection, data, frame, spec

N_MSGS = 2000
SMALL_BODY = 256
BIG_BODY = 64 * 1024
HEADERS = {
    'x-request-id': 'ab3f9c1e-1122-4c8f-9a7e-77aa0011bb22',
    'x-retry-count': 3,
    'x-source': 'bench-producer',
    'x-flag': True,
    'x-nested': {
        'a': 1,
        'b': 'two'
    },
}


class FakeConnection(connection.Connection):
    """Connection that never touches a socket."""

    sink: list

    def _adapter_connect_stream(self):
        pass

    def _adapter_disconnect_stream(self):
        pass

    def _adapter_call_later(self, delay, callback):
        return None

    def _adapter_remove_timeout(self, timeout_id):
        pass

    def _adapter_add_callback_threadsafe(self, callback):
        pass

    def _adapter_emit_data(self, data):
        self.sink.append(data)


def make_connection():
    with mock.patch.object(connection.Connection, '_adapter_connect_stream'):
        conn = FakeConnection()
    conn.sink = []
    conn._set_connection_state(conn.CONNECTION_OPEN)
    return conn


def make_open_channel(conn, number=1):
    chan = channel.Channel(conn, number, lambda *a: None)
    chan._state = chan.OPEN
    conn._channels[number] = chan
    return chan


def build_delivery_stream(count, body_size, headers=None):
    """Raw wire bytes for `count` Deliver + Header + Body triplets."""
    out = bytearray()
    body = b'x' * body_size
    props = spec.BasicProperties(content_type='application/json',
                                 delivery_mode=2,
                                 app_id='bench',
                                 headers=headers)
    for i in range(count):
        method = spec.Basic.Deliver('ctag1.bench', i + 1, False,
                                    'bench-exchange', 'bench.routing.key')
        out += frame.Method(1, method).marshal()
        out += frame.Header(1, len(body), props).marshal()
        out += frame.Body(1, body).marshal()
    return bytes(out)


STREAM_SMALL = build_delivery_stream(N_MSGS, SMALL_BODY)
STREAM_HDRS = build_delivery_stream(N_MSGS, SMALL_BODY, HEADERS)
STREAM_BIG = build_delivery_stream(100, BIG_BODY)


def _decode_all(buf):
    offset, size, decode = 0, len(buf), frame.decode_frame
    while offset < size:
        consumed, _f = decode(buf, offset)
        offset += consumed


def _consume(stream):
    conn = make_connection()
    chan = make_open_channel(conn)
    chan._consumers['ctag1.bench'] = lambda *a: None
    conn._on_data_available(stream)


def _publish(count, body, props):
    conn = make_connection()
    chan = make_open_channel(conn)
    for _ in range(count):
        chan.basic_publish('bench-exchange', 'bench.routing.key', body, props)
        conn.sink.clear()


def w_decode_small():
    _decode_all(STREAM_SMALL)


def w_decode_headers():
    _decode_all(STREAM_HDRS)


def w_decode_big():
    _decode_all(STREAM_BIG)


def w_consume_small():
    _consume(STREAM_SMALL)


def w_consume_headers():
    _consume(STREAM_HDRS)


def w_publish_small():
    _publish(
        N_MSGS, b'x' * SMALL_BODY,
        spec.BasicProperties(content_type='application/json',
                             delivery_mode=2,
                             app_id='bench'))


def w_publish_headers():
    _publish(
        N_MSGS, b'x' * SMALL_BODY,
        spec.BasicProperties(content_type='application/json',
                             delivery_mode=2,
                             app_id='bench',
                             headers=HEADERS))


def w_publish_big():
    _publish(200, b'x' * BIG_BODY, spec.BasicProperties(delivery_mode=2))


def w_table_roundtrip():
    for _ in range(20000):
        pieces = []
        data.encode_table(pieces, HEADERS)
        data.decode_table(b''.join(pieces), 0)


def w_ack():
    conn = make_connection()
    chan = make_open_channel(conn)
    for i in range(N_MSGS * 5):
        chan.basic_ack(i + 1)
        conn.sink.clear()


WORKLOADS = {
    'decode/small': w_decode_small,
    'decode/headers': w_decode_headers,
    'decode/64k': w_decode_big,
    'consume/small': w_consume_small,
    'consume/headers': w_consume_headers,
    'publish/small': w_publish_small,
    'publish/headers': w_publish_headers,
    'publish/64k': w_publish_big,
    'table/roundtrip': w_table_roundtrip,
    'ack': w_ack,
}


def timeit(fn, repeats=7):
    times = []
    for _ in range(repeats):
        gc.collect()
        gc.disable()
        start = time.perf_counter()
        fn()
        times.append(time.perf_counter() - start)
        gc.enable()
    return min(times), statistics.median(times)


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument('--save')
    parser.add_argument('--profile')
    parser.add_argument('--only')
    args = parser.parse_args()

    if args.profile:
        import cProfile
        import pstats
        profiler = cProfile.Profile()
        profiler.enable()
        WORKLOADS[args.profile]()
        profiler.disable()
        pstats.Stats(profiler).sort_stats('tottime').print_stats(20)
        return

    selected = args.only.split(',') if args.only else list(WORKLOADS)
    results = {}
    for name in selected:
        fn = WORKLOADS[name]
        fn()  # warm up
        best, median = timeit(fn)
        results[name] = best
        print(f'{name:20s} min={best * 1000:9.3f} ms  '
              f'median={median * 1000:9.3f} ms')

    if args.save:
        with open(args.save, 'w') as handle:
            json.dump(results, handle, indent=2)


if __name__ == '__main__':
    main()
```

</details>

<details>
<summary><code>ab.py</code></summary>

```python
"""Interleaved A/B runner for bench.py.

Checks out a baseline revision into a temporary git worktree and runs
bench.py against both trees, alternating which goes first each round so
CPU frequency drift hits both variants equally. Reports min-of-rounds and
a head-to-head win count, so a small delta can be told apart from noise.

    python3 ab.py --base main --rounds 11

Single before/after runs on a laptop drift by more than 10%, which is
larger than several of the effects being measured; that is what this is
for.
"""
import argparse
import json
import os
import shutil
import statistics
import subprocess
import sys
import tempfile

HERE = os.path.dirname(os.path.abspath(__file__))
BENCH = os.path.join(HERE, 'bench.py')


def run(root, out):
    subprocess.run([sys.executable, BENCH, '--save', out],
                   env=dict(os.environ, PYTHONPATH=root),
                   check=True,
                   stdout=subprocess.DEVNULL)
    with open(out) as handle:
        return json.load(handle)


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument('--base', default='main', help='baseline revision')
    parser.add_argument('--repo', default=os.getcwd())
    parser.add_argument('--rounds', type=int, default=11)
    args = parser.parse_args()

    tmp = tempfile.mkdtemp(prefix='pika-ab-')
    baseline = os.path.join(tmp, 'base')
    subprocess.run(
        ['git', '-C', args.repo, 'worktree', 'add', '--detach', baseline,
         args.base],
        check=True,
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL)
    try:
        variants = {'base': baseline, 'new': args.repo}
        samples = {name: [] for name in variants}
        for round_index in range(args.rounds):
            order = list(variants)
            if round_index % 2:
                order.reverse()
            for name in order:
                out = os.path.join(tmp, f'{name}.json')
                samples[name].append(run(variants[name], out))
            print(f'round {round_index + 1}/{args.rounds}', file=sys.stderr)
    finally:
        subprocess.run(
            ['git', '-C', args.repo, 'worktree', 'remove', '--force',
             baseline],
            check=False)
        shutil.rmtree(tmp, ignore_errors=True)

    print(f'\n{"workload":20s} {"base":>9s} {"new":>9s} {"delta":>8s} '
          f'{"speedup":>8s}  wins')
    for workload in samples['base'][0]:
        base = [s[workload] for s in samples['base']]
        new = [s[workload] for s in samples['new']]
        best_base, best_new = min(base), min(new)
        wins = sum(1 for b, n in zip(base, new) if n < b)
        print(f'{workload:20s} {best_base * 1000:9.3f} {best_new * 1000:9.3f} '
              f'{(best_new - best_base) / best_base * 100:+7.1f}% '
              f'{best_base / best_new:7.2f}x  {wins}/{len(base)}')

    print('\nmedian-of-rounds:')
    for workload in samples['base'][0]:
        base = statistics.median([s[workload] for s in samples['base']])
        new = statistics.median([s[workload] for s in samples['new']])
        print(f'{workload:20s} {base * 1000:9.3f} {new * 1000:9.3f} '
              f'{(new - base) / base * 100:+7.1f}%')


if __name__ == '__main__':
    main()
```

</details>

### DIFF
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -8,7 +8,7 @@ import logging
 import uuid
 from collections import deque
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Callable, ClassVar, Union
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Union, cast
 
 from pika import exceptions, frame, spec, validators
 from pika._utils import as_bytes, override
@@ -1143,11 +1143,12 @@ class Channel:
             return
 
         if response:
-            if isinstance(response[0].method, spec.Basic.Deliver):
+            method_index = response[0].method.INDEX
+            if method_index == spec.Basic.Deliver.INDEX:
                 self._on_deliver(*response)
-            elif isinstance(response[0].method, spec.Basic.GetOk):
+            elif method_index == spec.Basic.GetOk.INDEX:
                 self._on_getok(*response)
-            elif isinstance(response[0].method, spec.Basic.Return):
+            elif method_index == spec.Basic.Return.INDEX:
                 self._on_return(*response)
 
     def _on_cancel(self, method_frame: frame.Method) -> None:
@@ -1591,17 +1592,20 @@ class ContentFrameAssembler:
 
         :param frame_value: The frame to process
         """
-        if (isinstance(frame_value, frame.Method) and
-                spec.has_content(frame_value.method.INDEX)):
-            self._method_frame = frame_value
-            return None
-        if isinstance(frame_value, frame.Header):
-            self._header_frame = frame_value
-            if frame_value.body_size == 0:
+        frame_type = frame_value.frame_type
+        if frame_type == spec.FRAME_METHOD:
+            method_frame = cast(frame.Method, frame_value)
+            if spec.has_content(method_frame.method.INDEX):
+                self._method_frame = method_frame
+                return None
+        elif frame_type == spec.FRAME_HEADER:
+            header_frame = cast(frame.Header, frame_value)
+            self._header_frame = header_frame
+            if header_frame.body_size == 0:
                 return self._finish()
             return None
-        if isinstance(frame_value, frame.Body):
-            return self._handle_body_frame(frame_value)
+        elif frame_type == spec.FRAME_BODY:
+            return self._handle_body_frame(cast(frame.Body, frame_value))
         raise exceptions.UnexpectedFrameError(frame_value)
 
     def _finish(self) -> tuple[frame.Method, frame.Header, bytes]:
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -8,7 +8,7 @@ import logging
 import uuid
 from collections import deque
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Callable, ClassVar, Union, cast
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Union
 
 from pika import exceptions, frame, spec, validators
 from pika._utils import as_bytes, override
@@ -1594,18 +1594,19 @@ class ContentFrameAssembler:
         """
         frame_type = frame_value.frame_type
         if frame_type == spec.FRAME_METHOD:
-            method_frame = cast(frame.Method, frame_value)
-            if spec.has_content(method_frame.method.INDEX):
-                self._method_frame = method_frame
+            assert isinstance(frame_value, frame.Method)
+            if spec.has_content(frame_value.method.INDEX):
+                self._method_frame = frame_value
                 return None
         elif frame_type == spec.FRAME_HEADER:
-            header_frame = cast(frame.Header, frame_value)
-            self._header_frame = header_frame
-            if header_frame.body_size == 0:
+            assert isinstance(frame_value, frame.Header)
+            self._header_frame = frame_value
+            if frame_value.body_size == 0:
                 return self._finish()
             return None
         elif frame_type == spec.FRAME_BODY:
-            return self._handle_body_frame(cast(frame.Body, frame_value))
+            assert isinstance(frame_value, frame.Body)
+            return self._handle_body_frame(frame_value)
         raise exceptions.UnexpectedFrameError(frame_value)
 
     def _finish(self) -> tuple[frame.Method, frame.Header, bytes]:
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -1739,7 +1739,7 @@ class Connection(abc.ABC):
 
         :param value: The frame to evaluate
         """
-        return isinstance(value, frame.Method)
+        return value.frame_type == spec.FRAME_METHOD
 
     def _is_protocol_header_frame(self, value: frame.Frame) -> bool:
         """
@@ -2244,12 +2244,13 @@ class Connection(abc.ABC):
         :param frame_value: The frame to process
         """
         # Will receive a frame type of -1 if protocol version mismatch
-        if frame_value.frame_type < 0:
+        frame_type = frame_value.frame_type
+        if frame_type < 0:
             return
 
         # ProtocolHeader (frame_type == -1) is handled above; narrow
         # the type for mypy.
-        assert not isinstance(frame_value, frame.ProtocolHeader)
+        frame_value = cast(frame.Frame, frame_value)
 
         # Keep track of how many frames have been read
         self.frames_received += 1
@@ -2259,7 +2260,7 @@ class Connection(abc.ABC):
             return
 
         # If a heartbeat is received, update the checker
-        if isinstance(frame_value, frame.Heartbeat):
+        if frame_type == spec.FRAME_HEARTBEAT:
             if self._heartbeat_checker:
                 self._heartbeat_checker.received()
             else:
--- a/pika/connection.py
+++ b/pika/connection.py
@@ -2250,7 +2250,7 @@ class Connection(abc.ABC):
 
         # ProtocolHeader (frame_type == -1) is handled above; narrow
         # the type for mypy.
-        frame_value = cast(frame.Frame, frame_value)
+        assert not isinstance(frame_value, frame.ProtocolHeader)
 
         # Keep track of how many frames have been read
         self.frames_received += 1
--- a/pika/data.py
+++ b/pika/data.py
@@ -9,7 +9,6 @@ from datetime import datetime, timezone
 from typing import Any
 
 from pika import exceptions
-from pika._utils import as_bytes
 
 # AMQP field-table value type tags, as integer byte values. decode_value
 # dispatches on the raw byte rather than a 1-byte slice to avoid a bytes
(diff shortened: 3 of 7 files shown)


lukebakken (approved):
Verified wire-format output is byte-identical to main across every field-table type, all method and properties classes, and truncation at every byte position. Control-flow parity holds, including the content-less Method fall-through to `UnexpectedFrameError`. Full matrix is green: ruff, docformatter, mypy on three OSes, CodeQL, docs, and the modern and legacy test grids with TLS on and off.

The follow-up commit shares the octet table, drops the hot-path `cast()` calls for `assert isinstance` narrowing, corrects the octet-table comment, and renames the signed `_PACK_SHORT` to avoid the cross-module name collision.

PR #1680, opened 2026-08-10
Fix: gh-1679 (Reject work scheduled on a closed ThreadSafeConnection)

### LINKED ISSUE #1679: `ThreadSafeConnection.add_callback_threadsafe` doesn't raise error when connection closed
When using `ThreadSafeConnection`, adding a callback on a closed connection, does not call the callback, and no error is raised.

```python
def hello():
    print("HELLO")

conn = pika.ThreadSafeConnection(pika.ConnectionParameters())
conn.close()
conn.add_callback_threadsafe(hello)
# Callback is not called, and no error is raised.
```

Compared to the traditional `BlockingConnection`, where an error is raised.

```python
conn = pika.BlockingConnection(pika.ConnectionParameters())
conn.close()
conn.add_callback_threadsafe(hello)
  File "<input>", line 1, in <module>
  File ".venv/lib/python3.14/site-packages/pika/adapters/blocking_connection.py", line 826, in add_callback_threadsafe
    raise exceptions.ConnectionWrongStateError('BlockingConnection.add_callback_threadsafe() called on closed or closing connection.')
pika.exceptions.ConnectionWrongStateError: BlockingConnection.add_callback_threadsafe() called on closed or closing connection.
```

I understand `ThreadSafeConnection` is not released yet, but I was trying it out since it looks like `BlockingConnection` is going to be deprecated.

Should `ThreadSafeConnection.add_callback_threadsafe` raise an error when the connection is closed? Or more specifically, when it knows the callback is not going to be called?
Accepting the callback but then silently never calling it doesn't seem like ideal behavior to me.

alonfaraj:
## Proposed Changes

`ThreadSafeConnection.add_callback_threadsafe` accepted work on a closed connection and then silently discarded it. The method delegated straight to the IOLoop:

```python
self._connection.ioloop.add_callback_threadsafe(callback)
```

Once the connection is closed the IOLoop thread has already exited, so the callback lands in a deque nobody drains. The caller got no error and the callback never ran, as reported in #1679:

```python
conn = pika.ThreadSafeConnection(pika.ConnectionParameters())
conn.close()
conn.add_callback_threadsafe(hello)  # accepted, never called, no error
```

Every other public method on the class already guards against this (`basic_publish`, `basic_ack` and friends via `ThreadSafeChannel._check_not_closed`, plus `channel()` and the blocked/unblocked registrars inline). `add_callback_threadsafe` was the one entry point that did not, which made it the only way to lose work without being told.

This adds a `ThreadSafeConnection._check_not_closed` helper and calls it at the top of `add_callback_threadsafe`. It raises in two cases:

- `_closed_reason` is set, so it re-raises that. This is deliberately the same exception the rest of the class raises once closed (for example `ConnectionClosedByClient`), rather than introducing a different error type for this one method.
- `_closed_reason` is still `None` but the underlying connection has reached the closed state, so it raises `ConnectionWrongStateError` with a message mirroring `BlockingConnection`'s. That covers the brief window on the IOLoop thread between the connection closing and `_on_connection_closed` recording the reason, and the force-close path where `_closed_reason` is set while the transport is still closing.

The helper names the calling method by frame introspection (`sys._getframe(1).f_code.co_name`) instead of taking a hand-passed string, so the message cannot drift from the method it labels. That lookup sits inside the error branch, so nothing on the publish hot path pays for it.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

## Fixes

- Fixes #1679

### Tests

Unit tests in `tests/unit/thread_safe_connection_tests.py`, one per branch of the guard:

- after `_on_connection_closed` fires, the recorded close reason is re-raised, and the callback is neither scheduled on the IOLoop nor invoked
- a closed connection with no recorded reason raises `ConnectionWrongStateError`, asserting the message names `add_callback_threadsafe` so the frame introspection is covered rather than just "something raised"

The pre-existing `test_add_callback_threadsafe_delegates` already covers the negative case, that the guard does not reject a healthy connection, so I did not duplicate it.

Acceptance test `TestAddCallbackThreadsafeAfterClose` in `tests/acceptance/thread_safe_connection_test.py` drives the reported scenario against a real broker: connect, `close()`, then `add_callback_threadsafe` must raise and must not run the callback. I confirmed all three tests fail without the fix.

### DIFF
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -17,7 +17,6 @@ from __future__ import annotations
 import itertools
 import logging
 import queue
-import sys
 import threading
 import time
 from threading import Event
@@ -471,7 +470,7 @@ class ThreadSafeChannel:
                 error[0] = exc
                 ready.set()
 
-        self._wrapper.add_callback_threadsafe(_invoke)
+        self._wrapper._schedule_unchecked(_invoke)
 
         try:
             if not ready.wait(timeout=timeout):
@@ -531,7 +530,7 @@ class ThreadSafeChannel:
                 if on_publish is not None:
                     on_publish(self._next_publish_seq_no)
 
-        self._wrapper.add_callback_threadsafe(_publish)
+        self._wrapper._schedule_unchecked(_publish)
 
     def basic_ack(self, delivery_tag: int = 0, multiple: bool = False) -> None:
         """
@@ -553,7 +552,7 @@ class ThreadSafeChannel:
                 LOGGER.warning('basic_ack failed (channel may have closed)',
                                exc_info=True)
 
-        self._wrapper.add_callback_threadsafe(_ack)
+        self._wrapper._schedule_unchecked(_ack)
 
     def basic_nack(self,
                    delivery_tag: int = 0,
@@ -580,7 +579,7 @@ class ThreadSafeChannel:
                 LOGGER.warning('basic_nack failed (channel may have closed)',
                                exc_info=True)
 
-        self._wrapper.add_callback_threadsafe(_nack)
+        self._wrapper._schedule_unchecked(_nack)
 
     def basic_reject(self, delivery_tag: int = 0, requeue: bool = True) -> None:
         """
@@ -602,7 +601,7 @@ class ThreadSafeChannel:
                 LOGGER.warning('basic_reject failed (channel may have closed)',
                                exc_info=True)
 
-        self._wrapper.add_callback_threadsafe(_reject)
+        self._wrapper._schedule_unchecked(_reject)
 
     def basic_qos(self,
                   prefetch_size: int = 0,
@@ -689,7 +688,7 @@ class ThreadSafeChannel:
                 error[0] = exc
                 ready.set()
 
-        self._wrapper.add_callback_threadsafe(_get)
+        self._wrapper._schedule_unchecked(_get)
 
         try:
             if not ready.wait(timeout=timeout):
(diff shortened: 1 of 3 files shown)


lukebakken:
Converted to draft as I'm making some changes. Thanks @alonfaraj!

lukebakken (approved):
## Approving

`add_callback_threadsafe` (and every other method) now rejects work scheduled on a closed `ThreadSafeConnection` instead of silently dropping it, and the closed-connection guard is handled consistently across all six raise sites.

Highlights:

- The guard stays off internal callers, so channel methods that register a waiter and schedule through `_schedule_unchecked` do not leak the waiter mid-teardown.
- Callers on the IOLoop thread are exempt, so a close callback that schedules follow-up work does not abort the rest of teardown.
- Re-raising the shared close reason no longer grows its traceback without bound. The close-time traceback is snapshotted and restored on every raise, which keeps the depth constant and stops rejected message bodies from being pinned alive. Verified by measurement: a 20k-iteration reject loop settles at a constant depth rather than accumulating tens of thousands of frames.
- All six raise sites re-raise `_closed_reason` itself (identity preserved), with a fresh `ConnectionWrongStateError` only in the brief window before the close reason is recorded.

Regression coverage added for the IOLoop-thread exemption, the guard bypass, all six raise sites, and the traceback helpers. Full CI is green across the 3.7-3.14 matrix on Linux, macOS, and Windows with TLS on and off, including the acceptance suite against a live broker.

---

Note: the changes on this branch were developed with AI assistance (Claude Code) and reviewed by me.

PR #1685, opened 2026-08-11
Fix: gh-1684 (Release each blocking RPC's channel callbacks)

### LINKED ISSUE #1684: Every blocking RPC on a `ThreadSafeChannel` leaks a channel close callback
`ThreadSafeChannel._blocking_rpc` registers a fresh `_on_chan_close` closure on the raw channel for every call, and never removes it:

https://github.com/pika/pika/blob/141f8457/pika/adapters/thread_safe_connection.py#L519

```python
def _on_chan_close(ch, reason) -> None:
    if not ready.is_set():
        ...

try:
    self._channel.add_on_close_callback(_on_chan_close)   # never removed
    channel_method(*args, **kwargs, callback=_on_ok)
```

`Channel.add_on_close_callback` registers with `one_shot=False` (channel.py:185), and each call passes a distinct closure, so `CallbackManager.add`'s duplicate check never collapses them. They accumulate for the life of the channel and are only freed when it closes.

`basic_get` does the same, and additionally leaks its `Basic.GetEmpty` one-shot whenever the get actually returns a message, because a one-shot is only consumed if it fires.

Affected: every blocking RPC, so `queue_declare`, `queue_bind`, `queue_delete`, `queue_purge`, `exchange_declare`, `exchange_bind`, `exchange_delete`, `basic_qos`, `basic_get`, `basic_consume`, `basic_cancel`, `confirm_delivery`, and `close`.

### This is a leak, not a correctness bug

The accumulated closures all guard on `if not ready.is_set()`, and for a completed RPC `ready` is already set, so firing them is a no-op. Nothing is mis-delivered or mis-acked. The cost is memory and CPU.

### Impact

Two effects, both proportional to how many RPCs the channel has served.

**Memory.** About 2 KB per RPC, held until the channel closes. A channel that serves 100k RPCs holds roughly 200 MB it will never release while open.

**CPU.** `CallbackManager.add` linearly rescans the existing list on every registration (callback.py:154). Since the closures never match, that scan is pure waste and grows without bound, making registration O(n^2) overall. Every blocking RPC pays it before a byte reaches the broker.

This matters specifically for the workload this adapter targets: a long-lived connection whose channels stay open for the process lifetime. A short-lived channel never notices.

### Reproduction

Needs RabbitMQ on localhost:5672.

```python
"""Every blocking RPC on a ThreadSafeChannel leaks a channel close callback."""

import time
import tracemalloc

import pika
from pika.adapters.thread_safe_connection import ThreadSafeConnection
from pika.callback import CallbackManager

N = 8000
WINDOW = 500
WARMUP = 500  # discarded, so connection warm-up cannot skew the first window


def registered_close_callbacks(ch):
    """How many _on_channel_close callbacks the raw channel is holding."""
    raw = ch._channel
    # CallbackManager normalizes the prefix to a string.
    stack = raw.callbacks._stack.get(str(raw.channel_number), {})
    return len(stack.get('_on_channel_close', []))


def measure_accumulation():
    """Drive real RPCs and report what the channel is left holding."""
    conn = ThreadSafeConnection(pika.ConnectionParameters('localhost'))
    ch = conn.channel()
    print(f'close callbacks before any RPC : {registered_close_callbacks(ch)}')

    for _ in range(WARMUP):
        ch.queue_declare(queue='leak-demo', durable=True)
    warm = registered_close_callbacks(ch)

    tracemalloc.start()
    base = tracemalloc.take_snapshot()

    timings = []
    for _ in range(N):
        start = time.perf_counter()
        ch.queue_declare(queue='leak-demo', durable=True)
        timings.append(time.perf_counter() - start)

    grown = sum(s.size_diff for s
                in tracemalloc.take_snapshot().compare_to(base, 'filename'))
    tracemalloc.stop()

    held = registered_close_callbacks(ch)
    first = sum(timings[:WINDOW]) / WINDOW * 1e3
    last = sum(timings[-WINDOW:]) / WINDOW * 1e3

    print(f'queue_declare calls issued     : {WARMUP} warmup + {N} measured')
    print(f'close callbacks after warmup   : {warm}')
    print(f'close callbacks at the end     : {held}')
    print(f'retained heap growth           : {grown / 1024:.0f} KiB')
    print()
    print(f'mean RPC latency, first {WINDOW:>4}   : {first:7.3f} ms')
    print(f'mean RPC latency, last  {WINDOW:>4}   : {last:7.3f} ms')
    print(f'slowdown                       : {last / first:.1f}x')

    ch.queue_delete(queue='leak-demo')
    ch.close()
    conn.close()


def measure_registration_cost():
    """Why it slows down: CallbackManager.add rescans the list on every add."""
    print()
    print('cost of registering one more close callback (callback.py:154):')
    print(f'  {"already registered":>18} {"cost of next add":>18}')
    for n in (0, 1000, 2000, 4000, 8000, 16000):
        manager = CallbackManager()
        for _ in range(n):
            manager.add(1, '_on_channel_close', lambda *a: None, False, 'c')
        probes = [lambda *a: None for _ in range(200)]
        start = time.perf_counter()
        for probe in probes:
            manager.add(1, '_on_channel_close', probe, False, 'c')
        per = (time.perf_counter() - start) / 200 * 1e6
        print(f'  {n:>18} {per:>15.1f} us')


if __name__ == '__main__':
    measure_accumulation()
    measure_registration_cost()
```

### Actual output on main (141f8457)

```
close callbacks before any RPC : 0
queue_declare calls issued     : 500 warmup + 8000 measured
close callbacks after warmup   : 500
close callbacks at the end     : 8500
retained heap growth           : 15903 KiB

mean RPC latency, first  500   :   0.373 ms
mean RPC latency, last   500   :   0.854 ms
slowdown                       : 2.3x

cost of registering one more close callback (callback.py:154):
  already registered   cost of next add
                   0             2.0 us
                1000            19.1 us
                2000            38.4 us
                4000            76.6 us
                8000           154.2 us
               16000           360.4 us
```

One callback retained per RPC, 15.9 MB held on a single channel, and registration cost rising linearly with what is already there, so a channel 16k RPCs into its life burns 0.36 ms of pure bookkeeping per call.

The callback count and the retained memory are exactly reproducible run to run. The end-to-end slowdown is noisy because it includes broker round trips; across runs here it ranged from 1.7x to 4.1x, always in the same direction. The registration-cost table is the deterministic measurement behind it.

### Expected

The count and memory should stay flat: a completed RPC has no further use for its close callback.

### Suggested fix

Unregister once the RPC completes, using the public APIs pika already provides, `Channel.remove_on_close_callback` and `Channel.remove_callback`. The removal has to be scheduled on the IOLoop thread, since that is what owns the callbacks dict. The `finally` in `_blocking_rpc` and `basic_get` that already calls `_unregister_waiter` is the natural place.

### Environment

- pika: main at 141f8457
- Python 3.14.4, Linux
- RabbitMQ 4.3.4, Erlang 27.3.4.16

alonfaraj:
## Proposed Changes

`ThreadSafeChannel._blocking_rpc` registered a fresh `_on_chan_close` closure on the raw channel for every call and never removed it:

```python
def _on_chan_close(ch, reason) -> None:
    if not ready.is_set():
        ...

try:
    self._channel.add_on_close_callback(_on_chan_close)   # never removed
    channel_method(*args, **kwargs, callback=_on_ok)
```

`Channel.add_on_close_callback` registers with `one_shot=False`, and each call passes a distinct closure, so `CallbackManager.add`'s duplicate check never collapses them. They accumulate for the life of the channel. `basic_get` does the same and additionally leaves its `Basic.GetEmpty` one-shot behind whenever the get returns a message, because a one-shot is only consumed if it fires.

This affects every blocking RPC: `queue_declare`, `queue_bind`, `queue_delete`, `queue_purge`, `exchange_declare`, `exchange_bind`, `exchange_delete`, `basic_qos`, `basic_get`, `basic_consume`, `basic_cancel`, `confirm_delivery`, and `close`.

Firing the strays is harmless, since each guards on `ready.is_set()` and a completed RPC has it set, so this costs memory and CPU rather than correctness. Two effects, both proportional to how many RPCs the channel has served:

- around 2 KiB per RPC, held until the channel closes
- `CallbackManager.add` linearly rescans the existing list on every registration, and since the closures never match, that scan is pure waste that grows without bound

It matters for the workload this adapter targets, a long-lived connection whose channels stay open for the process lifetime. A short-lived channel never notices.

The fix hoists the per-RPC closures out of the scheduled `_invoke` so they can be referenced afterwards, and adds `_release_close_callback`, called from the same `finally` that already unregisters the waiter. Removal goes through the public `Channel.remove_on_close_callback` and `Channel.remove_callback`, scheduled on the IOLoop thread because the channel's callback stack belongs to it. Ordering is safe: the `_invoke` that registers is queued first and the IOLoop drains its callback queue in FIFO order, so the removal can never run ahead of the registration.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

## Fixes

- Fixes #1684

### Measured effect

The reproduction script from #1684, run unchanged against `main` and this branch back to back on the same machine and broker (RabbitMQ 4.3.4, Python 3.14.4). 500 warmup plus 8000 measured `queue_declare` calls on one channel.

Before, on `main`:

```
close callbacks before any RPC : 0
queue_declare calls issued     : 500 warmup + 8000 measured
close callbacks after warmup   : 500
close callbacks at the end     : 8500
retained heap growth           : 15873 KiB

mean RPC latency, first  500   :   0.164 ms
mean RPC latency, last   500   :   0.453 ms
slowdown                       : 2.8x
```

After, on this branch:

```
close callbacks before any RPC : 0
queue_declare calls issued     : 500 warmup + 8000 measured
close callbacks after warmup   : 0
close callbacks at the end     : 1
retained heap growth           : 269 KiB

mean RPC latency, first  500   :   0.149 ms
mean RPC latency, last   500   :   0.141 ms
slowdown                       : 0.9x
```

| | before | after |
| --- | --- | --- |
| close callbacks retained | 8500 | 1 |
| retained heap growth | 15873 KiB | 269 KiB |
| mean RPC latency, first 500 | 0.164 ms | 0.149 ms |
| mean RPC latency, last 500 | 0.453 ms | 0.141 ms |
| slowdown across 8000 calls | 2.8x | 0.9x |

The point is that latency stops climbing: flat within noise across 8000 calls, against nearly tripling before. The retained `1` is not a leftover; removal is scheduled on the IOLoop, so at most one cleanup is in flight when the count is sampled. It drains to 0 and does not grow with the number of RPCs, which I checked at 100, 1000 and 5000.

End-to-end timings include broker round trips and are noisy, so the ratio is not a precise figure. Across repeated runs it was 1.7x to 4.1x before and 0.5x to 1.1x after, always in the same direction. The deterministic measurement behind it is the cost of one more registration, which rises linearly with what is already registered:

```
  already registered   cost of next add
                   0             2.1 us
                1000            19.8 us
                4000            79.9 us
               16000           331.6 us
```

That curve is a property of `CallbackManager` and is unchanged by this PR, which is why the script prints the same table before and after. What changes is that a long-lived channel no longer travels down it, staying at the top row instead of reaching the bottom.

Separately, 300 `basic_get` calls that all returned a message left 301 close callbacks and 301 `Basic.GetEmpty` entries; those are now 0 and 1, the 1 being pika's own permanent handler.

### Tests

Unit tests in `tests/unit/thread_safe_connection_tests.py`, one per property:

- a completed RPC removes the exact close callback it registered
- a timed-out RPC cleans up too, so a flaky broker does not leak one per call
- `basic_get` removes both its close callback and its `Basic.GetEmpty` one-shot
- the removal is scheduled on the IOLoop rather than run on the calling thread

Acceptance test `TestPerRPCCallbacksDoNotAccumulate` drives 50 `queue_declare` calls and 20 gets over a real connection and asserts the observable consequence, that the close-callback list does not grow with the RPCs served. It allows one in-flight cleanup, since removal is scheduled rather than inline.

All five fail against the unfixed adapter; the acceptance test reports `51 not less than or equal to 2`.

One existing assertion changed. `test_confirm_delivery_is_idempotent` counted callbacks scheduled on the IOLoop as a proxy for "one RPC was sent", which no longer holds now that a single RPC also schedules its own cleanup. It now counts the frames handed to the raw channel, which is what it was actually trying to assert.

### DIFF
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -478,6 +478,31 @@ class Channel:
             except ValueError:
                 pass
 
+    def _release_close_callback(self, on_chan_close, *extra) -> None:
+        """
+        Unregister the per-RPC channel callbacks now that the call has finished.
+
+        :meth:`~pika.channel.Channel.add_on_close_callback` registers with ``one_shot=False`` and
+        each RPC passes a distinct closure, so nothing ever collapses them: left in place they
+        accumulate for the life of the channel, and ``CallbackManager.add`` rescans that growing
+        list on every later registration.  A completed RPC has no use for its close callback, so
+        drop it.
+
+        Removal is scheduled rather than done inline because the channel's callback stack belongs to
+        the IOLoop thread.  Ordering is safe: the ``_invoke`` that registers is queued before this,
+        and the IOLoop drains its callback queue in FIFO order, so the removal never runs first.
+
+        :param on_chan_close: The close callback registered for this RPC.
+        :param extra: Zero-argument callables removing any other per-RPC callbacks.
+        """
+
+        def _remove() -> None:
+            self._channel.remove_on_close_callback(on_chan_close)
+            for remove_one in extra:
+                remove_one()
+
+        self._wrapper._schedule_unchecked(_remove)
+
     def _blocking_rpc(self, method_name: str, channel_method,
                       timeout: float | None, *args, **kwargs) -> Any:
         """
@@ -503,18 +528,17 @@ class Channel:
         ready, error = self._register_waiter()
         result = [None]
 
-        def _invoke() -> None:
+        def _on_ok(method_frame) -> None:
+            result[0] = method_frame
+            ready.set()
 
-            def _on_ok(method_frame) -> None:
-                result[0] = method_frame
+        def _on_chan_close(ch, reason) -> None:
+            if not ready.is_set():
+                if error[0] is None:
+                    error[0] = reason
                 ready.set()
 
-            def _on_chan_close(ch, reason) -> None:
-                if not ready.is_set():
-                    if error[0] is None:
-                        error[0] = reason
-                    ready.set()
-
+        def _invoke() -> None:
             try:
                 self._channel.add_on_close_callback(_on_chan_close)
                 channel_method(*args, **kwargs, callback=_on_ok)
@@ -530,6 +554,7 @@ class Channel:
                     f'{method_name} timed out after {timeout} seconds')
         finally:
             self._unregister_waiter(ready, error)
+            self._release_close_callback(_on_chan_close)
 
         if error[0] is not None:
             raise _reraisable(error[0], self._wrapper._closed_reason,
@@ -705,23 +730,22 @@ class Channel:
         ready, error = self._register_waiter()
         result = [None, None, None]
 
+        def _on_get_ok(ch, method, properties, body) -> None:
+            result[0] = method
+            result[1] = properties
+            result[2] = body
+            ready.set()
+
+        def _on_get_empty(method_frame) -> None:
+            ready.set()
+
+        def _on_chan_close(ch, reason) -> None:
+            if not ready.is_set():
+                if error[0] is None:
+                    error[0] = reason
+                ready.set()
+
         def _get() -> None:
-
-            def _on_get_ok(ch, method, properties, body) -> None:
-                result[0] = method
-                result[1] = properties
-                result[2] = body
-                ready.set()
-
-            def _on_get_empty(method_frame) -> None:
-                ready.set()
-
-            def _on_chan_close(ch, reason) -> None:
-                if not ready.is_set():
-                    if error[0] is None:
-                        error[0] = reason
-                    ready.set()
-
             try:
                 self._channel.add_on_close_callback(_on_chan_close)
                 self._channel.add_callback(
@@ -746,6 +770,11 @@ class Channel:
                     f'basic_get timed out after {timeout} seconds')
         finally:
             self._unregister_waiter(ready, error)
+            # A one-shot is only consumed if it fires, so the Basic.GetEmpty
+            # callback outlives every get that actually returned a message.
+            self._release_close_callback(
+                _on_chan_close, lambda: self._channel.remove_callback(
+                    _on_get_empty, [spec.Basic.GetEmpty]))
 
         if error[0] is not None:
             raise _reraisable(error[0], self._wrapper._closed_reason,
--- a/tests/acceptance/thread_safe_connection_test.py
+++ b/tests/acceptance/thread_safe_connection_test.py
@@ -400,5 +400,46 @@ class TestContextManager(ThreadSafeTestCaseBase):
         self.assertTrue(conn.is_closed)
 
 
+class TestPerRPCCallbacksDoNotAccumulate(ThreadSafeTestCaseBase):
+    """
+    A long-lived channel must not grow a callback per RPC it serves.
+
+    ``add_on_close_callback`` registers with ``one_shot=False`` and every RPC passes a distinct
+    closure, so anything left behind is held until the channel closes and is rescanned by every
+    later registration.  Unit tests assert the removal calls; this asserts the observable
+    consequence against a real broker.
+    """
+
+    @staticmethod
+    def _close_callback_count(ch):
+        raw = ch._channel
+        # CallbackManager normalizes the prefix to a string.
+        stack = raw.callbacks._stack.get(str(raw.channel_number), {})
+        return len(stack.get('_on_channel_close', []))
+
+    def test(self):
+        conn = self._connect()
+        ch = conn.channel()
+        queue = self._unique_queue()
+        ch.queue_declare(queue=queue, exclusive=True)
+
+        baseline = self._close_callback_count(ch)
+        for _ in range(50):
+            ch.queue_declare(queue=queue, exclusive=True)
+
+        # At most one cleanup may still be in flight on the IOLoop thread.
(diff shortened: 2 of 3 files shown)


lukebakken (approved):
> [!NOTE]

LGTM. Verified locally and in CI (59/59 green):

- The stale `ThreadSafeChannel` reference (rebase residue) now calls `Channel`; the four `PerRPCCallbackReleaseTests` run and pass, and ruff/yapf/docformatter are green again.
- The accumulation acceptance test is strengthened: it publishes one message per get and asserts both the `Basic.GetEmpty` and `_on_channel_close` buckets stay flat. Neutralizing the GetEmpty removal makes it fail, so it now catches the leak it guards.

PR #1689, opened 2026-08-13
Drop each closed channel from the connection's tracking list

### LINKED ISSUE #1688: `Connection._channels` is append-only, so every closed channel is retained forever
### Summary

`pika.adapters.thread_safe_connection.Connection` tracks its wrapper channels in `self._channels` so it can drain their consumer pools at shutdown. Nothing ever removes an entry. `Channel.close()` shuts the consumer pool down but leaves the wrapper on the list, so a long-lived connection with channel churn retains every `Channel` it has ever opened, each holding its `_BoundedWorkPool`, its raw `pika.channel.Channel`, and the per-RPC closures registered on it.

The underlying `pika.connection.Connection` releases its own channels correctly. Only the thread-safe wrapper leaks.

### Root cause

The list is created at `pika/adapters/thread_safe_connection.py:1478`, appended to at line 1702, and read at line 1593. Those are its only three references:

```
1478:        self._channels: list[Channel] = []
1593:            channels = list(self._channels)
1702:            self._channels.append(ch)
```

`Channel.close()` (line 1256) waits for `Channel.CloseOk` and then calls `_shutdown_pool` (line 426), which claims `_pool_shutdown` and joins the worker. Neither touches `self._wrapper._channels`.

Channel-per-unit-of-work is a normal pattern, and it is the pattern this class encourages, since a `Channel` here owns a dedicated worker thread and a bounded queue.

### Reproduction

```python
"""Connection._channels is append-only, so closed channels are retained."""
import gc

import pika

CHURN = 200

print(f'wrapper channels tracked at start        : {len(conn._channels)}')

for _ in range(CHURN):
    ch = conn.channel()
    ch.close()
del ch
gc.collect()

tracked = conn._channels
print(f'wrapper channels tracked after {CHURN} open/close: {len(tracked)}')
print(f'all of them report is_closed            : '
      f'{all(c.is_closed for c in tracked)}')
print(f'raw pika channels on the underlying conn : '
      f'{len(conn._connection._channels)}')
print(f'live _BoundedWorkPool objects retained   : '
      f'{sum(1 for c in tracked if c._consumer_work_pool is not None)}')
conn.close()
```

Output:

```
wrapper channels tracked at start        : 0
wrapper channels tracked after 200 open/close: 200
all of them report is_closed            : True
raw pika channels on the underlying conn : 0
live _BoundedWorkPool objects retained   : 200
```

All 200 wrappers are closed, are unreachable from user code, and are still held by the connection. The underlying connection has correctly dropped all 200 of its own channels.

### Expected behaviour

A closed `Channel` should be dropped from `Connection._channels`, so the list tracks live channels only, matching what the underlying `pika.connection.Connection` does with its own.

### Impact

- Unbounded memory growth proportional to the number of channels ever opened, in exactly the long-lived-connection deployments this adapter targets. Each retained wrapper holds a `_BoundedWorkPool` (with its `queue.Queue`), the raw channel, and that channel's `CallbackManager` entries.
- `_shutdown_all_consumer_pools` sweeps the full list on every close. Already-shut-down pools return early from `_shutdown_pool`, so this is cheap per entry, but it is O(all channels ever opened) rather than O(live channels).
- The leak is invisible to `is_closed` based diagnostics, since every retained entry reports itself closed.

### Suggested fix

Drop the channel from the tracking list when its pool shutdown is claimed, inside the lock that already guards that transition. In `Channel._shutdown_pool` (line 438):

```python
with self._wrapper._channel_waiters_lock:
    if self._pool_shutdown:
        return
    self._pool_shutdown = True
    try:
        self._wrapper._channels.remove(self)
    except ValueError:
        pass
```

`_shutdown_pool` is the single choke point for this transition, and `_pool_shutdown` makes it idempotent, so this removes each channel exactly once whether it was closed by the user or swept at connection shutdown.

One ordering detail to keep in mind: `_shutdown_all_consumer_pools` already snapshots the list under the lock (`channels = list(self._channels)`, line 1593) before iterating, so removing entries during that iteration is safe. `_signal_pool_shutdown` (line 416) deliberately does not claim `_pool_shutdown`, so the bounded path still signals every pool before joining any of them.

### Environment

- pika 1.4.0, commit 0bbc026f

### Notes

Not covered by the current test suite (938 unit tests pass; `hatch run typecheck` and `hatch run lint-check` are clean). No test asserts on the size of `Connection._channels` after a channel is closed.

alonfaraj:
## Proposed Changes

`Connection._channels` exists only so `_shutdown_all_consumer_pools` can reach every live consumer pool, but nothing ever removed an entry. `Channel.close()` shut the pool down and left the wrapper on the list, so a connection with channel churn retained every `Channel` it had ever opened, along with that channel's `_BoundedWorkPool`, raw channel and per-RPC callbacks. Channel-per-unit-of-work is the pattern this adapter encourages, and long-lived connections are what it targets, so the leak grows without bound in exactly the deployments it is built for.

This removes the channel in `_shutdown_pool`, under the lock that already guards the transition.

200 open/close cycles against RabbitMQ 4.3.4, using the reproduction from #1688:

| | before | after |
| --- | --- | --- |
| wrapper channels tracked | 200 | 0 |
| `_BoundedWorkPool` objects retained | 200 | 0 |

<details>
<summary>Reproduction output, before and after</summary>

Both runs are from the same working tree, reverting and restoring only `pika/adapters/thread_safe_connection.py` between them, against RabbitMQ 4.3.4 on Python 3.14.4.

Before:

```
wrapper channels tracked at start        : 0
wrapper channels tracked after 200 open/close: 200
all of them report is_closed            : True
raw pika channels on the underlying conn : 0
live _BoundedWorkPool objects retained   : 200
```

After:

```
wrapper channels tracked at start        : 0
wrapper channels tracked after 200 open/close: 0
all of them report is_closed            : True
raw pika channels on the underlying conn : 0
live _BoundedWorkPool objects retained   : 0
```

All 200 wrappers were closed and unreachable from user code in both runs, and the underlying `pika.connection.Connection` correctly dropped all 200 of its own channels in both. Only the wrapper's list differs. The `is_closed` line reads `True` vacuously in the second run, over an empty list.

<details>
<summary>Script</summary>

```python
import gc

import pika

CHURN = 200

print(f'wrapper channels tracked at start        : {len(conn._channels)}')

for _ in range(CHURN):
    ch = conn.channel()
    ch.close()
del ch
gc.collect()

tracked = conn._channels
print(f'wrapper channels tracked after {CHURN} open/close: {len(tracked)}')
print(f'all of them report is_closed            : '
      f'{all(c.is_closed for c in tracked)}')
print(f'raw pika channels on the underlying conn : '
      f'{len(conn._connection._channels)}')
print(f'live _BoundedWorkPool objects retained   : '
      f'{sum(1 for c in tracked if c._consumer_work_pool is not None)}')
conn.close()
```

The queue-free version above declares nothing, so it runs as-is. Note that on RabbitMQ 4.x a `queue_declare(auto_delete=True)` in a variant of this script is rejected (`transient_nonexcl_queues` is deprecated); use `durable=True`.

</details>
</details>

## Types of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

## Checklist

`hatch run unit`: 940 passed, 21 skipped. `lint-check`, `fmt-check`, `docfmt-check` and `typecheck` clean. `tests/acceptance/thread_safe_connection_test.py` and `thread_safe_bounded_queue_test.py` against a live broker: 12 passed.

Two unit tests added: one churns channels and asserts each is tracked while live and dropped on close, one covers the connection-wide sweep and pins the snapshot-under-lock reasoning that makes the mutation safe. Documentation is the updated `_shutdown_pool` docstring; no user-facing docs change, since the tracking list is internal.

## Fixes

- Fixes #1688

### DIFF
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -427,6 +427,14 @@ class Channel:
         """
         Shut down the consumer work pool, allowing in-flight work to finish.
 
+        Also drops this channel from the connection's tracking list.  That list exists only so
+        :meth:`~Connection._shutdown_all_consumer_pools` can reach every live pool, and a channel
+        whose pool is shut down has nothing left to drain; left in place the entries accumulate for
+        the life of the connection, retaining each closed channel's pool, raw channel and per-RPC
+        callbacks.  This is the one choke point for the transition and ``_pool_shutdown`` makes it
+        idempotent, so each channel is dropped exactly once whether it was closed by the user or
+        swept at connection shutdown.
+
         :param timeout: Seconds to wait for the worker to drain and exit. ``None`` waits
             indefinitely. A wedged worker that outlives a finite *timeout* is left running (it is a
             daemon thread) rather than stalling the caller; a warning is logged.
@@ -439,6 +447,13 @@ class Channel:
             if self._pool_shutdown:
                 return
             self._pool_shutdown = True
+            # Safe to mutate while a sweep is in flight:
+            # _shutdown_all_consumer_pools snapshots the list under this same
+            # lock before iterating it.
+            try:
+                self._wrapper._channels.remove(self)
+            except ValueError:
+                pass
         if not self._consumer_work_pool.shutdown(wait=True, timeout=timeout):
             LOGGER.warning(
                 'Channel %s consumer work pool did not drain within '
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -423,42 +423,87 @@ class Channel:
         """
         self._consumer_work_pool.shutdown(wait=False)
 
+    def _claim_pool_shutdown(self) -> bool:
+        """
+        Claim this channel's pool shutdown and drop it from connection tracking.
+
+        Returns ``True`` if this call is the first to claim the shutdown - so the caller owns any
+        follow-up work, such as joining the worker - and ``False`` if a prior close already claimed
+        it.  Being idempotent lets a user :meth:`close`, the broker-close hook
+        (:meth:`_on_broker_close`) and the connection-shutdown sweep race while each channel is
+        dropped from ``_channels`` exactly once.
+
+        The tracking list exists only so :meth:`~Connection._shutdown_all_consumer_pools` can reach
+        every live pool, and a channel whose pool is shutting down has nothing left to drain; left
+        in place the entries accumulate for the life of the connection, retaining each closed
+        channel's pool, raw channel and per-RPC callbacks.
+
+        :returns: ``True`` if this call claimed the shutdown, ``False`` if a prior close already did.
+        """
+        # Safe to mutate the list while a sweep is in flight:
+        # _shutdown_all_consumer_pools snapshots it under this same lock before
+        # iterating.
+        with self._wrapper._channel_waiters_lock:
+            if self._pool_shutdown:
+                return False
+            self._pool_shutdown = True
+            try:
+                self._wrapper._channels.remove(self)
+            except ValueError:
+                pass
+        return True
+
     def _shutdown_pool(self, timeout: float | None = None) -> None:
         """
-        Shut down the consumer work pool, allowing in-flight work to finish.
+        Shut down the consumer work pool, joining its worker.
 
-        Also drops this channel from the connection's tracking list.  That list exists only so
-        :meth:`~Connection._shutdown_all_consumer_pools` can reach every live pool, and a channel
-        whose pool is shut down has nothing left to drain; left in place the entries accumulate for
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
> [!NOTE]

LGTM. Verified locally (943 unit + 9 acceptance tests, lint/yapf/docformatter/mypy) and in CI (59/59 green). The broker-close hook drops a channel from tracking on the IOLoop thread via `shutdown(wait=False)` - no join, so it can't deadlock a worker mid-callback - and is gated on non-`ChannelClosedByClient` reasons so it never pre-empts the budgeted join in `close()`/the sweep. `_claim_pool_shutdown` keeps the user-close, hook, and sweep paths idempotent.

PR #1690, opened 2026-08-16
Contain exceptions from the basic_publish on_publish callback

### LINKED ISSUE #1687: An exception from `basic_publish(on_publish=...)` kills the whole connection
### Summary

`on_publish` is the only user callback in `pika/adapters/thread_safe_connection.py` invoked unguarded. It runs bare on the IOLoop thread, outside the `try/except` that wraps the publish, so an exception from user code propagates out of `ioloop.start()` and kills the IOLoop thread. The connection is dead from that point on, and every later operation raises the user's own exception as the recorded close reason.

Every other user callback in this module (deliveries, publisher confirms, returned messages, cancel notifications, `Connection.Blocked`/`Unblocked`) goes through `Channel._safe_dispatch`, which logs and swallows.

### Root cause

`pika/adapters/thread_safe_connection.py:590`. The `try` covers only the raw `basic_publish`; the `on_publish` call sits after it:

```python
def _publish() -> None:
    try:
        self._channel.basic_publish(...)
    except Exception:
        LOGGER.warning('basic_publish failed (channel may have closed)',
                       exc_info=True)
        return
    if self._next_publish_seq_no is not None:
        self._next_publish_seq_no += 1
        if on_publish is not None:
            on_publish(self._next_publish_seq_no)   # line 606, unguarded
```

`_publish` reaches the IOLoop via `_schedule_unchecked` -> `add_callback_threadsafe`, and the select IOLoop invokes queued callbacks without a guard (`pika/adapters/select_connection.py:542`, `callback()`). The exception unwinds out of `ioloop.start()` into `_run_ioloop`'s handler (line 1508), which records it as `_closed_reason` and wakes every blocked waiter with it.

### Reproduction

```python
"""A raising on_publish callback kills the IOLoop thread and the connection."""
import logging
import time

import pika

logging.basicConfig(
    level=logging.ERROR,
    format='%(asctime)s %(threadName)s %(levelname)s %(message)s')

QUEUE = 'repro_on_publish_raises'

ch = conn.channel()
ch.confirm_delivery(ack_nack_callback=lambda method_frame: None)


def on_publish(delivery_tag):
    raise ValueError(f'user bug while recording tag {delivery_tag}')


ch.basic_publish(exchange='',
                 routing_key=QUEUE,
                 body=b'one',
                 on_publish=on_publish)
time.sleep(1.0)

print(f'conn.is_open              = {conn.is_open}')
print(f'IOLoop thread alive       = {conn._ioloop_thread.is_alive()}')
print(f'recorded close reason     = {conn._closed_reason!r}')
try:
    ch.queue_declare(queue=QUEUE, passive=True, timeout=3)
    print('subsequent RPC           : succeeded, connection survived')
except Exception as exc:
    print(f'subsequent RPC           : {type(exc).__name__}: {exc}')
```

Output, with the middle of the traceback elided:

```
2026-08-13 12:26:12,811 pika-ioloop-1 ERROR IOLoop thread crashed
  File ".../pika/adapters/thread_safe_connection.py", line 1507, in _run_ioloop
  [...]
  File ".../pika/adapters/select_connection.py", line 542, in process_timeouts
    callback()
  File ".../pika/adapters/thread_safe_connection.py", line 606, in _publish
    on_publish(self._next_publish_seq_no)
  File "repro_on_publish_raises.py", line 28, in on_publish
    raise ValueError(f'user bug while recording tag {delivery_tag}')
ValueError: user bug while recording tag 1
conn.is_open              = True
IOLoop thread alive       = False
recorded close reason     = ValueError('user bug while recording tag 1')
subsequent RPC           : ValueError: user bug while recording tag 1
```

### Expected behaviour

An exception from `on_publish` should be logged and contained, exactly as one from a consumer callback or a publisher-confirm callback is, and the connection should stay usable.

### Impact

A bug in a user's confirm-tracking callback takes down the whole connection, along with every channel and consumer on it. The blast radius is widest where `on_publish` is most useful: high-rate confirm tracking, where one malformed tag or a `KeyError` in the application's in-flight map is enough.

### Suggested fix

Route `on_publish` through the existing helper, keeping it on the IOLoop thread as documented:

```python
if self._next_publish_seq_no is not None:
    self._next_publish_seq_no += 1
    if on_publish is not None:
        self._safe_dispatch('on_publish callback', on_publish,
                            self._next_publish_seq_no)
```

### Two related observations from the same run

`conn.is_open` returned `True` with the IOLoop thread dead. `Connection.is_open` (line 1992) delegates to the underlying `SelectConnection`, and `_run_ioloop`'s crash path records `_closed_reason` without closing it, so the object reports itself open and its socket stays open until GC. Methods correctly raise via `_check_not_closed`, so the properties contradict them. Worth tearing the underlying connection down on that path.

`_user_on_close_callback` (line 1581) and `_user_on_open_error_callback` (line 1566) are unguarded the same way. They fire during teardown, so an exception there aborts the rest of pika's teardown rather than a live connection, but the fix is the same shape.

### Environment

- pika 1.4.0, commit 0bbc026f

### Notes

Not covered by the current test suite (938 unit tests pass; `hatch run typecheck` and `hatch run lint-check` are clean). The four `on_publish` tests in `tests/unit/thread_safe_connection_tests.py` (lines 509-577) cover confirms-disabled, tag sequencing, the publish-raises path and seq-no increment without a callback, but none covers `on_publish` itself raising.

aryansk:
## What

Fixes #1687 — an exception from the `on_publish` callback of `basic_publish` is now logged and contained instead of propagating out of the IOLoop thread and killing the whole connection.

## Why

`on_publish` was the only user callback in `ThreadSafeChannel` invoked unguarded. It ran bare on the IOLoop thread, outside the `try/except` that wraps the `basic_publish` call itself, so an exception from user code unwound out of `ioloop.start()`. From then on the connection was dead: every later operation raised the user's own exception as the recorded close reason, taking every channel and consumer down with it.

## How

Route `on_publish` through the existing `_safe_dispatch` helper — the same one used for consumer callbacks, publisher confirms, returned messages, cancel notifications, and `Connection.Blocked`/`Unblocked` — keeping it on the IOLoop thread as documented:

```python
self._safe_dispatch('publish callback', on_publish,
                    self._next_publish_seq_no)
```

## Tests

Added `ChannelTests.test_on_publish_exception_is_contained` in `tests/unit/thread_safe_connection_tests.py`: an `on_publish` that raises `ValueError` must not propagate out of the scheduled publish callback, and the delivery tag is still consumed. Fails before the fix (the `ValueError` escapes), passes after. Full unit suite: 938 passed, 27 skipped; `mypy` and `ruff` clean.

🤖 Generated with Codebuff

### DIFF
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -663,7 +663,13 @@ class Channel:
             if self._next_publish_seq_no is not None:
                 self._next_publish_seq_no += 1
                 if on_publish is not None:
-                    on_publish(self._next_publish_seq_no)
+                    # Dispatch through _safe_dispatch like every other user
+                    # callback (consumer, publisher-confirm, ...) so an
+                    # exception from user code is logged and contained
+                    # instead of propagating out of the IOLoop thread and
+                    # killing the whole connection (#1687).
+                    self._safe_dispatch('publish callback', on_publish,
+                                        self._next_publish_seq_no)
 
         self._wrapper._schedule_unchecked(_publish)
 
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -663,11 +663,15 @@ class Channel:
             if self._next_publish_seq_no is not None:
                 self._next_publish_seq_no += 1
                 if on_publish is not None:
-                    # Dispatch through _safe_dispatch like every other user
-                    # callback (consumer, publisher-confirm, ...) so an
-                    # exception from user code is logged and contained
-                    # instead of propagating out of the IOLoop thread and
-                    # killing the whole connection (#1687).
+                    # Guard on_publish with _safe_dispatch so an exception
+                    # from user code is logged and contained instead of
+                    # propagating out of the IOLoop thread and killing the
+                    # whole connection (#1687).  Unlike the consumer, confirm,
+                    # return and cancel callbacks, which _submit_or_terminate
+                    # hands to the work pool, on_publish runs inline here on
+                    # the IOLoop thread as the basic_publish docstring
+                    # documents; _safe_dispatch adds only the exception guard,
+                    # not the pool dispatch.
                     self._safe_dispatch('publish callback', on_publish,
                                         self._next_publish_seq_no)
 
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -397,12 +397,12 @@ class Channel:
     @staticmethod
     def _safe_dispatch(label, callback, *args) -> None:
         """
-        Execute *callback* on the pool worker, logging any exception.
+        Run *callback*, logging any exception instead of letting it propagate.
 
-        Used to wrap user callbacks before submitting them to a
-        :class:`_BoundedWorkPool` so that exceptions are not silently lost
-        and so that one failing dispatch does not prevent subsequent
-        dispatches from being processed.
+        Wraps a user callback so one failing dispatch neither escapes into the caller nor prevents
+        later dispatches from running.  Most callers submit the wrapped callback to a
+        :class:`_BoundedWorkPool`, so it runs on a pool worker; :meth:`basic_publish` calls it
+        inline on the IOLoop thread. Either way the exception is contained here.
 
         :param label: Human-readable name of the callback for log lines.
         :param callback: The user callback.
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
> [!NOTE]

LGTM; CI is green (60/60) at 2184e869.

The fix routes `on_publish` through `_safe_dispatch`, so an exception from the callback is logged and contained on the IOLoop thread instead of propagating out of `ioloop.start()` and tearing down the connection (#1687). Every other user callback already goes through that helper, and the regression test covers the contained-exception path.

The follow-up commits close the documentation gaps the change opened: `_safe_dispatch` no longer claims pool-worker-only execution (it now names both the pool and the inline `basic_publish` path), and the `on_publish` parameter documents that its exceptions are logged and suppressed rather than propagated. A clean, contained bugfix with honest docs.

lukebakken:
Thank you @aryansk

aryansk:
Thank you for the merge @lukebakken — glad it landed cleanly.

PR #1692, opened 2026-08-16
Bound how long a channel retains nowait-cancelled consumer tags

### LINKED ISSUE #1691: `Channel._cancelled` grows without bound on nowait `basic_cancel`
## Summary

`pika.channel.Channel` keeps a cancelled consumer's tag in `_cancelled` (and, for an `auto_ack` consumer, in `_consumers_with_noack`) forever when the consumer is cancelled with nowait. Nothing reclaims those entries while the channel is open, so a long-lived channel with consumer churn grows without bound.

Same shape as #1688, an append-only container whose only reclaim point is object teardown, but in the core `Channel`, so it affects every async adapter (`SelectConnection`, `AsyncioConnection`, `TornadoConnection`, `TwistedProtocolConnection`).

## Root cause

`basic_cancel(consumer_tag)` with no `callback` is a nowait cancel. `pika/channel.py:339-351` already documents the consequence:

```python
        if nowait:
            # This is our last opportunity while the channel is open to remove
            # this consumer callback and help gc; unfortunately, this consumer's
            # self._cancelled and self._consumers_with_noack (if any) entries
            # will persist until the channel is closed.
            del self._consumers[consumer_tag]
```

`_cleanup_consumer_ref` (`pika/channel.py:1109`) is the only thing that discards from either container, and its two callers are both dead ends here: `_on_cancelok` (`:1166`) fires on `Basic.CancelOk`, which a nowait cancel asks the broker not to send, and `_on_cancel` (`:1154`) returns early for any tag already in `_cancelled`.

## Impact

Growth is linear in the number of nowait cancels and reclaimed only at channel close. The affected call is the idiomatic one in async pika code:

```python
channel.basic_cancel(consumer_tag)   # no callback -> nowait -> leaks two entries
```

Callers that pass a callback are unaffected, which is why this has gone unnoticed: `BlockingChannel.basic_cancel` and `thread_safe_connection.Channel.basic_cancel` both pass one, so only direct users of the core `Channel` API hit it.

About 40 bytes per tag held twice makes this a slow burn rather than a crash, but consumer bookkeeping never returns to a clean state: `_cancelled` becomes a permanent record of everything the channel has ever cancelled.

## Reproduction

Requires RabbitMQ on `localhost:5672`. Pure core, `SelectConnection` only, every channel call on the IOLoop thread. Runs 3000 consume/cancel cycles twice on one long-lived channel, once with nowait and once with a callback, isolating the cause to the missing `Basic.CancelOk`.

<details>
<summary>repro_cancelled_growth.py</summary>

```python
"""`Channel._cancelled` and `_consumers_with_noack` grow without bound on a
long-lived channel when consumers are cancelled with nowait."""
import sys
import threading

import pika
from pika.adapters.select_connection import SelectConnection
from pika.channel import Channel

CYCLES = 3000
HOST = 'localhost'


def report(label, channel):
    def flag(size):
        cap = getattr(Channel, '_MAX_RETAINED_CANCELLED_TAGS', None)
        if not size:
            return ''
        if cap is not None and size >= cap:
            return f'   <-- capped at {cap}'
        return '   <-- retained until the channel closes'

    cancelled = len(channel._cancelled)
    noack = len(channel._consumers_with_noack)
    print(f'{label}:')
    print(f'    _consumers            = {len(channel._consumers):>4}')
    print(f'    _cancelled            = {cancelled:>4}{flag(cancelled)}')
    print(f'    _consumers_with_noack = {noack:>4}{flag(noack)}')
    retained = sum(
        sys.getsizeof(t)
        for t in (channel._cancelled | channel._consumers_with_noack))
    print(f'    bytes held by the tag strings alone = {retained}')


class Repro:
    """Drive consume/cancel churn on the IOLoop thread."""

    def __init__(self, nowait):
        self.nowait = nowait
        self.done = threading.Event()
        self.channel = None
        self.queue = None
        self.cycle = 0
        self.connection = SelectConnection(
            pika.ConnectionParameters(HOST),
            on_open_callback=self._on_open,
            on_open_error_callback=self._on_open_error,
            on_close_callback=lambda *a: None)

    def run(self):
        thread = threading.Thread(target=self.connection.ioloop.start,
                                  daemon=True)
        thread.start()
        if not self.done.wait(timeout=60):
            raise RuntimeError('timed out')
        self.connection.ioloop.add_callback_threadsafe(
            self.connection.ioloop.stop)
        thread.join(timeout=10)
        return self.channel

    def _on_open_error(self, _conn, error):
        print(f'could not connect to {HOST}:5672 - {error}')
        self.done.set()

    def _on_open(self, _conn):
        self.connection.channel(on_open_callback=self._on_channel_open)

    def _on_channel_open(self, channel):
        self.channel = channel
        channel.queue_declare('',
                              exclusive=True,
                              callback=self._on_queue_declared)

    def _on_queue_declared(self, frame):
        self.queue = frame.method.queue
        self._step()

    def _step(self):
        """One consume/cancel cycle, then schedule the next."""
        if self.cycle >= CYCLES:
            self.done.set()
            return
        self.cycle += 1
        self.channel.basic_consume(self.queue,
                                   lambda *a: None,
                                   auto_ack=True,
                                   callback=self._on_consume_ok)

    def _on_consume_ok(self, frame):
        tag = frame.method.consumer_tag
        if self.nowait:
            # No callback -> Basic.Cancel(nowait=True) -> no Basic.CancelOk.
            self.channel.basic_cancel(tag)
            self.connection.ioloop.call_later(0, self._step)
        else:
            self.channel.basic_cancel(tag, callback=lambda _f: self._step())


def main():
    print(f'pika {pika.__version__} from {pika.__file__}')
    print(f'{CYCLES} consume/cancel cycles on ONE long-lived channel\n')

    leaky = Repro(nowait=True).run()
    if leaky is None:
        return 1
    report('A. basic_cancel(tag)                 (nowait, no CancelOk)', leaky)

    print()
    clean = Repro(nowait=False).run()
    report('B. basic_cancel(tag, callback=...)   (waits for CancelOk)', clean)

    print(f'\nA retained {len(leaky._cancelled)} + '
          f'{len(leaky._consumers_with_noack)} entries after {CYCLES} cycles; '
          f'B retained {len(clean._cancelled)} + '
          f'{len(clean._consumers_with_noack)}.')

    cap = getattr(Channel, '_MAX_RETAINED_CANCELLED_TAGS', None)
    if cap is None:
        print('Channel has no retention cap: growth is linear in the number '
              'of cancels and is only reclaimed when the channel closes.')
    else:
        print(f'Channel._MAX_RETAINED_CANCELLED_TAGS = {cap}: retention '
              'plateaus at the cap instead of tracking the cycle count.')
    return 0


if __name__ == '__main__':
    sys.exit(main())
```

</details>

Output on `main` at de535881:

```
3000 consume/cancel cycles on ONE long-lived channel

A. basic_cancel(tag)                 (nowait, no CancelOk):
    _consumers            =    0
    _cancelled            = 3000   <-- retained until the channel closes
    _consumers_with_noack = 3000   <-- retained until the channel closes
    bytes held by the tag strings alone = 237000

B. basic_cancel(tag, callback=...)   (waits for CancelOk):
    _consumers            =    0
    _cancelled            =    0
    _consumers_with_noack =    0
    bytes held by the tag strings alone = 0

A retained 3000 + 3000 entries after 3000 cycles; B retained 0 + 0.
Channel has no retention cap: growth is linear in the number of cancels and is only reclaimed when the channel closes.
```

Raising `CYCLES` scales both counters one for one.

## Proposed fix

The entries cannot simply be dropped when the cancel is sent. `_on_deliver` (`pika/channel.py:1270-1277`) consults both containers to reject a delivery the broker had already dispatched for the cancelled consumer; without them that delivery falls through to `LOGGER.error('Unexpected delivery: %r', ...)` and is neither acked nor rejected, so it stays unacknowledged until the channel closes. The retention is correct, only its duration is wrong.

Nor is there a reply to hang the reclaim on. Tying it to the next synchronous round trip does not work either: `_rpc` queues methods in `_blocked` while the channel is blocking, so the point at which the `Basic.Cancel` reaches the wire is not observable where a later reply lands, and a fence taken at reply time can retire a tag whose cancel has not been sent yet.

So bound the retention instead. The window in which a late delivery can arrive is bounded by what the broker had already dispatched, never by how many consumers have been cancelled since, which makes a cap on retained nowait-cancelled tags an approximation of that window rather than an arbitrary trim:

- add `Channel._MAX_RETAINED_CANCELLED_TAGS` (1024) and a `_nowait_cancelled` deque holding nowait-cancelled tags in cancellation order;
- have `basic_cancel`'s nowait path append to it and, past the cap, evict the oldest through the existing `_cleanup_consumer_ref`;
- clear `_cancelled`, `_consumers_with_noack` and `_nowait_cancelled` in `_cleanup`, so the entries are released at channel close rather than surviving on a closed `Channel` until it is collected;
- document the retention contract on `basic_cancel`.

Nothing on the wire, in the public API or in the non-nowait path changes. A cancel that expects a `Basic.CancelOk` still retires its own entry in `_on_cancelok` and is never queued for eviction.

With that applied, the same repro run unchanged reports (A arm; the B arm is 0 in both runs):

```
A. basic_cancel(tag)                 (nowait, no CancelOk):
    _consumers            =    0
    _cancelled            = 1024   <-- capped at 1024
    _consumers_with_noack = 1024   <-- capped at 1024
    bytes held by the tag strings alone = 80896

A retained 1024 + 1024 entries after 3000 cycles; B retained 0 + 0.
Channel._MAX_RETAINED_CANCELLED_TAGS = 1024: retention plateaus at the cap instead of tracking the cycle count.
```

5000 and 50000 cycles both report 1024 as well. Full suite (1511 passed, 61 skipped), yapf, ruff, docformatter and mypy are clean, with six unit tests covering retention after a single nowait cancel, FIFO eviction past the cap, the plateau at the default cap, reuse of an evicted tag by `basic_consume`, non-eviction of a cancel that expects `Basic.CancelOk`, and the `_cleanup` release.

## Environment

- pika `1.4.0` (`main`, at de535881)
- Python 3.14.4, Linux
- RabbitMQ 4.x

alonfaraj:
## Proposed Changes

Bounds how long `pika.channel.Channel` retains a consumer tag cancelled with nowait. #1691 has the analysis and the reproduction; this describes the change itself.

- `_MAX_RETAINED_CANCELLED_TAGS` (1024) and a `_nowait_cancelled` deque holding nowait-cancelled tags in cancellation order.
- `basic_cancel`'s nowait path appends to that deque and, past the cap, evicts the oldest through the existing `_cleanup_consumer_ref`, so eviction reuses the one function that already knows how to release consumer state.
- `_cleanup` now clears `_cancelled`, `_consumers_with_noack` and `_nowait_cancelled`, so a closed `Channel` that a caller still holds a reference to no longer pins them.
- `basic_cancel`'s docstring states the retention contract, replacing the inline comment that described the old behaviour as unavoidable.

Three invariants a reviewer may want to check:

- only the nowait path queues a tag. A cancel that expects `Basic.CancelOk` retires its own entry in `_on_cancelok` and is never in the deque, so it cannot be evicted while its reply is still in flight.
- a tag cannot be queued twice. `basic_consume` refuses a tag that is still in `_cancelled`, and on the nowait path only eviction takes it back out, so the deque cannot hold two entries for one tag and evicting an old entry cannot clobber a live consumer that reused the tag.
- eviction runs synchronously in `basic_cancel`, on the same thread that mutates the containers everywhere else. No new callback, timer or thread is involved.

Nothing on the wire changes: the frame sent is still `Basic.Cancel(nowait=True)`, and the non-nowait path is untouched.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

Six unit tests in `tests/unit/channel_tests.py` cover: the tag is still retained after a single nowait cancel (the behaviour `_on_deliver` depends on), eviction is in cancellation order past the cap, retention plateaus at the default cap, an evicted tag can be reused by `basic_consume`, a cancel with a callback is never queued for eviction, and `_cleanup` releases all consumer state.

Verified locally against RabbitMQ 4.x on Python 3.14.4 / Linux: full suite 1511 passed and 61 skipped, plus `hatch run fmt-check`, `lint-check`, `docfmt-check` and `typecheck` all clean.

The repro from #1691, run unchanged on this branch, now reports (A arm; the B arm is 0 before and after):

```
A. basic_cancel(tag)                 (nowait, no CancelOk):
    _consumers            =    0
    _cancelled            = 1024   <-- capped at 1024
    _consumers_with_noack = 1024   <-- capped at 1024
    bytes held by the tag strings alone = 80896

A retained 1024 + 1024 entries after 3000 cycles; B retained 0 + 0.
Channel._MAX_RETAINED_CANCELLED_TAGS = 1024: retention plateaus at the cap instead of tracking the cycle count.
```

Those lines read 3000, 3000 and 237000 on `main`. 5000 and 50000 cycles report 1024 as well, so retention stops tracking the cancel count entirely.

## Fixes

- Fixes #1691

### DIFF
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -82,6 +82,10 @@ class Channel:
 
     _ON_CHANNEL_CLEANUP_CB_KEY = '_on_channel_cleanup'
 
+    # Upper bound on the number of nowait-cancelled consumer tags retained in
+    # `_cancelled`; see `_retain_cancelled_tag`.
+    _MAX_RETAINED_CANCELLED_TAGS = 1024
+
     def __init__(self, connection: Connection, channel_number: int,
                  on_open_callback: _OnOpenCallback | None) -> None:
         """
@@ -110,6 +114,9 @@ class Channel:
         self._blocking: Any | None = None
         self._has_on_flow_callback: bool = False
         self._cancelled: set = set()
+        # Nowait-cancelled tags in cancellation order, bounding how many of
+        # them `_cancelled` retains; see `_retain_cancelled_tag`.
+        self._nowait_cancelled: deque = deque()
         self._consumers: dict = {}
         self._consumers_with_noack: set = set()
         self._on_flowok_callback: Callable[..., Any] | None = None
@@ -312,6 +319,12 @@ class Channel:
         basic.cancel from the client). This allows clients to be notified of the loss of consumers
         due to events such as queue deletion.
 
+        A cancelled consumer's tag is retained so that a delivery already in flight when the cancel
+        was sent is rejected rather than left unacknowledged. A cancel that expects a Basic.CancelOk
+        retires its own entry on arrival; a nowait cancel has no reply to retire it, so at most
+        `_MAX_RETAINED_CANCELLED_TAGS` of those are kept and the oldest are dropped beyond that (see
+        `_retain_cancelled_tag`).
+
         :param consumer_tag: Identifier for the consumer
         :param callback: callback(pika.frame.Method) for method Basic.CancelOk. If None, do not
             expect a Basic.CancelOk response, otherwise, callback must be callable
@@ -339,9 +352,7 @@ class Channel:
 
         if nowait:
             # This is our last opportunity while the channel is open to remove
-            # this consumer callback and help gc; unfortunately, this consumer's
-            # self._cancelled and self._consumers_with_noack (if any) entries
-            # will persist until the channel is closed.
+            # this consumer callback and help gc.
             del self._consumers[consumer_tag]
 
         if callback is not None:
@@ -349,6 +360,9 @@ class Channel:
                                callback)
 
         self._cancelled.add(consumer_tag)
+
+        if nowait:
+            self._retain_cancelled_tag(consumer_tag)
 
         self._rpc(spec.Basic.Cancel(consumer_tag=consumer_tag, nowait=nowait),
                   self._on_cancelok if not nowait else None,
@@ -1103,6 +1117,9 @@ class Channel:
         self.callbacks.process(self.channel_number,
                                self._ON_CHANNEL_CLEANUP_CB_KEY, self, self)
         self._consumers = {}
+        self._cancelled = set()
+        self._nowait_cancelled = deque()
+        self._consumers_with_noack = set()
         self.callbacks.cleanup(str(self.channel_number))
         self._cookie = None
 
@@ -1115,6 +1132,30 @@ class Channel:
         self._consumers_with_noack.discard(consumer_tag)
         self._consumers.pop(consumer_tag, None)
         self._cancelled.discard(consumer_tag)
+
+    def _retain_cancelled_tag(self, consumer_tag: str) -> None:
+        """
+        Retain a nowait-cancelled consumer tag, evicting the oldest once the cap is reached.
+
+        A cancelled tag stays in `_cancelled` (and `_consumers_with_noack`) so that a delivery the
+        broker had already dispatched when `Basic.Cancel` was sent is rejected rather than left
+        unacknowledged; see `_on_deliver`.  A cancel that expects a `Basic.CancelOk` retires its own
+        entry in `_on_cancelok`, but a nowait cancel asks the broker for no reply, so nothing
+        retires it and the entries would otherwise accumulate for the life of the channel.
+
+        Bound them here instead.  The window in which a late delivery can still arrive is bounded by
+        what the broker had already dispatched, never by how many consumers have been cancelled
+        since, so evicting in cancellation order past `_MAX_RETAINED_CANCELLED_TAGS` leaves the
+        reject behaviour intact while keeping retention constant in the number of cancels.
+
+        A tag cannot be queued twice: `basic_consume` refuses a tag that is still in `_cancelled`,
+        and on the nowait path only eviction takes it back out again.
+
+        :param consumer_tag: The consumer tag just cancelled with nowait
+        """
+        self._nowait_cancelled.append(consumer_tag)
+        while len(self._nowait_cancelled) > self._MAX_RETAINED_CANCELLED_TAGS:
+            self._cleanup_consumer_ref(self._nowait_cancelled.popleft())
 
     def _get_cookie(self) -> Any:
         """
--- a/pika/channel.py
+++ b/pika/channel.py
@@ -321,9 +321,13 @@ class Channel:
 
         A cancelled consumer's tag is retained so that a delivery already in flight when the cancel
         was sent is rejected rather than left unacknowledged. A cancel that expects a Basic.CancelOk
-        retires its own entry on arrival; a nowait cancel has no reply to retire it, so at most
-        `_MAX_RETAINED_CANCELLED_TAGS` of those are kept and the oldest are dropped beyond that (see
-        `_retain_cancelled_tag`).
+        retires its own entry on arrival; a nowait cancel has no reply to retire it, so its tag is
+        kept only until `_MAX_RETAINED_CANCELLED_TAGS` later nowait cancels displace it (see
+        `_retain_cancelled_tag`). Past that point its retention lapses: a straggling delivery for it
+        is logged as unexpected and left unacknowledged rather than rejected, and the tag may be
+        reused by `basic_consume`, so reusing the same explicit consumer_tag risks a delivery for
+        the old consumer reaching the new one. The cap is a generous approximation of the broker's
+        in-flight window, so this only bites under extreme nowait-cancel churn.
 
         :param consumer_tag: Identifier for the consumer
         :param callback: callback(pika.frame.Method) for method Basic.CancelOk. If None, do not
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
> [!NOTE]

LGTM; CI is green (59/59) at e039a3b8.

The cap bounds `_cancelled` / `_consumers_with_noack` / `_nowait_cancelled` growth on nowait cancels, evicting oldest-first through `_cleanup_consumer_ref` and clearing all three in `_cleanup`. Invariants hold: sync cancels never enter the deque, a tag can't be double-queued, and the non-nowait path is unchanged.

Approving with the documented tradeoff understood: past `_MAX_RETAINED_CANCELLED_TAGS`, an in-flight delivery for an evicted tag is logged as unexpected and left unacknowledged rather than rejected, and the tag becomes reusable (reusing an explicit tag then risks a straggling delivery reaching the new consumer). That is inherent to any bounded approach - the alternative is the unbounded growth this fixes - occurs only under extreme nowait-cancel churn, and the docstrings now state it accurately.

PR #1694, opened 2026-08-19
Fix: Stop `ThreadSafeChannel` confirm tags drifting from the broker

### LINKED ISSUE #1683: `ThreadSafeChannel` publisher-confirm tags drift out of sync with the broker
`ThreadSafeChannel.confirm_delivery()` arms the delivery-tag counter on the **calling** thread, after the RPC returns:

https://github.com/pika/pika/blob/141f8457/pika/adapters/thread_safe_connection.py#L875

```python
result = self._blocking_rpc(
    'confirm_delivery',
    self._channel.confirm_delivery,
    timeout,
    ack_nack_callback=_wrapped_ack_nack,
)
self._next_publish_seq_no = 0   # <-- calling thread
```

But `Confirm.Select` is written on the IOLoop thread, and `_publish` increments that counter on the IOLoop thread too. The broker starts numbering at the first publish it processes after `Confirm.Select`; the wrapper starts counting one network round trip later.

Any publish issued during that window gets a real delivery tag from the broker but is never counted by the wrapper, so **every tag the wrapper reports afterwards is permanently off**.

This is not a narrow race. The window is the full `Confirm.Select` -> `Confirm.SelectOk` round trip, and it is hit by exactly the pattern this class exists for: one thread enabling confirms while another publishes.

### Impact

An application that tracks in-flight messages by the tag from `on_publish` will confirm the wrong message when `Basic.Ack` arrives, and one message gets no `on_publish` callback at all, so it is never recorded as in flight.

### Second, related bug

The idempotency guard at the top of `confirm_delivery` (line 859) is unlocked:

```python
if self._confirm_select_ok is not None:
    return self._confirm_select_ok
```

Two threads both pass it, both send `Confirm.Select`, and both run `self._next_publish_seq_no = 0`. The second assignment resets a counter that is already tracking live publishes. The docstring promises this method is idempotent, on a class whose whole purpose is thread safety.

### Reproduction

Needs RabbitMQ on localhost:5672.

```python
"""Reproduces two confirm_delivery bugs in ThreadSafeConnection."""

import threading
import time

import pika
from pika.adapters.thread_safe_connection import ThreadSafeConnection

QUEUE = 'confirm-drift-demo'
PARAMS = pika.ConnectionParameters('localhost')


def bug_1_tags_drift():
    print('--- Bug 1: publish during the confirm_delivery round trip ---')
    conn = ThreadSafeConnection(PARAMS)
    ch = conn.channel()

    acked, reported = [], []

    # Signals when Confirm.Select reaches the wire.  This only makes the timing
    # deterministic; any thread publishing while another thread is inside
    # confirm_delivery lands in exactly the same window.
    raw = ch._channel
    real_confirm_delivery = raw.confirm_delivery
    select_sent = threading.Event()

    def hooked(*args, **kwargs):
        result = real_confirm_delivery(*args, **kwargs)
        select_sent.set()
        return result

    raw.confirm_delivery = hooked

    def publish(body):
        ch.basic_publish('', QUEUE, body,
                         on_publish=lambda tag, b=body: reported.append(
                             (b.decode(), tag)))

    enabler = threading.Thread(
        target=ch.confirm_delivery,
        args=(lambda frame: acked.append(frame.method.delivery_tag),))
    enabler.start()

    select_sent.wait(5)
    publish(b'A')  # published during the round trip
    enabler.join()

    publish(b'B')
    publish(b'C')
    time.sleep(1)

    print(f'broker Basic.Ack tags : {sorted(acked)}')
    print(f'on_publish tags       : {reported}')
    print(f'next_publish_seq_no   : {ch.next_publish_seq_no}')
    ch.close()
    conn.close()


def bug_2_racing_callers():
    print()
    print('--- Bug 2: two threads calling confirm_delivery ---')
    conn = ThreadSafeConnection(PARAMS)
    ch = conn.channel()

    raw = ch._channel
    real_confirm_delivery = raw.confirm_delivery
    frames = []

    def counted(*args, **kwargs):
        frames.append(1)
        return real_confirm_delivery(*args, **kwargs)

    raw.confirm_delivery = counted

    barrier = threading.Barrier(2)

    def enable():
        barrier.wait(5)
        ch.confirm_delivery(lambda frame: None)

    threads = [threading.Thread(target=enable) for _ in range(2)]
    for thread in threads:
        thread.start()
    for thread in threads:
        thread.join()

    print(f'Confirm.Select frames sent : {len(frames)}  '
          f'(docstring promises 1)')
    ch.close()
    conn.close()


if __name__ == '__main__':
    bug_1_tags_drift()
    bug_2_racing_callers()

    cleanup = ThreadSafeConnection(PARAMS)
    cleanup.channel().queue_delete(queue=QUEUE)
    cleanup.close()
```

### Actual output on main (141f8457)

```
--- Bug 1: publish during the confirm_delivery round trip ---
broker Basic.Ack tags : [1, 2, 3]
on_publish tags       : [('B', 1), ('C', 2)]
next_publish_seq_no   : 3

--- Bug 2: two threads calling confirm_delivery ---
Confirm.Select frames sent : 2  (docstring promises 1)
```

The broker confirmed three messages, tags 1 to 3. Message `A` got tag 1 from the broker and no `on_publish` callback at all. `B` is reported as tag 1 when the broker called it 2, and `C` as 2 when the broker called it 3. An application acking by these tags clears the wrong message every time.

### Expected output

```
--- Bug 1: publish during the confirm_delivery round trip ---
broker Basic.Ack tags : [1, 2, 3]
on_publish tags       : [('A', 1), ('B', 2), ('C', 3)]
next_publish_seq_no   : 4

--- Bug 2: two threads calling confirm_delivery ---
Confirm.Select frames sent : 1  (docstring promises 1)
```

### Suggested fix

Arm the counter on the IOLoop thread, immediately after `Confirm.Select` is written, so it is ordered ahead of every publish the IOLoop has queued behind it. That keeps every write to `_next_publish_seq_no` on the IOLoop thread, the same invariant #1679 established for the close path. Serializing `confirm_delivery` on a per-channel lock then makes the idempotency guard hold across threads as documented.

I have a patch and a before/after harness for both bugs, happy to open a PR.

### Environment

- pika: main at 141f8457
- Python 3.14.4, Linux
- RabbitMQ 4.3.4, Erlang 27.3.4.16

alonfaraj:
## Proposed Changes

`ThreadSafeChannel.confirm_delivery()` armed the delivery-tag counter on the **calling** thread, after the RPC returned. But `Confirm.Select` is written on the IOLoop thread, and `basic_publish`'s `_publish` closure increments that same counter on the IOLoop thread too. The broker starts numbering at the first publish it processes after `Confirm.Select`; the wrapper started counting a full network round trip later.

Every publish issued in that window was numbered by the broker and skipped by the wrapper, so **every tag reported afterwards was permanently behind the broker's**, and the message published during the round trip got no `on_publish` callback at all. An application correlating confirms by that tag acks the wrong message. This is not a narrow race: the window is the whole `Confirm.Select` -> `Confirm.SelectOk` round trip, and it is hit by exactly the pattern this class exists for, one thread enabling confirms while another publishes.

The idempotency guard at the top of the method was also unlocked, so two threads could both find `_confirm_select_ok` unset, both send `Confirm.Select`, and both arm the counter. The second arming reset a counter that was already numbering live publishes, handing the same tags out twice. The docstring promises idempotency, on a class whose whole purpose is thread safety.

### The fix

**Arm the counter on the IOLoop thread.** `_blocking_rpc` gains an `on_sent` hook that runs on the IOLoop thread the moment `channel_method` writes its frame, inside the same callback and therefore ahead of every publish the IOLoop has queued behind it. `confirm_delivery` arms `_next_publish_seq_no` from there. Every write to the counter is now on the IOLoop thread and the calling thread only reads it, the same invariant #1679 established for the close path.

Arming happens only once. A `confirm_delivery` that timed out may still have reached the broker, and the retry's second `Confirm.Select` does not reset the broker's counter, so neither may the wrapper's.

**Serialize `confirm_delivery` on a per-channel lock** held across the guard and the RPC, so the second caller waits and is then handed the cached `Confirm.SelectOk`. A per-channel lock rather than the connection-wide `_channel_waiters_lock`, because the call blocks for a broker round trip and must not stall operations on other channels.

### Before and after

Publishing `A` during the round trip, then `B` and `C`, against a broker that numbered all three 1 to 3:

|        | `on_publish` tags                | `next_publish_seq_no` |
| ------ | -------------------------------- | --------------------- |
| before | `[('B', 1), ('C', 2)]`           | 3                     |
| after  | `[('A', 1), ('B', 2), ('C', 3)]` | 4                     |

## Types of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Cosmetics

## Checklist

### What was tested

Full suite against a local RabbitMQ: 1512 passed, 61 skipped. `hatch run fmt-check`, `lint-check`, `docfmt-check` and `typecheck` are all clean.

Four new tests, each confirmed failing against unfixed `main` and passing with the fix:

- `test_confirm_delivery_arms_the_counter_before_the_response` pins that the counter is armed while `Confirm.SelectOk` is still in flight.
- `test_confirm_delivery_holds_its_lock_across_the_rpc` pins that a second caller cannot pass the guard mid round trip. Checked from inside the raw call, which runs while the lock is held, rather than by racing two threads and hoping to catch the window.
- `test_confirm_delivery_retry_does_not_rearm_the_counter` pins that a retry after a timeout does not reset a counter the broker never reset.
- `TestPublishDuringConfirmSelectRoundTrip` publishes through a real `Confirm.Select` round trip and compares the tags handed to `on_publish` against the tags the broker confirms. Against unfixed `main` it reproduces the table above exactly. The broker may coalesce confirms with `multiple=True`, so it asserts how far the confirmations reached rather than counting individual tags.

## Fixes

- Fixes #1683

### DIFF
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -380,6 +380,12 @@ class Channel:
         self._pool_shutdown = False
         self._next_publish_seq_no: int | None = None
         self._confirm_select_ok = None
+        # Serializes confirm_delivery so two threads cannot both pass the
+        # _confirm_select_ok guard while the first is still waiting for
+        # Confirm.SelectOk.  Per-channel rather than the connection-wide
+        # _channel_waiters_lock: the call blocks for a broker round trip, which
+        # must not stall operations on other channels.
+        self._confirm_lock = threading.Lock()
 
     def _check_not_closed(self) -> None:
         """
@@ -943,10 +949,17 @@ class Channel:
         """
         Enable publisher confirms and block until Confirm.SelectOk arrives.
 
-        Idempotent: calling this method more than once on the same channel
-        returns the original Confirm.SelectOk without sending a frame to
-        the broker or resetting the delivery-tag counter, matching the
+        Idempotent: calling this method more than once on the same
+        channel - concurrently from several threads included - returns
+        the original Confirm.SelectOk without sending a frame to the
+        broker or resetting the delivery-tag counter, matching the
         behavior of the RabbitMQ Java and .NET clients.
+
+        Concurrent callers are serialized on a per-channel lock, so a
+        second thread waits for the first call's round trip to finish
+        before it is handed the cached frame.  Its own *timeout* bounds
+        only the wait for the broker's response, not that queueing, so
+        it can block for the first caller's timeout on top of its own.
 
         The *ack_nack_callback* is dispatched on the channel's worker
         thread (same as delivery callbacks), not the IOLoop thread.
@@ -966,39 +979,40 @@ class Channel:
         :raises Exception: if the connection is closed before the response arrives.
         :raises TimeoutError: if *timeout* expires before the response arrives.
         """
-        if self._confirm_select_ok is not None:
-            return self._confirm_select_ok
+        with self._confirm_lock:
+            if self._confirm_select_ok is not None:
+                return self._confirm_select_ok
 
-        def _wrapped_ack_nack(method_frame) -> None:
-            _submit_or_terminate(
-                self._consumer_work_pool, self._wrapper._connection,
-                'Publisher confirm dropped: work pool shut down',
-                self._safe_dispatch, 'publisher confirm callback',
-                ack_nack_callback, method_frame)
+            def _wrapped_ack_nack(method_frame) -> None:
+                _submit_or_terminate(
+                    self._consumer_work_pool, self._wrapper._connection,
+                    'Publisher confirm dropped: work pool shut down',
+                    self._safe_dispatch, 'publisher confirm callback',
+                    ack_nack_callback, method_frame)
 
-        def _arm_seq_no() -> None:
-            # Runs on the IOLoop thread the moment Confirm.Select is written,
-            # so it is ordered ahead of every publish the IOLoop has queued
-            # behind it.  Arming from the calling thread once the RPC returns
-            # instead would miss each publish issued during the round trip: the
-            # broker numbers those from 1, leaving the counter permanently
-            # behind the broker's delivery tags.
-            #
-            # Arm only once.  A confirm_delivery that timed out may still have
-            # reached the broker, and the retry's second Confirm.Select does not
-            # reset the broker's counter, so neither may this.
-            if self._next_publish_seq_no is None:
-                self._next_publish_seq_no = 0
+            def _arm_seq_no() -> None:
+                # Runs on the IOLoop thread the moment Confirm.Select is
+                # written, so it is ordered ahead of every publish the IOLoop
+                # has queued behind it.  Arming from the calling thread once the
+                # RPC returns instead would miss each publish issued during the
+                # round trip: the broker numbers those from 1, leaving the
+                # counter permanently behind the broker's delivery tags.
+                #
+                # Arm only once.  A confirm_delivery that timed out may still
+                # have reached the broker, and the retry's second Confirm.Select
+                # does not reset the broker's counter, so neither may this.
+                if self._next_publish_seq_no is None:
+                    self._next_publish_seq_no = 0
 
-        result = self._blocking_rpc(
-            'confirm_delivery',
-            self._channel.confirm_delivery,
-            timeout,
-            on_sent=_arm_seq_no,
-            ack_nack_callback=_wrapped_ack_nack,
-        )
-        self._confirm_select_ok = result
-        return result
+            result = self._blocking_rpc(
+                'confirm_delivery',
+                self._channel.confirm_delivery,
+                timeout,
+                on_sent=_arm_seq_no,
+                ack_nack_callback=_wrapped_ack_nack,
+            )
+            self._confirm_select_ok = result
+            return result
 
     def basic_consume(self,
                       queue,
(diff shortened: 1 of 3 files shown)


lukebakken (approved):
LGTM at 1866c430.

The core fix is correct and well-tested: arming the delivery-tag counter atomically inside the same IOLoop callback that writes Confirm.Select, ahead of any queued publish, and serializing confirm_delivery under a per-channel lock. It closes the tag drift in both scheduling orders and the acceptance test exercises the real broker.

- Moved the `_blocking_rpc` `on_sent` hook out of the frame-send `try`, so a hook that raises after `channel_method` delivered its response synchronously can no longer mask that result (which the waiter re-raises ahead of it). Harmless for the current `_arm_seq_no` assignment, but removes the latent trap. Covered by two new unit tests.
- Documented that a `confirm_delivery` `TimeoutError` does not mean confirms are off: the Confirm.Select frame may have reached the broker, so the counter stays armed and later publishes still fire `on_publish` and advance `next_publish_seq_no`. Callers should retry rather than assume disabled.
- Added an acceptance test that retries `confirm_delivery` against a real broker, proving the raw channel accepts a second Confirm.Select without raising and the counter survives the retry, closing the gap left by the mock-only retry unit test.

CI is green (60/60); the single Windows failure was the transient RabbitMQ/Erlang version-resolution setup step and passed on rerun.

PR #1695, opened 2026-08-19
Check `pika/spec.py` against `codegen.py` output in CI

### LINKED ISSUE #1677: suggestion: Enforce spec regen in CI
## Problem

Nothing in CI checks that `pika/spec.py` matches what `utils/codegen.py` would produce. The rule lives in `AGENTS.md` and in the generated file's own docstring, but it is enforced by review only, so two kinds of drift can land:

- `utils/codegen.py` changes without a regenerated `pika/spec.py`
- `pika/spec.py` edited by hand with no `utils/codegen.py` change

The regeneration recipe also has a second step that is easy to miss: `fmt` excludes `pika/spec.py`, so yapf has to be run against it directly or the result differs from the committed file by line wrapping alone.

## Proposal

Add a `spec-check` script plus a workflow that runs it, same shape as `lint.yaml` and `mypy.yaml` (thin wrapper around `hatch run ...`) so local and CI behavior cannot drift.

The check regenerates into a temp dir and diffs against the committed file:

1. Fetch the two files codegen needs (`amqp_codegen.py`, `amqp-rabbitmq-0.9.1.json`) from `rabbitmq/rabbitmq-server` at a pinned SHA. Both are stdlib-only and total 38 KB, so no clone of that repo.
2. Run `utils/codegen.py` with cwd set to a temp tree. It hardcodes `./pika/spec.py`, so cwd redirects the output without touching the working tree.
3. Run `yapf --in-place --style google` on the result.
4. Compare against `pika/spec.py`. On mismatch, print a unified diff and exit 1 telling the author to run `hatch run spec-regen`.

Same script without `--check` regenerates in place, so the fix a failing PR is told to run is the same code path CI just ran.

Two details worth calling out:

- **Pin the upstream ref.** Otherwise a spec change in rabbitmq-server starts failing unrelated PRs. Bumping the pin becomes a deliberate commit that carries the regenerated `spec.py` with it.
- **Run on every PR, not path-filtered.** GitHub required status checks never report on skipped path-filtered jobs, so the PR sits pending forever. Running unconditionally also catches the hand-edited `spec.py` case. Costs about 20 seconds.

Files touched: new `utils/regen_spec.py`, two scripts in `pyproject.toml` (`spec-regen`, `spec-check`), new `.github/workflows/codegen.yaml`.

## Open questions if this suggestion land

- Pin as a SHA constant in `utils/regen_spec.py`, or in a separate file so a bump is an obvious one-line diff? or just using latest?
- Should `codegen` be added to the required checks / the `tests-passed` gate, or left as a standalone workflow like `lint` and `mypy`?

alonfaraj:
## Proposed Changes

Nothing in CI verified that the committed `pika/spec.py` matches what `utils/codegen.py` produces. The rule lives in `AGENTS.md` and in the generated file's own docstring, but it is enforced by review alone, so two kinds of drift can land: a `utils/codegen.py` change without a regenerated `pika/spec.py`, and a `pika/spec.py` edited by hand with no generator change.

This adds `utils/regen_spec.py`, a single code path with two modes:

- It downloads `amqp_codegen.py` and `amqp-rabbitmq-0.9.1.json` from `rabbitmq/rabbitmq-server` into a temporary tree laid out the way the generator expects. Both are stdlib-only and total about 38 KB, so there is no clone of that repository and no new dependency (`urllib` from the stdlib).
- It runs `utils/codegen.py` with its working directory set to that temporary tree. The generator hardcodes its output path as `./pika/spec.py`, so the cwd redirects the output and the working copy is never touched until we decide what to do with the result.
- It runs `yapf --in-place --style google` on the generated file. This is the step that is easy to miss by hand: `fmt` excludes `pika/spec.py`, so a manual regeneration produces output that differs from the committed file by line wrapping alone.
- With `--check` it prints a unified diff and exits 1; without it, it writes `pika/spec.py` in place.

Two `hatch` scripts wrap it, alongside `typecheck`:

```
spec-regen = "python utils/regen_spec.py"
spec-check = "python utils/regen_spec.py --check"
```

Because both modes are the same code, the fix a failing pull request is told to run is exactly the code path CI just ran.

`.github/workflows/codegen.yaml` runs `hatch run spec-check` on push and pull request, in the same thin shape as `lint.yaml` and `mypy.yaml`, so local and CI behavior cannot drift.

## Types of Changes

- [ ] Bugfix
- [x] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Cosmetics

## Checklist

The check is its own test, so there is no new unit test. What was verified locally:

- `hatch run spec-check` reproduces the committed `pika/spec.py` byte for byte and exits 0.
- Hand-editing a constant in `pika/spec.py` makes it print the diff and the remediation message, and exit 1.
- `hatch run spec-regen` leaves `pika/spec.py` unchanged.
- `python utils/regen_spec.py --check --ref dd24a82798f3b98f4116c073480476b8a674e92f` also matches, confirming the flag works against an explicit revision.
- `hatch run unit` (947 passed, 21 skipped), `lint-check`, `fmt-check`, `docfmt-check`, and `typecheck` all pass. `utils/regen_spec.py` is deliberately not added to ruff's `extend-exclude`: unlike `utils/codegen.py` it is ordinary code and should be linted.

The whole check takes about a second of work beyond installing hatch.

## Fixes

- Fixes #1677

## Further Comments

Two decisions worth calling out, both answering open questions in the issue.

**The upstream ref is not pinned.** `DEFAULT_REF` is `main`, for the same reason `AGENTS.md` already gives for leaving ruff unpinned: when upstream changes the AMQP spec, CI should say so on the next push instead of letting `pika/spec.py` sit quietly behind. A `--ref` flag generates from another revision when investigating, so a pin is never needed just to reproduce something. The cost of this choice is real and worth stating: unlike a ruff failure, which the author fixes by editing their own code, an upstream spec change fails whichever pull request is open next and asks that author to commit a `spec.py` diff unrelated to their work. Given how rarely the spec json actually changes (the last change to the generated output was the FRAME-MIN-SIZE bump in March 2025; everything since is copyright-line noise that never reaches `spec.py`), that seems like the right trade. The failure message names the ref it generated against and points at `AGENTS.md` so the author can tell the two cases apart, and `AGENTS.md` says an upstream-driven regeneration belongs in its own commit rather than buried in an unrelated pull request. If it ever does become annoying, pinning is a one-line change to the default.

**The workflow is not path-filtered.** GitHub required status checks never report on a skipped path-filtered job, so the pull request would sit pending forever. Running unconditionally is also what catches the hand-edited `spec.py` case, which a filter on `utils/codegen.py` would miss entirely.

**Gate placement.** Left standalone, like `lint` and `mypy`, rather than folded into the `tests-passed` gate. That gate exists to collapse a large matrix plus docs into one required check; a single fast job does not need the indirection, and folding it in would couple spec verification to the test matrix's lifecycle. If it should be required, adding `codegen / spec regen check` to branch protection directly is enough.

Docs updated alongside the code: the auto-generated code section of `AGENTS.md` now describes the supported recipe and the no-pin rule, the CI workflow list gains a `codegen` entry, the repository layout lists `utils/regen_spec.py`, and the `utils/codegen.py` docstring points at the wrapper while keeping the manual clone path as a fallback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lukebakken (approved):
> [!NOTE]

LGTM; CI is green (62/62) at ffe5d255.

The `spec-check` gate is correct: it fetches the upstream codegen inputs, runs `utils/codegen.py` plus yapf in a temp tree, and diffs against the committed `pika/spec.py`, failing on drift or a hand-edit. I confirmed the happy path end-to-end against live `main` (`hatch run spec-check` -> up to date).

Review follow-ups addressed in ffe5d255: `_fetch` now retries transient upstream failures with backoff and fails fast on a non-retryable status, so a momentary `raw.githubusercontent.com` hiccup no longer flakes the required check with a traceback (it exits 2 with a clear message, distinct from the exit-1 drift path); and the two `:param ref:` docstrings were reworded so they render correctly under docformatter.

The unpinned ref, live-`main` fetch, and unconditional (non-path-filtered) gate are the intended design - the same "surface drift as a CI failure" philosophy as the unpinned linters - and are documented in AGENTS.md, so I did not treat those as defects.

PR #1697, opened 2026-08-23
Fix `Connection.close()` self-deadlock from a consumer callback

### LINKED ISSUE #1686: `Connection.close()` called from a consumer callback self-deadlocks
### Summary

`pika.adapters.thread_safe_connection.Connection.close()` deadlocks against its own IOLoop cleanup tail when it is called from a consumer callback. The consumer worker thread blocks joining the IOLoop thread, while the IOLoop thread blocks joining that same consumer worker.

With the default `timeout=10` the deadlock is broken only by burning the close timeout budget twice, so a close with nothing pending takes 20 seconds and logs a false "IOLoop thread did not exit" warning. With `close(timeout=None)` it never returns.

This is the pattern the class exists to support. `Channel`'s docstring states that all `Channel` methods are safe to call from within a delivery callback, and consume-until-sentinel-then-close is the natural next step from there.

### Root cause

`Connection.close()` exempts only the IOLoop thread from the join (`pika/adapters/thread_safe_connection.py:1727`):

```python
def close(self, timeout: float | None = 10) -> None:
    if threading.current_thread() is self._ioloop_thread:
        try:
            self._connection.close()
        except Exception:
            LOGGER.debug('connection.close() raised from IOLoop thread', exc_info=True)
        return
    ...
    self._connection.ioloop.add_callback_threadsafe(_safe_close)
    self._ioloop_thread.join(timeout=timeout)     # line 1748
```

A consumer callback runs on the per-channel `pika-consumer` worker thread, which gets no such exemption, so it reaches the join at line 1748.

Meanwhile the IOLoop thread, once `ioloop.start()` returns, runs its cleanup tail (line 1525):

```python
# IOLoop has exited - safe to block waiting for pool workers.
self._shutdown_all_consumer_pools()
self._shutdown_connection_pool()
```

`_shutdown_all_consumer_pools()` is called with `timeout=None`, which takes the unbounded branch (line 1596) and calls `ch._shutdown_pool()` per channel, which calls `_BoundedWorkPool.shutdown(wait=True, timeout=None)`, which joins the worker thread (line 312).

The cycle is therefore:

- consumer worker: inside the user's callback, inside `Connection.close()`, joining `_ioloop_thread`
- IOLoop thread: past `ioloop.start()`, inside `_shutdown_all_consumer_pools()`, joining the consumer worker

`_BoundedWorkPool.shutdown` already has a self-join guard for exactly this shape (line 305), and `Channel.close()` inherits it:

```python
if thread is threading.current_thread():
    # A delivery callback is shutting down its own pool (e.g. it called
    # channel.close()).  The worker cannot join itself; the flag is
    # already set, so it exits once this callback returns.
    return True
```

`Connection.close()` has no equivalent guard. It does not recognise that the calling thread is a pool worker whose exit the IOLoop thread is waiting on.

The 20 second figure comes from the two joins in `close()`: the clean join at line 1748 (`timeout`), then the forced join at line 1765 under a fresh `timeout` budget. After both expire the method logs the warning, returns, the user callback returns, the worker finally exits, and the IOLoop thread's join completes.

### Reproduction

```python
"""Connection.close() from a consumer callback self-deadlocks."""
import logging
import threading
import time

import pika

logging.basicConfig(
    level=logging.WARNING,
    format='%(asctime)s %(threadName)s %(levelname)s %(message)s')

QUEUE = 'repro_close_from_consumer_callback'

ch = conn.channel()
ch.basic_publish(exchange='', routing_key=QUEUE, body=b'stop')

done = threading.Event()
elapsed = [None]


def on_message(channel, method, properties, body):
    channel.basic_ack(method.delivery_tag)
    print(f'consumer callback got {body!r}; calling conn.close()', flush=True)
    started = time.monotonic()
    conn.close()  # default timeout=10
    elapsed[0] = time.monotonic() - started
    done.set()


ch.basic_consume(queue=QUEUE, on_message_callback=on_message)

if not done.wait(timeout=120):
    print('conn.close() had not returned after 120 seconds')
else:
    print(f'conn.close() returned after {elapsed[0]:.1f}s '
          f'(nothing was pending; expected ~0s)')
```

Output:

```
consumer callback got b'stop'; calling conn.close()
2026-08-13 12:26:12,645 pika-consumer WARNING Connection 1 IOLoop thread did not exit within the close timeout; abandoning it
conn.close() returned after 20.0s (nothing was pending; expected ~0s)
```

Note that the warning is emitted on `pika-consumer`, the very thread the IOLoop is waiting for, and that it is false: the IOLoop thread does exit, as soon as the caller stops blocking it.

### Reproduction, unbounded variant

The same script with `conn.close(timeout=None)` never returns.

```python
"""Same self-deadlock, unbounded, with Connection.close(timeout=None)."""
import threading
import time

import pika

QUEUE = 'repro_close_from_consumer_callback_unbounded'

ch = conn.channel()
ch.basic_publish(exchange='', routing_key=QUEUE, body=b'stop')

returned = threading.Event()


def on_message(channel, method, properties, body):
    channel.basic_ack(method.delivery_tag)
    print('consumer callback calling conn.close(timeout=None)', flush=True)
    conn.close(timeout=None)
    returned.set()


ch.basic_consume(queue=QUEUE, on_message_callback=on_message)

started = time.monotonic()
if returned.wait(timeout=30):
    print(f'conn.close(timeout=None) returned after '
          f'{time.monotonic() - started:.1f}s')
else:
    print('conn.close(timeout=None) still blocked after 30s: '
          'permanent deadlock')
```

Output:

```
consumer callback calling conn.close(timeout=None)
conn.close(timeout=None) still blocked after 30s: permanent deadlock
```

The process has to be killed; the worker thread is still wedged in the join.

### Expected behaviour

`Connection.close()` from a consumer callback should return promptly, as it does from any other non-IOLoop thread, and should not log a warning when nothing is actually stuck. `close(timeout=None)` should not deadlock.

### Impact

- Any consumer that shuts the connection down from inside its own delivery callback stalls for `2 * timeout` and logs a misleading warning that points at the wrong culprit.
- `close(timeout=None)`, which the docstring documents as "wait indefinitely", hangs permanently in this case.
- The warning invites the wrong diagnosis. It names the IOLoop thread and says its worker is being abandoned, when the actual cause is the caller.

### Suggested fix

Give `Connection.close()` the same self-join guard that `_BoundedWorkPool.shutdown` already has: if the calling thread is one of this connection's pool worker threads, initiate the close (schedule `_safe_close`) and return without joining the IOLoop thread. The worker will exit once the callback returns, at which point the IOLoop tail's join completes on its own.

That requires the connection to be able to recognise its worker threads. `_BoundedWorkPool` already holds `self._thread`, so a predicate over the channel pools plus the connection pool is enough, for example a helper that tests `threading.current_thread()` against each pool's `_thread`.

Documenting the restriction instead would be a weaker fix, since it contradicts the `Channel` docstring's existing promise and leaves `close(timeout=None)` able to hang.

### Environment

- pika 1.4.0, commit 0bbc026f

### Notes

Not covered by the current test suite (938 unit tests pass; `hatch run typecheck` and `hatch run lint-check` are clean). No existing test calls `Connection.close()` from inside a consumer callback.

zanarellidev:
## Proposed Changes

`Connection.close()` (the `ThreadSafeConnection` wrapper) deadlocked when called from inside a consumer's `on_message_callback` or from the connection's own event-callback worker, exactly as described in #1686.

Root cause: `close()` joined `_ioloop_thread` from any calling thread other than the IOLoop thread itself. But the IOLoop thread's own post-`ioloop.start()` cleanup tail joins every pool worker before it exits - including the very worker currently running the callback that called `close()`. That worker was joining the IOLoop thread while the IOLoop thread was joining it back: a hard self-deadlock. With the default `timeout=10` this burned the close budget twice (20s) before giving up and logging a warning that named the wrong thread; with `timeout=None` it never returned at all.

Fix: `close()` now detects when it is running on one of the connection's own pool worker threads (a channel's consumer worker, or the connection's event-callback worker) and, in that case, schedules the close and returns immediately without joining - the same shortcut it already takes when called from the IOLoop thread itself. The worker's own pool already has a self-join guard for this shape (`_BoundedWorkPool.shutdown`), so it exits normally once the callback returns, and the IOLoop thread's join on it then completes on its own.

## Types of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

## Checklist

## Fixes

- Fixes #1686

## Further Comments

Verified against a real broker, not just mocks: reverting only the production fix (keeping the new tests) reproduces the issue's exact reported symptom, including the false "IOLoop thread did not exit within the close timeout" warning, for both the bounded-timeout and `timeout=None` cases. Restoring the fix, both pass and the full existing acceptance and unit suites for this file stay green.

### DIFF
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -1683,6 +1683,29 @@ class Connection:
         for ch in channels:
             ch._shutdown_pool(timeout=max(0.0, deadline - time.monotonic()))
 
+    def _is_pool_worker_thread(self) -> bool:
+        """
+        True if the calling thread is one of this connection's pool worker threads.
+
+        Covers every channel's consumer worker plus the connection's own event-callback worker
+        (:attr:`_connection_work_pool`).  :meth:`close` uses this to detect that it is running from
+        inside a delivery/event callback so it can avoid joining :attr:`_ioloop_thread`: that thread's
+        post-``ioloop.start()`` cleanup tail (:meth:`_shutdown_all_consumer_pools`,
+        :meth:`_shutdown_connection_pool`) joins every pool worker in turn, including this one, so
+        joining it here would deadlock each thread waiting on the other (issue #1686).
+
+        Reading each pool's ``_thread`` without its lock is safe here: a worker only executes this
+        check from within its own run loop, by which point ``_thread`` was already assigned (under
+        the pool's lock, before ``start()``) by the thread that created it - that assignment
+        happens-before anything the worker itself observes.
+        """
+        current = threading.current_thread()
+        if current is self._connection_work_pool._thread:
+            return True
+        with self._channel_waiters_lock:
+            channels = list(self._channels)
+        return any(current is ch._consumer_work_pool._thread for ch in channels)
+
     def _shutdown_connection_pool(self, timeout: float | None = None) -> None:
         """
         Shut down the connection-level event-callback pool.
@@ -1796,6 +1819,14 @@ class Connection:
         callback).  When called from the IOLoop thread the close is initiated synchronously and the
         method returns immediately without joining.
 
+        Also safe to call from a delivery or connection-event callback running on one of this
+        connection's pool worker threads (e.g. a consumer's ``on_message_callback``).  The IOLoop
+        thread's own post-close cleanup joins every pool worker, including whichever one is running
+        the callback that called ``close()``, so joining :attr:`_ioloop_thread` from there would
+        deadlock the two threads against each other. In that case the close is scheduled and this
+        method returns immediately without joining; the worker exits normally once the callback
+        returns.
+
         Calling ``close()`` on an already-closed connection is a no-op.
 
         :param timeout: Seconds to wait for a clean close before force-stopping the IOLoop. Defaults
@@ -1824,6 +1855,18 @@ class Connection:
                     exc_info=True)
 
         self._connection.ioloop.add_callback_threadsafe(_safe_close)
+
+        if self._is_pool_worker_thread():
+            # Called from within a delivery or connection-event callback running on one of this
+            # connection's pool worker threads. The IOLoop thread's post-close cleanup tail
+            # (_shutdown_all_consumer_pools / _shutdown_connection_pool) joins every pool worker,
+            # including this one, once ioloop.start() returns - so joining self._ioloop_thread here
+            # would deadlock: this thread waiting on the IOLoop thread, which is waiting on this
+            # thread to return from the very callback that called close(). The close is already
+            # scheduled above; leave the pool's own self-join guard
+            # (_BoundedWorkPool.shutdown) to let this worker exit normally once the callback returns.
+            return
+
         self._ioloop_thread.join(timeout=timeout)
         if self._ioloop_thread.is_alive():
             self._connection.ioloop.add_callback_threadsafe(
--- a/pika/adapters/thread_safe_connection.py
+++ b/pika/adapters/thread_safe_connection.py
@@ -1825,14 +1825,17 @@ class Connection:
         the callback that called ``close()``, so joining :attr:`_ioloop_thread` from there would
         deadlock the two threads against each other. In that case the close is scheduled and this
         method returns immediately without joining; the worker exits normally once the callback
-        returns.
+        returns.  Because this path does not join the IOLoop thread, *timeout* is not applied on it:
+        the close is best-effort and an unresponsive broker is not force-stopped from here, so a
+        healthy handshake is assumed (see #1706).
 
         Calling ``close()`` on an already-closed connection is a no-op.
 
         :param timeout: Seconds to wait for a clean close before force-stopping the IOLoop. Defaults
             to 10 seconds, which is sufficient for a healthy broker on any reasonable network. The
             pathological force path may wait up to twice this (once for the clean join, once for the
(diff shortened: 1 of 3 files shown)


lukebakken (approved):
LGTM at 63f66979.

The fix for the #1686 self-deadlock is correct and well-scoped. `_is_pool_worker_thread()` checks exactly the thread set the IOLoop cleanup tail joins (the connection event pool plus each tracked channel's consumer pool), the lock-free `_thread` reads are safe as atomic identity comparisons, and the three regression tests genuinely fail without the fix.

- Corrected the return-path comment. The worker does not exit via the `_BoundedWorkPool.shutdown` self-join guard (that fires only when a worker shuts down its own pool); here the IOLoop cleanup tail calls `shutdown()` from the IOLoop thread and joins the worker, which exits via the drained-and-shutdown check in `_run_worker`.
- Documented that `timeout` is not applied on the pool-worker path: the close is best-effort there and an unresponsive broker is not force-stopped.
- Strengthened the negative unit test to point the pools at real but distinct worker threads, so it exercises a genuine thread-identity mismatch rather than comparing against an unset `_thread`.

Filed #1706 to track the deferred force-stop enhancement (restore `timeout` parity on the worker path via an IOLoop `call_later`), which is out of scope for this deadlock fix.

CI is green (60/60).

PR #1699, opened 2026-08-24
Stop an out-of-range `timestamp` header from closing the connection

### LINKED ISSUE #1698: An out-of-range field-table timestamp tears down the connection
### Summary

The AMQP 0-9-1 field-table `timestamp` type (`'T'`) is a 64-bit unsigned integer. `pika.data.decode_value` passes it straight to `datetime.fromtimestamp()` with no guard, so any value past year 9999 raises. The raise happens inside the frame decoder, where nothing expects a non-pika exception, so the transport's generic read-error handler converts it to `StreamLostError` and drops the whole connection.

RabbitMQ neither range-checks nor rejects the field (`rabbit_binary_parser` reads it as `Value:64/unsigned`, `rabbit_binary_generator` writes it back as `<<V:64>>`), so such a header travels end to end. If the message was not acked it comes back on reconnect and the client dies again, so one header becomes an unbreakable crash loop, reported as a network error rather than a decode error.

Values that trigger it are not exotic:

- a millisecond epoch, which plenty of producers put in a header
- a pre-1970 time: the RabbitMQ Java client writes `T` as a signed long (`ValueWriter.writeTimestamp` / `ValueReader.readTimestamp` use `DataInputStream.readLong`), so a negative one arrives as a large u64
- any sentinel near `2**64 - 1`

### Environment

- Present since 955966a7 (2011-03-04), so every pika release. Checked on `main` at 2126d43a, in-tree `__version__` 1.4.0.
- Python 3.14.4, Linux
- No broker needed to reproduce; the script below includes a minimal fake one

### Root cause

`pika/data.py`, in `decode_value`:

```python
# Timestamp
elif kind == _FIELD_TIMESTAMP:
    value = datetime.fromtimestamp(
        _PACK_UNSIGNED_LONG_LONG.unpack_from(encoded, offset)[0],
        timezone.utc)
    offset += 8
```

`datetime` spans years 1 to 9999, i.e. POSIX seconds 0 through 253402300799, a small corner of the u64 range the field allows. Outside it `fromtimestamp` raises, and which exception you get depends on the value and the platform. All three are stdlib exceptions, so no caller can reasonably catch them:

| u64 on the wire | result |
| --- | --- |
| `253402300799` | `datetime(9999, 12, 31, 23, 59, 59)`, the last value that works |
| `253402300800` (first second of year 10000) | `ValueError: year must be in 1..9999, not 10000` |
| `1772000000000` (millisecond epoch) | `ValueError: year must be in 1..9999, not 58122` |
| `9223372036854775807` (`2**63 - 1`) | `OSError: [Errno 75] Value too large for defined data type` |
| `18446744073709551615` (`2**64 - 1`) | `OverflowError: timestamp out of range for platform time_t` |

Traceback:

```
  File "<string>", line 5, in <module>
    spec.BasicProperties().decode(struct.pack('>H', spec.BasicProperties.FLAG_HEADERS) + struct.pack('>I', len(tbl)) + tbl)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alon/repos/pika/pika/spec.py", line 2256, in decode
    (self.headers, offset) = data.decode_table(encoded, offset)
                             ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/alon/repos/pika/pika/data.py", line 207, in decode_table
    value, offset = decode_value(encoded, offset)
                    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/alon/repos/pika/pika/data.py", line 256, in decode_value
    value = datetime.fromtimestamp(
        _PACK_UNSIGNED_LONG_LONG.unpack_from(encoded, offset)[0],
        timezone.utc)
ValueError: year must be in 1..9999, not 58122
```

### Reproduction

```python
"""
Minimal repro: a field-table timestamp pika cannot fit in `datetime` kills the
connection.

The AMQP 0-9-1 field-table `timestamp` type is a u64 (see the RabbitMQ errata),
and RabbitMQ neither range-checks nor rejects one. pika hands the raw value to
`datetime.fromtimestamp()` unguarded, so a value beyond year 9999 raises out of
the frame decoder; the transport relabels it as StreamLostError and the whole
connection goes down. The message is never acked, so it returns on reconnect:
an infinite crash loop.

Usage: python3 repro.py
"""
import logging
import socket
import struct
import threading
import warnings

import pika
from pika import data, frame, spec

warnings.filterwarnings('ignore')  # BlockingConnection deprecation notice
logging.disable(logging.CRITICAL)  # pika logs the same error itself

CASES = [
    (1772000000000, 'a millisecond epoch'),
    (2**64 - 14256000, "1969-07-20 as two's complement, per the Java client"),
]


def content_header(chan, seconds):
    """A Basic content header carrying one `T` field in `headers`."""
    table = b'\x02ts' + b'T' + struct.pack('>Q', seconds)
    props = (struct.pack('>HHQ', spec.BasicProperties.INDEX, 0, 5) +
             struct.pack('>H', spec.BasicProperties.FLAG_HEADERS) +
             struct.pack('>I', len(table)) + table)
    return (struct.pack('>BHI', spec.FRAME_HEADER, chan, len(props)) + props +
            bytes((spec.FRAME_END,)))


def broker(listener, seconds):
    """Hand the same never-acked message to every consumer that subscribes."""
    while True:
        sock, _ = listener.accept()
        buf = bytearray()
        send = lambda f: sock.sendall(f.marshal())  # noqa: E731
        sock.recv(8)  # protocol header
        send(frame.Method(0, spec.Connection.Start(0, 9, {}, 'PLAIN', 'en_US')))
        while True:
            consumed, f = frame.decode_frame(buf, 0)
            if f is None:
                chunk = sock.recv(65536)
                if not chunk:
                    break  # client died; wait for its reconnect
                buf += chunk
                continue
            del buf[:consumed]
            if not isinstance(f, frame.Method):
                continue
            name, chan = f.method.NAME, f.channel_number
            if name == 'Connection.StartOk':
                send(frame.Method(0, spec.Connection.Tune(0, 131072, 0)))
            elif name == 'Connection.Open':
                send(frame.Method(0, spec.Connection.OpenOk()))
            elif name == 'Channel.Open':
                send(frame.Method(chan, spec.Channel.OpenOk()))
            elif name == 'Basic.Consume':
                tag = f.method.consumer_tag
                send(frame.Method(chan, spec.Basic.ConsumeOk(tag)))
                send(frame.Method(
                    chan, spec.Basic.Deliver(tag, 1, False, 'ex', 'q')))
                sock.sendall(content_header(chan, seconds))
                send(frame.Body(chan, b'hello'))
            elif name == 'Basic.Cancel':
                send(frame.Method(chan,
                                  spec.Basic.CancelOk(f.method.consumer_tag)))
            elif name == 'Channel.Close':
                send(frame.Method(chan, spec.Channel.CloseOk()))
            elif name == 'Connection.Close':
                send(frame.Method(0, spec.Connection.CloseOk()))


def consume(seconds):
    """Run a consumer with the textbook reconnect loop against that value."""
    listener = socket.socket()
    listener.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    listener.bind(('127.0.0.1', 0))
    listener.listen(8)
    threading.Thread(target=broker, args=(listener, seconds),
                     daemon=True).start()
    params = pika.ConnectionParameters(host='127.0.0.1',
                                       port=listener.getsockname()[1],
                                       heartbeat=0,
                                       blocked_connection_timeout=None)
    for attempt in range(1, 4):
        conn = pika.BlockingConnection(params)
        channel = conn.channel()

        def on_message(ch, method, props, body):
            print(f'    delivered, headers={props.headers}')
            ch.stop_consuming()

        channel.basic_consume('q', on_message)
        try:
            channel.start_consuming()
            print(f'    attempt {attempt}: consumed cleanly')
            conn.close()
            return
        except Exception as exc:
            print(f'    attempt {attempt}: {type(exc).__name__}: {exc}')


for value, label in CASES:
    print(f'header holds the u64 {value} ({label})')
    print('  pika.data.decode_table:')
    table = b'\x02ts' + b'T' + struct.pack('>Q', value)
    try:
        print('   ', data.decode_table(struct.pack('>I', len(table)) + table,
                                       0)[0])
    except Exception as exc:
        print(f'    {type(exc).__name__}: {exc}')
    print('  consumer with a reconnect loop:')
    consume(value)
    print()
```

### Actual output

```
header holds the u64 1772000000000 (a millisecond epoch)
  pika.data.decode_table:
    ValueError: year must be in 1..9999, not 58122
  consumer with a reconnect loop:
    attempt 1: StreamLostError: Stream connection lost: ValueError('year must be in 1..9999, not 58122')
    attempt 2: StreamLostError: Stream connection lost: ValueError('year must be in 1..9999, not 58122')
    attempt 3: StreamLostError: Stream connection lost: ValueError('year must be in 1..9999, not 58122')

header holds the u64 18446744073695295616 (1969-07-20 as two's complement, per the Java client)
  pika.data.decode_table:
    OverflowError: timestamp out of range for platform time_t
  consumer with a reconnect loop:
    attempt 1: StreamLostError: Stream connection lost: OverflowError('timestamp out of range for platform time_t')
    attempt 2: StreamLostError: Stream connection lost: OverflowError('timestamp out of range for platform time_t')
    attempt 3: StreamLostError: Stream connection lost: OverflowError('timestamp out of range for platform time_t')
```

### Expected output

The header decodes to something the application can inspect and the connection survives, so the message is delivered on the first attempt.

alonfaraj:
## Proposed Changes

A field-table `timestamp` that falls outside the range `datetime` can represent takes down the whole connection instead of being decoded.

`'T'` is a 64-bit unsigned integer per the AMQP 0-9-1 errata, and RabbitMQ neither range-checks nor rejects it (`rabbit_binary_parser` reads `Value:64/unsigned`, `rabbit_binary_generator` writes it back as `<<V:64>>`). `pika.data.decode_value` passed the raw value to `datetime.fromtimestamp()` unguarded, so anything past year 9999 raised from inside the frame decoder. Nothing on that path expects a stdlib exception, so the transport's generic read-error handler relabelled it as `StreamLostError` and dropped the connection. On a consumer the message was never acked, so it came back on reconnect and the client died again: one header became an unbreakable crash loop, reported as a network fault rather than a decode fault.

The trigger is not exotic. A millisecond epoch does it, and so does any pre-1970 time written by the RabbitMQ Java client, which encodes `'T'` as a signed long (`ValueWriter.writeTimestamp` and `ValueReader.readTimestamp` go through `DataInputStream.readLong`) and therefore puts a large u64 on the wire.

This guards the conversion in `decode_value` and falls back to the raw seconds:

```python
elif kind == _FIELD_TIMESTAMP:
    seconds = _PACK_UNSIGNED_LONG_LONG.unpack_from(encoded, offset)[0]
    try:
        value = _EPOCH + timedelta(seconds=seconds)
    except OverflowError:
        value = seconds
    offset += 8
```

The conversion adds a `timedelta` to a fixed UTC epoch instead of calling `datetime.fromtimestamp`. `fromtimestamp` goes through the platform `time_t`, so the range that decodes varies by OS (on Windows it stops near year 3000) and it raises a mix of `OSError`, `ValueError`, and `OverflowError` depending on the value and platform, as the unguarded code on `main` did:

| u64 on the wire | `datetime.fromtimestamp` on `main` (Linux) |
| --- | --- |
| `253402300799` | `datetime(9999, 12, 31, 23, 59, 59)`, the last value that works |
| `253402300800` | `ValueError: year must be in 1..9999, not 10000` |
| `1772000000000` | `ValueError: year must be in 1..9999, not 58122` |
| `9223372036854775807` | `OSError: [Errno 75] Value too large for defined data type` |
| `18446744073709551615` | `OverflowError: timestamp out of range for platform time_t` |

Adding a `timedelta` is platform-independent: every value past what `datetime` can hold raises `OverflowError`, from either the `timedelta` construction or the addition to the epoch, so that single type is caught and the raw seconds returned. Yielding the raw `int` rather than raising follows what the module already does with a long string that is not valid UTF-8: it decodes to `bytes` instead of failing the frame.

`decode_value`'s docstring gains a `:returns:` line recording the fallback, since that is the user-visible contract.

This does not change how pika interprets the field. pika reads `'T'` as unsigned, which agrees with the broker and the errata, and that is left alone.

## Types of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

## Checklist

## Fixes

- Fixes #1698

### Compatibility

Every value that decoded to a `datetime` before decodes to the same `datetime` now. Only values that previously killed the connection behave differently. Confirmed by decoding 50k random u64 values (none raise) and 20k random in-range values (all identical to the previous expression).

Callers reading a timestamp out of `properties.headers` can now see an `int` where they saw a `datetime`. Nothing that previously worked stops working, since those cases previously lost the connection, so I have not marked this a breaking change. If you would rather treat the widened return type as one, say so and I will relabel it and add a release note.

### Tests

Two regression tests in `tests/unit/data_tests.py` and one in `tests/unit/frame_tests.py`. The frame-level one matters most: the failure boundary is an exception escaping `frame.decode_frame`, which is what reached the connection's read path.

Before the fix, on `main` (unguarded `datetime.fromtimestamp`):

```
tests/unit/data_tests.py::DataTests::test_decode_value_timestamp_max_representable PASSED
tests/unit/data_tests.py::DataTests::test_decode_value_timestamp_out_of_datetime_range FAILED
tests/unit/frame_tests.py::FrameTests::test_decode_header_frame_timestamp_out_of_datetime_range FAILED

FAILED tests/unit/data_tests.py::DataTests::test_decode_value_timestamp_out_of_datetime_range - ValueError: year must be in 1..9999, not 10000
FAILED tests/unit/frame_tests.py::FrameTests::test_decode_header_frame_timestamp_out_of_datetime_range - ValueError: year must be in 1..9999, not 58122
2 failed, 1 passed, 969 deselected
```

After the fix, all three pass.

### DIFF
--- a/pika/data.py
+++ b/pika/data.py
@@ -264,10 +264,14 @@ def decode_value(encoded: bytes, offset: int) -> tuple[Any, int]:
         seconds = _PACK_UNSIGNED_LONG_LONG.unpack_from(encoded, offset)[0]
         try:
             value = _EPOCH + timedelta(seconds=seconds)
-        except (OverflowError, ValueError):
-            # Raising would escape the frame decoder, where the transport
-            # reports anything unexpected as a lost stream and drops the
-            # connection.
+        except OverflowError:
+            # A u64 past what `datetime` can hold overflows the `timedelta` or
+            # its addition to the epoch; both raise `OverflowError` (never
+            # `ValueError`, unlike `datetime.fromtimestamp`). Raising would
+            # escape the frame decoder, where the transport reports anything
+            # unexpected as a lost stream and drops the connection, so fall
+            # back to the raw seconds. This loses the `timestamp` type: re-
+            # encoding the int yields a long-long ('l'), not a 'T'.
             value = seconds
         offset += 8
 
--- a/pika/data.py
+++ b/pika/data.py
@@ -216,6 +216,8 @@ def decode_value(encoded: bytes, offset: int) -> tuple[Any, int]:
 
     :param encoded: The binary encoded data to decode
     :param offset: The starting byte offset
+    :returns: tuple of (decoded value, new offset). A `timestamp` is a u64, so one outside the range
+        `datetime` can hold decodes to its raw integer seconds.
     :raises: pika.exceptions.InvalidFieldTypeException
     """
     # Dispatch on the raw type-tag byte (avoids a bytes allocation per field);
@@ -253,9 +255,14 @@ def decode_value(encoded: bytes, offset: int) -> tuple[Any, int]:
 
     # Timestamp
     elif kind == _FIELD_TIMESTAMP:
-        value = datetime.fromtimestamp(
-            _PACK_UNSIGNED_LONG_LONG.unpack_from(encoded, offset)[0],
-            timezone.utc)
+        seconds = _PACK_UNSIGNED_LONG_LONG.unpack_from(encoded, offset)[0]
+        try:
+            value = datetime.fromtimestamp(seconds, timezone.utc)
+        except (OSError, OverflowError, ValueError):
+            # Raising would escape the frame decoder, where the transport
+            # reports anything unexpected as a lost stream and drops the
+            # connection.
+            value = seconds
         offset += 8
 
     # Field Table
--- a/pika/data.py
+++ b/pika/data.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import calendar
 import decimal
 import struct
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from typing import Any
 
 from pika import exceptions
@@ -51,6 +51,12 @@ _PACK_TAG_INT = struct.Struct('>ci')
 _PACK_TAG_LONG_LONG = struct.Struct('>cq')
 _PACK_TAG_UNSIGNED_LONG_LONG = struct.Struct('>cQ')
 _PACK_TAG_BYTE_INT = struct.Struct('>cBi')
+
+# Reference point for decoding the `timestamp` field type. Adding a
+# `timedelta` is a little slower than `datetime.fromtimestamp` but stays in
+# Python, so the range that decodes does not vary by platform: on Windows
+# `fromtimestamp` goes through a `time_t` that stops at year 3000.
+_EPOCH = datetime(1970, 1, 1, tzinfo=timezone.utc)
 
 # Single-byte `bytes` objects indexed by value. Shared with `pika.spec`,
 # which imports this table for its bit-field buffers.
@@ -257,8 +263,8 @@ def decode_value(encoded: bytes, offset: int) -> tuple[Any, int]:
     elif kind == _FIELD_TIMESTAMP:
         seconds = _PACK_UNSIGNED_LONG_LONG.unpack_from(encoded, offset)[0]
         try:
-            value = datetime.fromtimestamp(seconds, timezone.utc)
-        except (OSError, OverflowError, ValueError):
+            value = _EPOCH + timedelta(seconds=seconds)
+        except (OverflowError, ValueError):
             # Raising would escape the frame decoder, where the transport
             # reports anything unexpected as a lost stream and drops the
             # connection.
--- a/tests/unit/data_tests.py
+++ b/tests/unit/data_tests.py
@@ -206,9 +206,7 @@ class DataTests(unittest.TestCase):
 
     def test_decode_value_timestamp_out_of_datetime_range(self):
         # `timestamp` is a u64 (AMQP 0-9-1 errata), so the wire carries values
-        # `datetime` cannot hold; they must not raise out of the decoder. These
-        # cover each exception `datetime.fromtimestamp` raises across that
-        # range: ValueError, OSError and OverflowError.
+        # `datetime` cannot hold; they must not raise out of the decoder.
         values = (253402300800, 1772000000000, 2**63 - 1, 2**64 - 1)
         decoded = {}
         for seconds in values:
--- a/tests/unit/data_tests.py
+++ b/tests/unit/data_tests.py
@@ -190,3 +190,30 @@ class DataTests(unittest.TestCase):
         result, _ = data.decode_table(encoded, 0)
         self.assertIsInstance(result['k'], bytes)
         self.assertEqual(result['k'], raw)
+
+    def test_decode_value_timestamp_max_representable(self):
+        encoded = b'\x00\x00\x00\x0b\x01kT' + struct.pack('>Q', 253402300799)
+        result, _ = data.decode_table(encoded, 0)
+        self.assertEqual(
+            result['k'],
+            datetime.datetime(9999,
+                              12,
+                              31,
+                              23,
+                              59,
+                              59,
+                              tzinfo=datetime.timezone.utc))
+
+    def test_decode_value_timestamp_out_of_datetime_range(self):
+        # `timestamp` is a u64 (AMQP 0-9-1 errata), so the wire carries values
+        # `datetime` cannot hold; they must not raise out of the decoder. These
+        # cover each exception `datetime.fromtimestamp` raises across that
+        # range: ValueError, OSError and OverflowError.
+        values = (253402300800, 1772000000000, 2**63 - 1, 2**64 - 1)
+        decoded = {}
+        for seconds in values:
+            encoded = (b'\x00\x00\x00\x0b\x01kT' + struct.pack('>Q', seconds))
+            result, offset = data.decode_table(encoded, 0)
+            self.assertEqual(offset, 15)
+            decoded[seconds] = result['k']
+        self.assertEqual(decoded, {seconds: seconds for seconds in values})
--- a/tests/unit/frame_tests.py
+++ b/tests/unit/frame_tests.py
@@ -1,5 +1,6 @@
 """Tests for pika.frame."""
 
+import struct
 import unittest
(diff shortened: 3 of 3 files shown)


lukebakken (approved):
LGTM at 75be46d7.

Guarding the `timestamp` conversion and falling back to the raw seconds is the right fix: an out-of-range u64 header no longer escapes the frame decoder and gets relabelled as a lost stream. Decoding via `_EPOCH + timedelta` rather than `datetime.fromtimestamp` is a nice touch, since it keeps the decodable range off the platform `time_t`.

- Narrowed the `except` to `OverflowError`. With the `timedelta` conversion every out-of-range value raises `OverflowError` from the construction or the epoch addition, never `ValueError`, so the `ValueError` clause was dead code carried over from the `fromtimestamp` approach.
- Documented that the raw-`int` fallback loses the `timestamp` type, so republishing re-encodes it as a long-long ('l') rather than a 'T'.
- Reconciled the PR description with the shipped `timedelta` implementation and its exception surface.

CI is green (60/60); the one Windows failure was a transient RabbitMQ setup step and passed on rerun.

PR #1701, opened 2026-08-27
Resolve "Drop the versionless License :: classifier now that license is a valid SPDX expression

### LINKED ISSUE #1700: Drop the versionless `License ::` classifier now that `license` is a valid SPDX expression
> This issue is part of a larger batch of similar-scoped issues we distribute in the context of our internal vulnerability and dependency tracking activities. The batch covers multiple projects, this is only one of them. The goal is to rise awareness and trigger/initiate the change towards clean project metadata in projects we depend on.

## Summary

This project already declares a valid SPDX license id/expression in its packaging metadata (`License: <BSD-3-Clause>` in the built METADATA), but the deprecated classifier `<License :: OSI Approved :: BSD License>` is still present. That classifier has no version and [PEP 639](https://peps.python.org/pep-0639/)-aware SBOM tooling like [cyclonedx-bom](https://pypi.org/project/cyclonedx-bom/) can't map it to an SPDX id, so it falls back to emitting the raw classifier text as a free-text license name. This free-text field is just noise and causes downstream tooling to flag this project as non-SPDX-compliant.

## Suggested fix

Remove the redundant `License :: OSI Approved :: BSD License` from `classifiers` in the packaging config. No change to the `license` field is needed.

## How to reproduce

```bash
uv tool install cyclonedx-bom
uv venv .venv-repro
uv pip install --python .venv-repro/bin/python pika
cyclonedx-py environment .venv-repro -o sbom.cdx.json
```

... and then check the `.licenses` property of the component. There should only something like:

```json
// This is how it should look like
...
    "license": {
        "acknowledgement": "declared",
        "id": "<VALID SPDX IDENTIFIER>"
    }
...
```

and _not_:

```json
// This is how it should not look like
...
    "license": {
       "acknowledgement": "declared",
       "name": "<CLASSIFIER TAKEN FROM METADATA"
    }
...
```

I'm happy to provide a PR to resolve this timely after approval!

btschwertfeger:
## Proposed Changes

This PR drops the deprecated classifier from the metadata as outlined in the
issue.

Testing done after applying the change:

```bash
❯ uv build .
❯ unzip dist/pika-1.4.0-py3-none-any.whl
❯ cat pika-1.4.0.dist-info/METADATA | grep License
License: BSD-3-Clause
License-File: LICENSE
  <a href='https://github.com/pika/pika/blob/main/LICENSE'><img src='https://img.shields.io/pypi/l/pika.svg?' alt='License'></a>
```
... the classifier is now gone and the metadata looks sane.

> Without the change, the metadata contains the deprecated classifier:
> 
> ```bash
> ❯ cat pika-1.4.0.dist-info/METADATA|grep License
> License: BSD-3-Clause
> Classifier: License :: OSI Approved :: BSD License
> License-File: LICENSE
>   <a href='https://github.com/pika/pika/blob/main/LICENSE'>License</a>
> ```
> ... which is also shown in the output when building:
>
>        ********************************************************************************
>        Please consider removing the following classifiers in favor of a SPDX license expression:
>
>        License :: OSI Approved :: BSD License
>
>        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
>        ********************************************************************************

## Types of Changes

- [x] Cosmetics

## Checklist

Closes https://github.com/pika/pika/issues/1700

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,6 @@ description = "Pika Python AMQP Client Library"
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
-    "License :: OSI Approved :: BSD License",
     "Natural Language :: English",
     "Operating System :: OS Independent",
     "Programming Language :: Python :: 3",


btschwertfeger:
@lukebakken how to trigger the review process?

lukebakken (approved):
Thanks!

PR #1703, opened 2026-08-28
Stop `StreamLostError` from hiding the original traceback

### LINKED ISSUE #1390: SelectConnection catches all errors and hides details.
Originally from an [SO question](https://stackoverflow.com/q/73736444/17404810) but I'll report it here via @lukebakken's request. 

`SelectConnection` seems to catch all errors and doesn't provide any stack trace information. This makes debugging difficult. 

For instance, here's an example where it divides by 0 in the `_on_connection_open` callback. 
```python
import pika

class AsyncServer():
    def __init__(self):
        self._channel = None

    def run(self):
        self._connection = self._connect()

    def _connect(self) -> pika.SelectConnection:
        return pika.SelectConnection(
            parameters=pika.ConnectionParameters(host='localhost'),
            on_open_callback=self._on_connection_open,
            on_close_callback=self._on_connection_closed
        )

    def _on_connection_open(self, connection):
        # Cause an error
        x = 1/0

    def _on_connection_closed(self, connection, reason):
        print("Reason for closing: ",reason)
        print("Exception class: ",reason.__class__)
        self._connection.ioloop.stop()


if __name__ == '__main__':
    recv = AsyncServer()
    recv.run()
```
And the output: 
```
Reason for closing:  Stream connection lost: ZeroDivisionError('division by zero')
Exception class:  <class 'pika.exceptions.StreamLostError'>
```

Notice the exception is a `StreamLostError` instead of a `ZeroDivisionError` which squashes all details of the inner exception. `traceback.print_exc()` won't work here because the exception has already been caught. 

Some possible fixes: 
- Add a flag to disable exception handling and raise them to the top like `BlockingConnection` does
- Send traceback information to the callback provided by `on_close_callback`

jholaj:
## Proposed Changes

An exception raised from user code (an `on_open_callback`, for example) tears down the stream and reaches close callbacks as a `StreamLostError` carrying only the `repr` of the original exception. The traceback showing where it happened is lost.

Set `__cause__` on the wrapping `StreamLostError` instead, so `traceback.print_exception(reason)` prints the full chain, and `BlockingConnection` does so automatically since it re-raises the reason. The `add_on_close_callback` docstring mentions this. The error message is unchanged.

## Types of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

## Checklist

## Fixes

- Fixes #1390

## Further Comments

The issue's preferred fix, propagating the exception out of `ioloop.start()`, would change the error-delivery contract of every async adapter, so this does the smaller thing and just preserves the traceback. The `error`/`reason` split in `_proto_connection_lost` is cosmetic - no behavior change.

### DIFF
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -445,21 +445,24 @@ class BaseConnection(connection.Connection):
 
         if error is None:
             # Either result of `eof_received()` or abort
+            reason = None
             if self._got_eof:
-                error = pika.exceptions.StreamLostError(
+                reason = pika.exceptions.StreamLostError(
                     'Transport indicated EOF',
                     host=self.params.host,
                     port=self.params.port)
         else:
-            error = pika.exceptions.StreamLostError(
+            reason = pika.exceptions.StreamLostError(
                 f'Stream connection lost: {error!r}',
                 host=self.params.host,
                 port=self.params.port)
+            # pika/pika#1390: keep the original traceback reachable
+            reason.__cause__ = error
 
-        LOGGER.log(logging.DEBUG if error is None else logging.ERROR,
-                   'connection_lost: %r', error)
+        LOGGER.log(logging.DEBUG if reason is None else logging.ERROR,
+                   'connection_lost: %r', reason)
 
-        self._on_stream_terminated(error)
+        self._on_stream_terminated(reason)
 
     def _proto_eof_received(self) -> bool:
         """
--- a/pika/adapters/base_connection.py
+++ b/pika/adapters/base_connection.py
@@ -456,7 +456,10 @@ class BaseConnection(connection.Connection):
                 f'Stream connection lost: {error!r}',
                 host=self.params.host,
                 port=self.params.port)
-            # pika/pika#1390: keep the original traceback reachable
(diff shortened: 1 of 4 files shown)


lukebakken (approved):
LGTM at 803a9df6.

Chaining the original error onto `StreamLostError` via `__cause__` is a clean way to stop the wrapper from hiding where the failure came from, and the None/EOF/error branching is preserved exactly.

- The pre-open heuristic in `_on_stream_terminated` replaced `self._error` with a fresh `IncompatibleProtocolError`/`ProbableAuthenticationError`/`ProbableAccessDeniedError` built from `repr(...)`, which carried no `__cause__`, so the chained traceback was dropped for handshake-time failures. It now chains the prior error onto the deduced one, with a test covering the path.
- The `add_on_close_callback` docstring recommended `traceback.print_exception(reason)`, whose single-argument form only exists on Python 3.10+ and breaks on the 3.7-3.9 legacy matrix; it now points at `reason.__cause__` and notes the wrapping is adapter-dependent (TwistedConnection delivers the raw error).
- Documented the traceback/frame-locals retention as an accepted diagnosability cost.

CI is green (60/60).

PR #1704, opened 2026-08-29
Reject decimal.Decimal values with no AMQP wire representation

afonsojanu:
## Proposed Changes

`encode_value()` computes a scale and mantissa for any `decimal.Decimal` passed as an AMQP table value, but never checks that either one actually fits the wire format before packing it. The AMQP decimal type stores the scale as an unsigned octet and the value as a signed 32-bit integer (section 4.2.5.3 of the 0-9-1 spec).

Three ways to hit this while reading through `data.py`:

- A `Decimal` with more than 255 digits after the decimal point (e.g. `Decimal('0.' + '0' * 300 + '1')`) produces a scale that doesn't fit in a byte.
- A `Decimal` whose scaled mantissa is too large for a signed 32-bit int (e.g. `Decimal('99999999999999999999.99')`) overflows the other half of the format.
- `Decimal('NaN')` or `Decimal('Infinity')` have a string exponent (`'n'`/`'N'`/`'F'`) instead of an int, so the existing `exponent < 0` comparison raises `TypeError` rather than doing anything sensible.

All three currently propagate a low-level exception (`struct.error` or `TypeError`) straight out of `encode_value()`, rather than a `pika` exception an application can catch and handle. `encode_short_string()` already guards its own 255-byte limit and raises `ShortStringTooLong`; this does the equivalent for decimals, checking `Decimal.is_finite()` and the scale/mantissa bounds up front and raising a new `UnencodableDecimalError` instead.

## Types of Changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

## Checklist

## Fixes

No issue was filed for this; I ran into it while reading `data.py` and confirmed all three crashes with a small reproduction script before writing the fix.

## Further Comments

Ran the full unit suite (`pytest tests/unit`, excluding the tornado/asyncio adapter tests that need optional extras not installed locally) plus `ruff check`, `yapf --diff --style google`, and `mypy` on the touched files; all clean. The existing `test_encode_table` case (which round-trips a plain `Decimal('3.14')`) still passes, so the common case is unaffected.

### DIFF
--- a/pika/data.py
+++ b/pika/data.py
@@ -167,16 +167,24 @@ def encode_value(pieces: list[bytes], value: Any) -> int:
     elif isinstance(value, decimal.Decimal):
         if not value.is_finite():
             raise exceptions.UnencodableDecimalError(value)
-        value = value.normalize()
-        exponent = value.as_tuple().exponent
+        # Derive the mantissa and scale from as_tuple() using integer
+        # arithmetic. normalize() and Decimal multiplication both apply the
+        # thread-local context precision (28 digits by default), which would
+        # silently round a high-precision value down to a small mantissa that
+        # passes the range check below and encodes as a different number.
+        sign, digits, exponent = value.as_tuple()
         assert isinstance(exponent, int)  # guaranteed by is_finite() above
+        raw = 0
+        for digit in digits:
+            raw = raw * 10 + digit
+        if sign:
+            raw = -raw
         if exponent < 0:
             decimals = -exponent
-            raw = int(value * (decimal.Decimal(10)**decimals))
         else:
             # per spec, the "decimals" octet is unsigned (!)
             decimals = 0
-            raw = int(value)
+            raw *= 10**exponent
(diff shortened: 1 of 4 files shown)


lukebakken (approved):
LGTM at 1218e2e9.

Rejecting decimals with no faithful AMQP wire representation is the right fix, and the range checks against the unsigned-octet scale and signed-long mantissa match spec 4.2.5.3.

One follow-up was addressed on the branch: the original `value.normalize()` plus `Decimal` multiplication both apply the thread-local decimal context precision (28 digits by default), so a finite value with more significant digits was silently rounded to a small mantissa that passed the range check and encoded as a different number. Deriving the mantissa and scale from `value.as_tuple()` with integer arithmetic closes that hole, and it also preserves the given scale so trailing-zero decimals round-trip faithfully. Added tests cover the excess-precision rejection, context independence, and trailing-zero preservation.

CI is green (60/60).

Issue #599, opened 2015-06-10
Decoded floats and doubles get cast to long ints


everilae:
https://github.com/pika/pika/blob/master/pika/data.py#L186 uses the correct format, but then forcibly casts the float to an integer.

vitaly-krugl:
@everilae, what would be an example of float or double from the broker that pika would be decoding? Thx.

everilae:
Well, it is a bit hard to produce since pika as it is now doesn't actually support sending floats... but if pika is supposed to be "an AMQP 0-9-1 client library for Python", it isn't that far fetched to imagine some other client publishing data with floats that pika then would consume and cast to integer. Just take a gander at the linked piece of code:

```python
    value = long(struct.unpack_from('>f', encoded, offset)[0])
```

It does the right-thing-tm up until the ``long()`` wrapping.

vitaly-krugl:
@everilae, I am not finding any references to actual usage of floats, doubles, and decimals in 0.9.1 AMQP specs. It sound like they could easily show up in something like the headers table of BasicProperties, though.

vitaly-krugl:
I am going to be mostly without access to Internet for the next week or so. I will check in when I can.

everilae:
Both https://www.rabbitmq.com/amqp-0-9-1-errata.html#section_3 and

    4.2.1 Formal Protocol Grammar
    ...
    field-value = 't' boolean
                  ...
                  / 'f' float
                  / 'd' double
                  ...
    float = 4*OCTET ; IEEE-754
    double  = 8*OCTET  ; rfc1832 XDR double

define 'f' and 'd' field values.

Pika also tries to support these field values, but casts them to integers. Simple as that.

vitaly-krugl:
@everilae, where are we with this one? Thanks.

everilae:
https://github.com/pika/pika/blob/master/pika/data.py#L238 and https://github.com/pika/pika/blob/master/pika/data.py#L243, still long() casts.

vitaly-krugl:
@everilae: At some point, @gmr suggested switching to https://github.com/gmr/pamqp/tree/master/pamqp, which apparently deals with PY3, floats/doubles, and the like.

everilae:
Aye, been waiting for news on that. gmr/pamqp seemed nice, with a few warts of its own ;)

jholaj:
[...]

Issue #139, opened 2012-03-29
Void field type support


ramdyne:
Hi,

For some reason I've found a RabbitMQ server that is sending fields that pike doesn't appear to support:

Traceback (most recent call last):
  File "zaat.py", line 55, in <module>
    connection.ioloop.start()
  File "/usr/local/lib/python2.6/dist-packages/pika-0.9.5-py2.6.egg/pika/adapters/select_connection.py", line 124, in start
    self.poller.start()
  File "/usr/local/lib/python2.6/dist-packages/pika-0.9.5-py2.6.egg/pika/adapters/select_connection.py", line 374, in start
    self.poll()
  File "/usr/local/lib/python2.6/dist-packages/pika-0.9.5-py2.6.egg/pika/adapters/select_connection.py", line 388, in poll
    self._handler(events[0][0], events[0][1])
  File "/usr/local/lib/python2.6/dist-packages/pika-0.9.5-py2.6.egg/pika/adapters/base_connection.py", line 134, in _handle_events
    self._handle_read()
  File "/usr/local/lib/python2.6/dist-packages/pika-0.9.5-py2.6.egg/pika/adapters/base_connection.py", line 162, in _handle_read
    self._on_data_available(data)
  File "/usr/local/lib/python2.6/dist-packages/pika-0.9.5-py2.6.egg/pika/connection.py", line 569, in _on_data_available
    consumed_count, frame = pika.frame.decode_frame(self._buffer)
  File "/usr/local/lib/python2.6/dist-packages/pika-0.9.5-py2.6.egg/pika/frame.py", line 365, in decode_frame
    out = properties.decode(frame_data[12:])
  File "/usr/local/lib/python2.6/dist-packages/pika-0.9.5-py2.6.egg/pika/spec.py", line 2145, in decode
    (self.headers, offset) = data.decode_table(encoded, offset)
  File "/usr/local/lib/python2.6/dist-packages/pika-0.9.5-py2.6.egg/pika/data.py", line 83, in decode_table
    value, offset = decode_value(encoded, offset)
  File "/usr/local/lib/python2.6/dist-packages/pika-0.9.5-py2.6.egg/pika/data.py", line 128, in decode_value
    kind)
pika.exceptions.InvalidTableError: Unsupported field kind V during decoding

If my Google fu is correct a V field is a void type but other than that I have no idea. Can this field be added or point me in the right direction to add it my self.?

PR #1705, opened 2026-09-02
Encode Python `float` values as the AMQP `d` field type

dylanpulver:
## Proposed Changes

`decode_value` decodes both `f` (`data.py:311`) and `d` (`data.py:316`) into Python floats, but `encode_value` has no `float` branch, so it falls through to `raise exceptions.UnsupportedAMQPFieldException`. `decode` is therefore not invertible over its own output: a field table pika just decoded cannot be handed back to pika.

Measured on this tree:

```
decode a table carrying {'score': d 3.5, 'ratio': f 1.5}
  -> {'score': 3.5, 'ratio': 1.5}   both Python float
re-encode the very dict pika produced
  -> UnsupportedAMQPFieldException: 3.5
spec.BasicProperties(headers={'score': 3.5}).encode()
  -> UnsupportedAMQPFieldException: 3.5
```

So forwarding a message with its headers intact (a shovel, a dead-letter republisher, a retry wrapper) fails on a header any AMQP peer is allowed to send, and `BasicProperties(headers={'k': 3.5})` cannot be published at all.

`d` is the right tag: a Python float is a C double, AMQP 0-9-1 section 4.2.5.5 gives `double = 8*OCTET`, and py-amqp 5.3.1 emits `d` for a Python float (I checked: `dumps('F', [{'score': 3.5}])` gives `0000000f 0573636f7265 64 400c000000000000`, and pika decodes those bytes back to `3.5`).

Note the one thing this does not preserve: a `f` field off the wire is 32 bit, and it comes back out as `d`. The value survives, the tag width does not. Keeping `f` on the way out would need a wrapper type, which seemed like more machinery than the problem deserves; say the word if you would rather have that.

There is precedent in the repo for exactly this shape. Type `x` was decode-only until issue #1011, and the fix was commit 9eaedf00 "Added test, added encode", a +6 line branch in `encode_value` plus a test. Issue #139 is the same story for `V`. Issue #599, the decode half of this one, was closed last week as fixed.

## Why no test caught it

`FIELD_TBL_VALUE` in `tests/unit/data_tests.py`, the only encode corpus, has no Python `float` in it (its `decimal` entry is a `decimal.Decimal`). `test_decode_value_float` and `test_decode_value_double` decode and stop. `test_encode_raises` pins the exception for a `set`, which locks in "unknown types raise" without saying anything about `float`. Nothing round-trips.

## What I ran

`pytest tests/unit`, same venv, both legs in one session: **945 passed / 24 skipped before, 948 / 24 after** (+3 = the new tests). On unmodified `main` all three new tests fail. `yapf --diff --style google`, `ruff check`, `docformatter --check` and `mypy pika/ tests/typing/` all clean (ruff 0.16.5, yapf 0.43.0). Acceptance tests were not run; no RabbitMQ node here.

Encoding as `f` instead of `d` was tried as a check on the tag choice: it fails the byte assertion and the round trip, because 3.14 does not survive float32.

## Types of Changes

- [ ] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

Left unticked; this is a bugfix, non-breaking (it turns an exception into a valid encoding).

## Checklist

Boxes left unticked deliberately, with the facts instead: `CONTRIBUTING.md` and `AGENTS.md` were read; the unit suite passes locally (948/24) but the acceptance suite was not run because there is no broker here; three tests were added and all three fail without the fix; no docs describe the field-table type mapping, so nothing to update.

## Fixes

See #599 (the decode half of the same gap).

---

AI assistance: this change was drafted with Claude Opus 5 (`claude-opus-5`) via Claude Code. Every number above was measured on this branch.

### DIFF
--- a/pika/data.py
+++ b/pika/data.py
@@ -159,7 +159,9 @@ def encode_value(pieces: list[bytes], value: Any) -> int:
             return 9
     elif isinstance(value, float):
         # A Python float is a C double, so 'd' is the lossless tag. It is also
-        # what decode_value yields for both 'f' and 'd'.
+        # what decode_value yields for both 'f' and 'd'. Non-finite values
+        # (inf, nan) are valid IEEE-754 doubles and encode losslessly, unlike
+        # decimal.Decimal, which has no wire representation for them.
         pieces.append(_PACK_TAG_DOUBLE.pack(b'd', value))
         return 9
     elif isinstance(value, decimal.Decimal):
--- a/pika/data.py
+++ b/pika/data.py
@@ -51,6 +51,7 @@ _PACK_TAG_INT = struct.Struct('>ci')
 _PACK_TAG_LONG_LONG = struct.Struct('>cq')
 _PACK_TAG_UNSIGNED_LONG_LONG = struct.Struct('>cQ')
 _PACK_TAG_BYTE_INT = struct.Struct('>cBi')
+_PACK_TAG_DOUBLE = struct.Struct('>cd')
 
 # Single-byte `bytes` objects indexed by value. Shared with `pika.spec`,
 # which imports this table for its bit-field buffers.
@@ -156,6 +157,11 @@ def encode_value(pieces: list[bytes], value: Any) -> int:
         except struct.error:
             pieces.append(_PACK_TAG_LONG_LONG.pack(b'l', value))
             return 9
+    elif isinstance(value, float):
+        # A Python float is a C double, so 'd' is the lossless tag. It is also
+        # what decode_value yields for both 'f' and 'd'.
+        pieces.append(_PACK_TAG_DOUBLE.pack(b'd', value))
+        return 9
     elif isinstance(value, decimal.Decimal):
         value = value.normalize()
         if value.as_tuple().exponent < 0:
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
LGTM at 6c0e58b7.

The core change closes the decode/encode asymmetry cleanly: a `float` encodes as the AMQP `d` (double) tag, matching what `decode_value` already yields for both `f` and `d`. Tag, 9-byte length, and branch placement after the `bool`/`int` returns are all correct.

Follow-up review items are addressed:

- `test_encode_decode_nan_roundtrip` verifies the "every float decode can yield must be encodable" invariant for NaN, which cannot ride the `assertEqual` loop.
- The shared `FIELD_TBL_VALUE` corpus now includes a float, so the broad encode/decode table tests cover the branch.
- A comment documents why non-finite floats encode losslessly as IEEE-754 doubles, unlike `decimal.Decimal`.

Note for merge ordering: open PR #1704 edits the adjacent `decimal` elif in `encode_value`. Whichever lands second should be rebased and re-tested.

CI is green (60/60).

PR #1707, opened 2026-09-03
CI: Test against pre-release Python 3.15 in a non-blocking leg

alonfaraj:
## Proposed Changes

Python 3.15 ships in under a month. Good time to start surfacing friction against it now, without declaring official support and without blocking CI on failures.

https://peps.python.org/pep-0790/#release-schedule

## Types of Changes

- [ ] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

## Fixes

- Fixes #

## Further Comments

Checked again local CPython 3.15.0rc2 (uv-managed), Ubuntu 26.04
Status: 1539 passed, 61 skipped, 0 failed

### DIFF
--- a/.github/workflows/_test.yaml
+++ b/.github/workflows/_test.yaml
@@ -20,6 +20,11 @@ on:
         required: false
         type: boolean
         default: false
+      allow-prereleases:
+        description: 'Let setup-python resolve pre-release Python builds (alpha, beta, rc)'
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   build-windows:
@@ -36,6 +41,7 @@ jobs:
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
           python-version: ${{ matrix.python-version }}
+          allow-prereleases: ${{ inputs.allow-prereleases }}
       - name: Resolve RabbitMQ and Erlang versions
         id: versions
         shell: pwsh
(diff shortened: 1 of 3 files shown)


lukebakken:
@alonfaraj `macos` is slow and flaky. I'm becoming more and more tempted to drop it. I doubt that anyone runs a serious application that uses Pika on `macos` in production anyway.

lukebakken (approved):
LGTM at 69bd9dba.

Adding an `allow-prereleases` input and a non-gating `test-preview` leg is a clean way to get early warning on Python 3.15 without letting an unreleased interpreter block merges. The input is correctly declared with a safe `false` default and wired to all three setup-python steps.

- The preview leg is now Linux-only (`linux-only` input skips the windows and macos jobs), so it no longer burns the slow macOS and Windows runners every push and pull request for a smoke test.
- It opts out of coverage upload (`upload-coverage: false`), so pre-release-interpreter coverage no longer mixes into the Codecov report the `coverage` job collects.
- The non-blocking intent is documented as depending on branch protection requiring only the `tests-passed` check, not every job.
- AGENTS.md is updated to describe the reusable workflow as called three times, including the preview leg.

The live run confirms it: `test-preview` runs only its two Linux py3.15 jobs (both green), the windows/macos preview jobs skip, and no 3.15 coverage artifact is produced. CI is green (62/62, 2 skipped). The one macOS failure was the pre-existing `TestCloseFromConsumerCallback` close-timing flake, unrelated to this CI-config change, and passed on rerun.

alonfaraj:
> @alonfaraj `macos` is slow and flaky. I'm becoming more and more tempted to drop it. I doubt that anyone runs a serious application that uses Pika on `macos` in production anyway.

Yeah it's super slow which is kinda frustrating.
Guess you right, but might be a problem for end user running it locally for dev purpose.
I can remember once or twice we came across specific issues in macos-only so it might be fine, we can always bring it back if needed.

PR #1709, opened 2026-09-03
CI: run macOS on one Python version in a minimal non-blocking leg

lukebakken:
## Summary

macOS runners are slow and occasionally flaky, so testing every supported Python version there is not worth the cost. This restructures CI to test only the latest supported CPython on macOS, in a minimal non-blocking leg, while keeping the platform covered where it matters.

- Removed macOS from the reusable `_test.yaml`, which now covers Linux and Windows. Dropped the `macos-runner` input and its use in `test-modern` and `test-legacy`.
- Added a standalone `test-macos` job to `main.yaml` running the latest supported CPython (3.14) on `macos-latest` across both TLS modes. It is absent from the `tests-passed` needs list, so a macOS flake does not block a merge, but it still exercises the BSD-only `KQueuePoller` path.
- The `coverage` job now waits for `test-macos` so the macOS coverage of `KQueuePoller` is merged into the Codecov report, while tolerating a macOS flake so a failure there does not drop the whole upload.
- Kept `macos-latest` in the mypy matrix so `KQueuePoller`'s kqueue code stays type-checked (mypy narrows the platform for the type checker, so only a darwin run reaches that code).
- Updated `.ci/README.md` and `AGENTS.md` to describe the split.

## Context

This came out of a review of the change. A prior draft removed macOS entirely, which surfaced that macOS was the only environment exercising `KQueuePoller` both at runtime (its tests are `skipIf(not hasattr(select, 'kqueue'))`) and for type checking, and that dropping its coverage risked the Codecov project status. Keeping a single latest-Python macOS leg preserves all of that at a fraction of the runtime cost.

## What was tested

- All three workflows parse; `_test.yaml` runs Linux and Windows only; `test-macos` runs `macos-latest` with `python-version: ['3.14']` and `test-tls: [true, false]`; `test-macos` is excluded from `tests-passed`; `mypy` still includes `macos-latest`.
- CI on this PR.

## Note

If branch protection lists required checks by name, the old per-Python `build/test on macos-*` entries no longer report and may need pruning from repo settings.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -21,7 +21,6 @@ jobs:
     with:
       python-versions: '["3.10","3.11","3.12","3.13","3.14"]'
       ubuntu-runner: ubuntu-latest
-      macos-runner: macos-latest
       legacy: false
 
   test-legacy:
@@ -29,7 +28,6 @@ jobs:
     with:
       python-versions: '["3.7","3.8","3.9"]'
       ubuntu-runner: ubuntu-22.04
-      macos-runner: macos-15-intel
       legacy: true
 
   # Pre-release Python. Deliberately absent from the `tests-passed` needs list:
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yaml
alonfaraj: `expensive macOS and Windows legs` - > `expensive Windows leg`

alonfaraj (changes requested):
LGTM, just one comment (in the comments..)

PR #1711, opened 2026-09-10
Rewrite proposal-recovery.md using state machine

suchitd:
State every design decision directly, inspired by `findings.md` and `design-state-machine.md`: lifecycle states, the guard/exception pattern, the locking model, the persistent-loop redial sequence, the topology ledger, and the channel/connection recovery races.


lukebakken (approved):
I'll merge, then I will review the `design/connection-recovery` branch.

PR #1714, opened 2026-09-16
Add the connection recovery design working area

lukebakken:
> [!NOTE]
> This pull request description was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who approved it before filing. The documents themselves are not Claude's in origin: `proposal-recovery.md` came from @suchitd's gist and was rewritten by @suchitd in e24746ec, and `findings.md` records measurements @lukebakken ran against real clients. Claude wrote the subsequent hardening commits and the review that drove them.

## Summary

Adds `design/connection-recovery/`, a versioned working area for pika's built-in connection and topology recovery, tracking discussion #1654. **No recovery code ships here and nothing under `pika/` changes.** Two files outside `design/` do change, and are the only reason this is not a documentation-only pull request: `pyproject.toml` adds `design/check_docs.py` to the seven format, lint and typecheck recipes, and `AGENTS.md` records that coverage. The design directory itself sits outside `docs/`, so it is not built into the published documentation site.

The point of landing it as a PR is to make the design reviewable and diffable in place, rather than in a gist and a discussion thread.

## What is here

- `README.md` - entry point and index, and the place the pika 2.0.0 constraints that govern this design are recorded once: one `Connection`/`Channel` type, all other adapters removed, and the public API free to change.
- `proposal-recovery.md` - the design proposal and the authoritative document. Public API, lifecycle state, the topology ledger, the redial sequence, per-entity skip/abort handling, a file-by-file change list, and a test plan. It also carries a glossary and an editing-conventions section, both added because reviews kept turning up the same ambiguities.
- `design-state-machine.md` - the framing the proposal follows, written up while the direction was being evaluated. Kept for the derivation, the AMQP 1.0 Java client precedent it borrows from, and the questions it raised.
- `findings.md` - what we measured about how the RabbitMQ Java client and `amqp091-go` behave when an application publishes during recovery. Both fail fast; neither buffers or replays. Harnesses at https://github.com/lukebakken/amqp091-misc.
- `check_docs.py` - a checker for the invariants reviews kept catching: backticked `Class.member` citations verified against the real source and its base classes, the manifest's bare backticked names against the class its bullet declares, `file:line` citations against actual file lengths, cross-references against headings pooled across the tree, fenced Python blocks parsed, the formatting rules the documents state about themselves, a rule that manifest sections name and point rather than explain, and two defects that recurred often enough to mechanise: a stated count disagreeing with the list beneath it, and a sentence beginning with a conjunction after a full stop, which is the seam an insertion leaves. **Read its coverage line rather than its exit code**: it reports 284 symbol citations seen, 92 verified, 119 unresolvable, 39 uppercase-skipped, 34 proposed-exempt and 0 ambiguous, because resolving a receiver like `self._channel._state` needs type inference it does not attempt. The buckets are asserted to sum to the number seen, since for several commits they did not and the line could not be read as the breakdown it looks like.

## Where the design currently stands

Recovery is a first-class `RECOVERING` state on the adapter `Connection`/`Channel`, every public operation guards on it, `ConnectionRecovering`/`ChannelRecovering` are raised during it, and the whole thing is driven on the connection's own persistent IOLoop rather than a dedicated recovery thread.

That last choice is a deliberate divergence from **all three** reference clients - Java, .NET and `amqp091-go` each drive recovery on a separate thread of control - taken to avoid a second thread synchronising with the state machine it mutates. The proposal now states the divergence explicitly and, more importantly, states its price: the persistent loop relocates complexity rather than removing it, and buys three absolute obligations in exchange. Nothing on the loop thread may block, recovery must undo the adapter-wide teardown the drop path has already performed, and nothing on the loop thread may raise. A nine-row table enumerates that teardown and says, per row, whether recovery suppresses or reverses it.

Consumer delivery tags get the offset scheme Java and .NET use, which `amqp091-go` lacks - worth noting because this proposal otherwise ports Go's recovery shape, so following it there would have reproduced the defect.

## What review is being asked for

Three adversarial passes have run over this, and the findings are all dispositioned in the commit history. What is left is judgement rather than defect-hunting:

- Is fail-fast with a catchable `RECOVERING` exception the right contract for pika? `findings.md` is the empirical case for it, but it is still a product call.
- Is the persistent-loop divergence worth its three obligations, given every reference client chose otherwise?
- The proposal records four open questions it does not resolve: the exact adapter state set, an opt-in block-until-open mode, the exception hierarchy now that the compatibility argument for it has been retracted, and whether to expose the moment between a successful redial and finished replay.
- The 2.0 Python floor is not recorded in the constraints and should be, since the annotations depend on it.

## What the reviews changed, and what to read sceptically

The document has been through six max-effort review passes and roughly forty hardening commits. The defects clustered into four kinds worth knowing about as a reviewer:

- Corrections applied to one instance and not to the class - the most common by far.
- Rules inferred from a symbol's name rather than read from its body. `Channel._check_not_closed` is a *connection*-closed check, which one draft got wrong with real consequences.
- Pairs of statements in the same document that could not both hold. The sharpest: one section concluded that neither `_check_not_closed` nor `_register_waiter` can host the recovery guard, and another named both as its hosts - which would have raised from `add_callback_threadsafe`, the documented way to react to recovery, and from `ch.close()`.
- A requirement stated and then never satisfied. The guard section asked for the guarded list to be "enumerated rather than derived" and never enumerated it; that table exists now.
- Checks and tests that could not fail. Six of `check_docs.py`'s own checks were inert or wrong in ways earlier commits called proven, and twice the faulty thing was the proof rather than the check. If you read one thing sceptically, make it a claim that something was verified.

Mechanism described in prose is therefore worth more scrutiny than the code sketches, and any claim about pika's current behaviour names the symbol it was checked against so you can re-verify it. That convention changed during review and the change is worth flagging: it previously asked for `file:line` on every such claim and was met twice in eight hundred lines against roughly ninety-five claims, so it now asks for the symbol, which `check_docs.py` can resolve against parsed source, and reserves `file:line` for where a specific line is the evidence. Say if you would rather have the stricter rule enforced instead. The checker verifies that citations resolve; it cannot check that they mean what the sentence says.

## Testing

No behaviour changes to ship, so the checks are the tests. `hatch run fmt-check`, `lint-check`, `docfmt-check`, `typecheck` and `spec-check` all pass, and each of the first four was proven to actually cover the newly added file by planting a defect in it and confirming the gate fails - a path added to a glob that nothing checks looks identical to one that works. `hatch run docs:build` passes with `strict: true`, verified rather than inferred, since the build reads `docs_dir: docs` and never references `design/`. `python3 design/check_docs.py` reports zero problems. Every check added or repaired here was proven against planted defects *and* against near-miss variants that must stay quiet, which is what caught two false positives and one real hole in the checker's own fence handling.

Deliberately not added: a workflow step running the docs check. These are planning artefacts nothing imports or ships, and a checker that resolves well under half the citations it sees should not gate unrelated pull requests. The reasoning is recorded in the commit that wires the recipes.

## Related

See #1654.


lukebakken:
@suchitd FYI I'm working on revisions as I have time, thanks for your latest updates. I'm using the Java, .NET, and Go client libs to refine the doc.

